PR #1, opened 2023-12-07
Update to latest version of libp2p

adlrocha:
Depends on: https://github.com/consensus-shipyard/libp2p-bitswap/pull/1 

This PR upgrades to the latest version of `libp2p` so we can pick up the right version of `subtle` expected by `filecoin-proofs`. The upgrade from our latest version includes a lot of breaking changes, and on top of this we may actually no be even able to use the unmaintained version of Bitswap that we currently use.


adlrocha:
Related to the libp2p upgrade: https://github.com/libp2p/rust-libp2p/pull/5003

adlrocha:
Being tackled with the new code structure here: https://github.com/consensus-shipyard/ipc-monorepo/pull/522

PR #3, opened 2023-12-19
CHORE: Pull upstream

aakoshh:
Ran `git subtree pull -P fendermint https://github.com/consensus-shipyard/fendermint main` to test getting the latest changes.

Let's not use _Squash_ on these and then we can maintain a nice git history.


adlrocha (approved):
Really nice. I actually shared a comment in the other PR in case it saved you a bit of time, but I once again you beat me :)

PR #499, opened 2023-12-21
Fix fendermint compilation issues

cryptoAtwill:
Fix fendermint compilation issues, i.e. clippy and also format. 

Currently tests are still breaking:
```
Caused by:
    failed to execute contract call to f410forjzm4nb2ly4r4qaqjv2xjtfc6pvhinxl5py4zi:
    code: 33
    error: <no data; potential ABI mismatch>
    info: message failed with backtrace:
    00: f065 (method 3844450837) -- contract reverted (33)
    01: f065 (method 6) -- contract reverted (33)
    02: f010 (method 2) -- send aborted with code 33 (33)
    03: f01 (method 3) -- constructor failed: send aborted with code 33 (33)
    04: f0120 (method 1) -- constructor reverted (33)
    ', fendermint/fendermint/testing/contract-test/tests/staking/machine.rs:130:14
```
Will fix this in follow up PR.


fendermint/fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -31,12 +30,29 @@ use super::{
...
+    topdown_router: ContractCaller<
+        DB,
+        TopDownFinalityFacet<MockProvider>,
+        top_down_finality_facet::TopDownFinalityFacetErrors,
+    >,
+    xnet_messaging_facet: ContractCaller<
+        DB,
+        XnetMessagingFacet<MockProvider>,
+        xnet_messaging_facet::XnetMessagingFacetErrors,
+    >,
```
aakoshh, line 55: The naming convention was `WhateverFacet` became `whatever` field, so I think these should be called `checkpointing`, `topdown` and `xnet_messaging`. I can't see why it's `checkpoint_router` as "router" does not appear in the type, and all of them are facets so `_facet` in the third is redundant.

ipc/ipc/sdk/src/subnet.rs
```diff
@@ -28,7 +28,7 @@ pub enum PermissionMode {
 }
 
 /// Defines the supply source of a subnet on its parent subnet.
-#[derive(Clone, Debug, Serialize_tuple, Deserialize_tuple, PartialEq, Eq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
```
aakoshh, line 31: Why change these now?

cryptoAtwill: we are supporting lotus related now right? `Serialize_tuple` produces a different result than `Serialize`, just be consistent with other serialization methods.

aakoshh: Yeah, I think they chose this in some cases to save on the field names. I think conceptually it's like `{foo: 1, bar: "hello"}` vs `[1, "hello"]`.

raulk: I'm a bit confused because these SDK-level structs shouldn't need any kind of serialization? They are just convenience types for decoupling the contract ABIs from the application-level, right? The data from these structs is pulled into the binding-generated ABI types in `create_subnet`, but never serialized in this form?

raulk: Bottom line: you may be able to remove the serde derive altogether here.

cryptoAtwill: It's a bit legacy code as previously it's used in lotus.

raulk: You mean that these types in the past _were_ previously serialized to feed them into legacy or Wasm actors, right? In that case, they would've been tuple-encoded. However, I suggest you simply drop the serde derivation here all together.

contracts/.gitignore
```diff
@@ -6,6 +6,7 @@ crytic-export/
 .env
 broadcast/
 out/
+binding/src
```
raulk, line 9: @aakoshh When contracts/bindings change and a dev is adjusting the SDK and other downstream components, is the expectation that they'll locally generate the bindings and adapt the dependent code, but only check in the latter and not the former? Then the CI build would generate these _before_ trying to build and would push the updated bindings?

fendermint/fendermint/testing/contract-test/tests/staking/machine.rs
```diff
@@ -110,13 +110,15 @@ impl StateMachine for StakingMachine {
...
-            min_activation_collateral: from_fvm::to_eth_tokens(&parent_ipc.gateway.min_collateral)
-                .unwrap(),
+            // activate the subnet by default
+            min_activation_collateral: et::U256::zero(),
             min_validators: 1,
             min_cross_msg_fee: et::U256::zero(),
-            permissioned: false,
+            permission_mode: 0, // collateral based
+            supply_source: ipc_actors_abis::register_subnet_facet::SupplySource {
+                kind: 0, // native token
```
raulk, line 119: Can we use an enum entry here instead?

contracts/.gitignore
aakoshh: This is up for debate, but currently I tried to follow what you said about not checking in generated code at all. So you are expected to update the SDK and Fendermint, check in those changes, but not check in the Ethers generated bindings, and CI will build it but _also not_ push it either, it will never appear on Github.

That said, if it's useful to be able to share links to the Rust code, we can check it in. The reason I didn't do it is because people reported it changes for no reason. 

However today I noticed that `binding/src/lib.rs` changes in a predictable way: when @cryptoAtwill pushes the modules are in alphabetical order; when I push they are in the order of appearance in `build.rs`. So I think there is something that formats the generated code on his end, it's not the fault of `ethers` being different.

PR #501, opened 2023-12-22
Fixing fendermint tests

cryptoAtwill:
Initial attempt to fix `fendermint` tests. Currently the `staking` tests are still breaking, due to:

1. In solidity contract, `L2+` support seems cannot be enabled, see https://github.com/consensus-shipyard/ipc-monorepo/blob/056e389730858dd829109be3435d86db99e02042/contracts/src/gateway/GatewayManagerFacet.sol#L35 `fendermint` tests have subnets with 3 address routes and is blocked by the above check. This PR temporarily disabled the above check to confirm the check is causing `prop_cross_msgs_hash` to fail.
2. The configuration number in `fendermint` seems not updated correctly, which is still under investigation. It seems the [activated](https://github.com/consensus-shipyard/ipc-monorepo/blob/056e389730858dd829109be3435d86db99e02042/fendermint/fendermint/testing/contract-test/tests/staking/state.rs#L199) flag is not updated (or updated somewhere) and causing the configuration number to be incorrect.
3. This PR updated: https://github.com/consensus-shipyard/ipc-monorepo/blob/5010d3b2ed58a7fe87b41175afa5545a0b40215b/fendermint/fendermint/testing/contract-test/tests/staking/machine.rs#L113 so that contract will not revert.
4. This PR also added https://github.com/consensus-shipyard/ipc-monorepo/blob/5010d3b2ed58a7fe87b41175afa5545a0b40215b/fendermint/fendermint/testing/contract-test/tests/staking/machine.rs#L114 This could be the cause of `configurationNumber` not correct, but from the [check](https://github.com/consensus-shipyard/ipc-monorepo/blob/5010d3b2ed58a7fe87b41175afa5545a0b40215b/fendermint/fendermint/testing/contract-test/tests/staking/machine.rs#L176) here and the list of validators [joined](https://github.com/consensus-shipyard/ipc-monorepo/blob/5010d3b2ed58a7fe87b41175afa5545a0b40215b/fendermint/fendermint/testing/contract-test/tests/staking/machine.rs#L137) for bootstrap, it seems like the correct change? Might need @aakoshh's help to confirm.


aakoshh:
> It seems the [activated](https://github.com/consensus-shipyard/ipc-monorepo/blob/056e389730858dd829109be3435d86db99e02042/fendermint/fendermint/testing/contract-test/tests/staking/state.rs#L199) flag is not updated (or updated somewhere) and causing the configuration number to be incorrect

You deleted the code that set it:

https://github.com/consensus-shipyard/ipc-monorepo/pull/499/files#diff-4442c6c802087c4d316f14e0d0e3a0299bb4a07cc994f19ebcfdc63d515c4aed

cryptoAtwill:
> > It seems the [activated](https://github.com/consensus-shipyard/ipc-monorepo/blob/056e389730858dd829109be3435d86db99e02042/fendermint/fendermint/testing/contract-test/tests/staking/state.rs#L199) flag is not updated (or updated somewhere) and causing the configuration number to be incorrect
> 
> You deleted the code that set it:
> 
> https://github.com/consensus-shipyard/ipc-monorepo/pull/499/files#diff-4442c6c802087c4d316f14e0d0e3a0299bb4a07cc994f19ebcfdc63d515c4aed

haha, oops, I will add it back

contracts/binding/src/lib.rs
```diff
@@ -2,47 +2,47 @@
 #[macro_use]
 mod convert;
 #[allow(clippy::all)]
-pub mod i_diamond;
+pub mod bottom_up_router_facet;
```
aakoshh, line 5: Ah, look, here's a reason this changes: I think you are formatting the generated code. 

`build.rs` produces these modules in [alphabetical order](https://github.com/consensus-shipyard/ipc-monorepo/blob/c4e8e0eb5db12fe0aa13b207af375e76b7e1cbb6/contracts/binding/build.rs#L38-L73), and I specifically asked [rustfmt to ignore](https://github.com/consensus-shipyard/ipc-monorepo/blob/c4e8e0eb5db12fe0aa13b207af375e76b7e1cbb6/rustfmt.toml#L4) these files, yet it changes.

fendermint/fendermint/testing/contract-test/tests/staking/machine.rs
```diff
@@ -361,9 +363,11 @@ impl StateMachine for StakingMachine {
                     result.expect("checkpoint submission should succeed");
                 }
             }
-            StakingCommand::Join(_, value, _) => {
+            StakingCommand::Join(eth_addr, value, _) => {
                 if value.is_zero() {
                     result.expect_err("should not join with 0 value");
+                } else if pre_state.has_staked(&eth_addr) {
+                    result.expect_err("should not join again");
```
aakoshh, line 370: If this changed you also have to update [join](https://github.com/consensus-shipyard/ipc-monorepo/blob/c4e8e0eb5db12fe0aa13b207af375e76b7e1cbb6/fendermint/fendermint/testing/contract-test/tests/staking/state.rs#L348-L351) in the state to be similar to [stake](https://github.com/consensus-shipyard/ipc-monorepo/blob/c4e8e0eb5db12fe0aa13b207af375e76b7e1cbb6/fendermint/fendermint/testing/contract-test/tests/staking/state.rs#L370-L371) and return if the validator has already joined.

PR #502, opened 2023-12-22
Update subnet activation logic

cryptoAtwill:
This PR updates the subnet activation logic in `State` so that it aligns with that in the contract.


raulk:
Unfortunately we didn't have the latest Solidity code here. I just updated to the last and final version here: https://github.com/consensus-shipyard/ipc-monorepo/pull/502/commits/6653ce44780adf9405e299582386908e43dba734.

Of course, that broke the build, so I'm fixing it now.

PR #504, opened 2023-12-25
Fund with token

cryptoAtwill:
Implement `fund_with_token` api for erc20 token transfers, see [here](https://github.com/consensus-shipyard/ipc-monorepo/blob/0bae28b402a9bfef072e70048480d22e8d70e59c/contracts/src/gateway/GatewayManagerFacet.sol#L178C14-L178C27).

### DIFF
--- a/ipc/ipc/provider/src/lib.rs
+++ b/ipc/ipc/provider/src/lib.rs
@@ -479,6 +479,30 @@ impl IpcProvider {
             .await
     }
 
+    /// Funds an account in a child subnet with erc20 token, provided that the supply source kind is
+    /// `ERC20`. If `from` is None, it will use the default address config in `ipc.toml`.
+    /// If `to` is `None`, the `from` account will be funded.
+    pub async fn fund_with_token(
+        &mut self,
+        subnet: SubnetID,
+        from: Option<Address>,
+        to: Option<Address>,
+        amount: TokenAmount,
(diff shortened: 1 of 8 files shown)


ipc/ipc/provider/src/lib.rs
```diff
@@ -479,6 +479,30 @@ impl IpcProvider {
...
+    /// `ERC20`. If `from` is None, it will use the default address config in `ipc.toml`.
+    /// If `to` is `None`, the `from` account will be funded.
+    pub async fn fund_with_token(
+        &mut self,
+        subnet: SubnetID,
+        from: Option<Address>,
+        to: Option<Address>,
+        amount: TokenAmount,
+    ) -> anyhow::Result<ChainEpoch> {
+        let parent = subnet.parent().ok_or_else(|| anyhow!("no parent found"))?;
```
raulk, line 492: Does it make sense to perform a sanity check that the subnet indeed has a ERC20 supply source? If not, it will be caught anyway by the gateway actor, so probably not worth it. However, is this able to return the corresponding Solidity exception in a human-friendly form?

raulk (approved):

ipc/ipc/provider/src/lib.rs
cryptoAtwill: yep, should be able to. From past experience, in contract execution, errors are usually returned.

PR #507, opened 2023-12-28
eth_getCode, eth_getStorageAt: return empty on failure to invoke EVM.

raulk:
Workaround for https://github.com/consensus-shipyard/ipc-monorepo/issues/506. I have the actual solution in a half-baked branch, but aiming to unblock Fluence with this patch.


fendermint/fendermint/eth/api/src/apis/eth.rs
```diff
@@ -731,14 +731,19 @@ where
...
             evm::Method::GetStorageAt,
             params,
             height,
         )
-        .await?;
+        .await
+    else {
+        // TODO Actually compare the actor type and return empty if not an EVM actor.
+        //  See https://github.com/consensus-shipyard/ipc-monorepo/issues/506.
+        return Ok(Default::default());
```
aakoshh, line 745: With the placholder actor, where `None` is returned by the query, the existing code returns it as `"00...00"`. To be consistent with that, this shouldn't return `""` either.

raulk: True, good catch!

PR #510, opened 2024-01-04
[PARKED] BUILD: Use the stipper trick locally

aakoshh:
Changes the `local` version of the docker build to use the `stripper` stage like on CI to build first the dependencies, then the application code, so that the dependencies can be cached.

This is motivated by the observation that sometimes switching between branches results in a full compilation, even if seemingly not much changed, which takes ~10 minutes. Other times, after some changes in the application, it's done much quicker because of the caching done by docker buildkit, rebuilding just `fendermint_app` for example. 

There is a trade-off with this, however, because this way it will _always_ recompile all of our own crates, resulting in a 2 minute build. Ostensibly this could be faster if the previous output would be cached.

I'm putting it out here for reference, and then see if people see the reason for it to be merged.

### DIFF
--- a/fendermint/Makefile
+++ b/fendermint/Makefile
@@ -74,14 +74,14 @@ docker-deps: $(BUILTIN_ACTORS_BUNDLE) $(IPC_ACTORS_GEN)
 # it can handle for anyone with `DOCKER_BUILDKIT=1`.
 docker-build: docker-deps $(FENDERMINT_CODE)
 	if [ "$(PROFILE)" = "ci" ]; then \
-		cat docker/builder.ci.Dockerfile docker/runner.Dockerfile > docker/Dockerfile ; \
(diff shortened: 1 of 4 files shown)


raulk (commented):
> This is motivated by the observation that sometimes switching between branches results in a full compilation, even if seemingly not much changed, which takes ~10 minutes.

This is interesting -- do you have an idea of what could be causing this? Perhaps an inconspicuous change in Cargo.lock of some modules deep in the build tree?

aakoshh:
I am not sure yet. Now that we have a monorepo I have to pass the entire root directory to get everything we need to compile Rust, which can bring in all kinds of stuff. After the first compilation it seems to be fast (without this change). If you observe anything let me know.

jsoares:
I'm going to cancel this for now, given our plans to take a deeper look. But it's linked in the corresponding linear task so we don't lose track of the direction.

PR #518, opened 2024-01-08
IPC-512: Update code layout

fridrik01:
This PR updates the code layout as described in https://github.com/consensus-shipyard/ipc-monorepo/issues/512. 

Some comments:
- I removed install-infra script as we have fendermint in repo now, and updated instead infra paths to use infra/fendermint
- I moved the ipc/README to README and made adjustments to it. 
- I tried to have the build prerequisites in main README and refer to it in other READMES
- When testing I saw that existing docs (like quickstart-calibration) did not work (missing params in subnet create), I left those for future PR.

Items remaining:
- [ ] Repos cleanup
- [x] Update all documentation under `docs/` to reflect new code layout changes
- [x] Update the README 
- [x] Make sure infra scripts work
- [x] Make sure docker scripts work


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,44 @@
+name: Build and Test Workflow
```
aakoshh, line 1: Can we call this from `ci.yaml` like the rest of them? By the looks of it it is only about contracts, so it should be named `contracts-build-and-test.yaml` to fit in with its brethren. Then it will also benefit from the conditional execution tags.

Makefile
```diff
@@ -1,7 +1,7 @@
 # Each major sub-repository in this monorepo has their own Makefiles;
 # instead of making an even more compilicated common one, let's delegate to them.
 
-SUBTREES := fendermint ipc ipld-resolver fvm-utils contracts
+SUBTREES := fendermint ipc ipld contracts
```
aakoshh, line 4:
```suggestion
SUBTREES := fendermint ipc ipld/resolver contracts
```

I haven't seen yet if there is now a `Makefile` under `ipld`, but this results in a `cd <subtree>` and expect `make lint` to work for example.

contracts/Makefile
```diff
@@ -63,7 +63,7 @@ check-subnet:
 # Development support
 
 lint: fmt
-	solhint 'src/**/*.sol'
+	npx solhint 'src/**/*.sol'
```
aakoshh, line 66: :heart:

Cargo.toml
```diff
@@ -3,32 +3,28 @@ resolver = "2"
...
+  "fendermint/app/options",
+  "fendermint/crypto",
+  "fendermint/app/settings",
+  "fendermint/eth/*",
+  "fendermint/rocksdb",
+  "fendermint/rpc",
+  "fendermint/storage",
+  "fendermint/testing",
+  "fendermint/testing/*-test",
+  "fendermint/vm/*",
```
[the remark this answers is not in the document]
aakoshh: I see `docs` were moved

aakoshh (commented):
Thank you for taking this on, much appreciated! 

The only change I'd do is move `cli` to be `ipc/cli` because currently it doesn't seem to have a corresponding `Makefile` to make sure it's tested and formatted. Notably `ipc/Makefile` doesn't list it any more, and unlike `fendermint/Makefile` it doesn't automatically discovery it from cargo packages either, so I don't think it will be exercised on CI.

Oh and the unified approach to workflows rooted at `ci.yaml`.

.github/workflows/contracts-build-and-test.yaml
```diff
@@ -1,4 +1,4 @@
-name: Build and Test Workflow
+name: Contract Build and Test Workflow
 
 on:
   push:
```
aakoshh, line 4: This needs to change to the pattern in the other workflow files to allow invocation.

adlrocha (commented):
@cryptoAtwill , @aakoshh, what do you think about removing `./testing`? Those testing frameworks were using the agent and Lotus and we haven't updated them since. We can keep them in the archived repo for reference in case we need to get some inspiration from them in the future, but I don't know if it makes sense to keep them as part of this repo. WDYT?

adlrocha (approved):
Overall LGTM (although I have to say that reviewing has been a bit of a pain because my browser kept crashing with so much files :see_no_evil:)

Before merging remember removing `./testing` (as the team seems to be onboard with this) :pray:

Cargo.toml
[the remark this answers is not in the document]
raulk: AFAIK the `src/` convention works within crates (and that's how Rust devs expect it's used). If we apply it here, the `src/` dir would be _outside_ a crate and we'd end up with pedantic paths like `fendermint/src/abci/src`. FWIW, the only non-crate dir left behind is `docker`.

PR #525, opened 2024-01-10
Adding msg batch handling in fendermint

cryptoAtwill:
Adding msg batch related code in fendermint. This is almost the same process as compared to bottom up checkpoint, just function names and types are different.

Currently this PR has adding the over process flow, there are some duplicated code that can be removed, will do the clean up after integration testing.

### DIFF
--- a/contracts/src/gateway/router/BottomUpRouterFacet.sol
+++ b/contracts/src/gateway/router/BottomUpRouterFacet.sol
@@ -108,6 +108,15 @@ contract BottomUpRouterFacet is GatewayActorModifiers {
             membershipWeight: membershipWeight,
             majorityPercentage: s.majorityPercentage
         });
+
+        if (LibGateway.bottomUpBatchMsgsExists(batch.blockHeight)) {
+            return;
+        }
+
+        if (batch.blockHeight % s.bottomUpMsgBatchPeriod != 0) {
+            revert InvalidBatchEpoch();
+        }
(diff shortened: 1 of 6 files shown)


aakoshh (commented):
Looks correct to me. 

I believe the sheer level of code duplication shows what a profoundly bad idea this is. Let me remind (as everyone knows by now) that this makes it possible for message batches and checkpoints to arrive out of order and be rejected because of invalid signatures. 

The simpler alternative is to keep messages inside the checkpoints, lift the rule that checkpoints arrive at regular intervals, add a link to the previous checkpoint, and create them on-demand. Then the order is fixed and there is no problem, as well as no code duplication across the entire stack. For example no need for a second Merkle tree stuff, no need to two kinds of signatures to be broadcasted, no need for two kinds of things to be relayed, etc.

aakoshh:
It's a nitpick but I wouldn't call this feature `fix-bottomup`. This wasn't a bug, it's a design change, and I'll argue to the end it's actually the opposite of a fix, it's a bug lure. Is there a ticket for this? Can we use the ticket number in the branch name to make it easy to identify?

The same applies to https://github.com/consensus-shipyard/ipc-monorepo/pull/526

adlrocha:
@cryptoAtwill, this one has been deprecated by https://github.com/consensus-shipyard/ipc/pull/564, right? I am closing it for now. Feel free to re-open it if this is not the case.

PR #534, opened 2024-01-11
fix: fendermint docker image name

fridrik01:
After moving to monorepo, we can not use the repository name anymore as the docker image


aakoshh (approved):
Thank you :pray:

Issue #519, opened 2024-01-08
Upgrade to FVM 4.0.0


fridrik01:
We need to upgrade to FVM 4.0.0 which has newer dependencies (cid, libp2p, etc) than we are currently using here. 

Then we need to upgrade to whatever FVM 4 version which has the pluggable syscall feature allowing us to deploy custom kernels and syscalls

PR #542, opened 2024-01-11
IPC-519: Upgrade to FVM 4.0.0

fridrik01:
See: https://github.com/consensus-shipyard/ipc/issues/519

This PR upgrades the FVM to version 4.0.0.

Although the https://github.com/filecoin-project/ref-fvm/pull/1922 feature is not in 4.0.0, it should be easy to add through a tag/branch dependency, will add that in a followup PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ integer-encoding = { version = "3.0.3", default-features = false }
 jsonrpc-v2 = { version = "0.11", default-features = false, features = ["bytes-v10"] }
 k256 = "0.11" # Same as tendermint-rs
 lazy_static = "1.4"
-libipld = { version = "0.14", default-features = false, features = ["dag-cbor"] }
+libipld = { version = "0.16", default-features = false, features = ["dag-cbor"] }
 libp2p = { version = "0.53", default-features = false, features = [
   "gossipsub",
   "kad",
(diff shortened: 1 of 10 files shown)


fridrik01:
Moving back to draft as tests were failing

Cargo.toml
```diff
@@ -150,15 +150,15 @@ openssl = { version = "0.10", features = ["vendored"] }
 # Using the 3.3 version of the FVM because the newer ones update the IPLD dependencies
 # to version which are different than the ones in the builtin-actors project, and since
 # they are 0.x cargo cannot upgrade them automatically, which leads to version conflicts.
-fvm = { version = "~3.2", default-features = false }     # no opencl feature or it fails on CI
-fvm_shared = { version = "~3.2", features = ["crypto"] }
-fvm_sdk = { version = "~3.2" }
+fvm = { version = "4.0.0", default-features = false }     # no opencl feature or it fails on CI
+fvm_shared = { version = "4.0.0", features = ["crypto"] }
+fvm_sdk = { version = "4.0.0" }
```
aakoshh, line 155: I wonder if it's worth limiting the updates to eg. `"~4.0"` so we accept patches but not new features automatically? Based on prior experience they can be breaking even on a minor release.

aakoshh (approved):
LGTM :+1:

Cargo.toml
fridrik01: Probably not atm, will need to update this to support the pluggable syscalls for custom kernel

PR #543, opened 2024-01-11
IPC-530: Execute `IpcEnvelope`

### LINKED ISSUE #530: GMP: Execution of `IpcEnvelope`
- [x] Implementation of new data structures for `IpcEnvelope` and cross-net messages as described [here](https://www.notion.so/pl-strflt/General-Message-Passing-7686edb5b65c4f9086fa4d4288c35906?pvs=4#55294071dfd648c1ae6868a71db28fda)
- [x] Implementation of the new subnet actor and gateway methods required in the gateway to forward and trigger the execution of the `IpcEnvelope` as detailed [here](https://www.notion.so/pl-strflt/General-Message-Passing-7686edb5b65c4f9086fa4d4288c35906?pvs=4#5a98a85172f24e17b1526f014b34877e). Remember that the design still mentions the enforcement of firewall policies, but these are out of scope for the first iteration. Cross-net message ACLs are explicitly implemented as part of the IPC-enabled user smart contracts.

adlrocha:
Closes #530


contracts/src/gateway/GatewayMessengerFacet.sol
```diff
@@ -3,51 +3,59 @@ pragma solidity 0.8.19;
...
-    using StorableMsgHelper for StorableMsg;
     using SupplySourceHelper for address;
+    using CrossMsgHelper for IpcEnvelope;
 
     /**
      * @dev sends a general-purpose cross-message from the local subnet to the destination subnet.
      *
      * IMPORTANT: `msg.value` is expected to equal to the value sent in `crossMsg.value` plus the cross-messaging fee.
+     * Only smart contracts are allowed to trigger these cross-net messages, users
+     * can always send funds from their address to the destination subnet and the run the transaction in the destination
```
maciejwitowski, line 30:
```suggestion
     * can always send funds from their address to the destination subnet and then run the transaction in the destination
```

contracts/examples/contracts/cross-token/ERC20TokenMessenger.sol
```diff
@@ -101,7 +101,7 @@ contract ERC20TokenMessenger is ReentrancyGuard {
             wrapped: false
         });
 
-        return messenger.sendUserXnetMessage{value: msg.value}(crossMsg);
+        return messenger.sendContractXnetMessage{value: msg.value}(crossMsg);
```
maciejwitowski, line 104: Not strictly related to this PR, but should `return` be used here if `sendContractXnetMessage` doesn't return anything?

adlrocha: We will end up deprecating this contract as part of the GMP work. This was a sample contract that used the previous implementation of GMP, so we can tackle it then.

contracts/src/lib/LibGateway.sol
```diff
@@ -339,35 +349,38 @@ library LibGateway {
...
-            revert InvalidCrossMsgDstSubnet();
-        }
-
-        // If the crossnet destination is NOT the current network (network where the gateway is running),
-        // we add it to the postbox for further propagation.
-        if (!crossMsg.message.to.subnetId.equals(s.networkName)) {
-            bytes32 cid = crossMsg.toHash();
-            s.postbox[cid] = crossMsg;
+        if (crossMsg.to.subnetId.isEmpty()) {
+            sendReceipt(crossMsg, false, abi.encodeWithSelector(InvalidCrossMsgDstSubnet.selector));
```
raulk, line 363: We need to distinguish between system-generated failures, and callee-generated failures. Currently it would be possible for a callee to trick the caller to think that the system generated a failure, if it throws a system error e.g. `InvalidCrossMsgDstSubnet`.

Proposal: make the result field an enum and have three values: Ok, SysErr, CallerErr,

.github/workflows/ci.yaml
```diff
@@ -63,10 +63,6 @@ jobs:
     uses: ./.github/workflows/contracts-deployment-test.yaml
     needs: [ contracts-prettier ]
 
-  contracts-build-and-test:
-    uses: ./.github/workflows/contracts-build-and-test.yaml
-    needs: [contracts-prettier]
-
```
raulk, line 69: @dnkolegov Mind elaborating on why we're removing this workflow?

dnkolegov: It is the same as contracts-deployment-test. It was renamed and then appeared again probably after merging.

raulk: Good catch!

raulk: Actually, not really. They run different make targets in the end.

raulk: - contracts-deployment.yaml runs `make deploy-ipc`
- contracts-build-and-test.yaml runs `make`

dnkolegov: Yes, I changed that to run the explicit target

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,115 @@
...
+interface IpcHandler {
+    error CallerIsNotGateway();
+    error UnsupportedMsgKind();
+    error UnrecognizedResult();
+
+    /// @notice Entrypoint for handling xnet messages in IPC-aware contracts.
+    function handleIpcMessage(IpcEnvelope calldata envelope) external payable returns (bytes memory ret);
+}
+
+abstract contract IpcExchange is IpcHandler, Ownable {
```
maciejwitowski, line 25: @raulk should this file be renamed to `IpcExchange.sol` or `IpcHandler` to avoid introducing more names?

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
maciejwitowski, line 1: @raulk should this file be renamed to `IpcExchange.sol` or `IpcHandler` to avoid introducing more names?

raulk: I think we can separate the interface and the abstract contract into two separate files.

contracts/src/lib/LibGateway.sol
```diff
@@ -386,20 +400,130 @@ library LibGateway {
...
+        if (crossMsg.isEmpty()) {
+            revert CannotCreateIpcReceipt();
+        }
+
+        // if we get a `Transfer` or a `Receipt` do nothing, no need to send receipts.
+        // - Transfers don't need to be acknowledged.
+        // - And sending a `Receipt` to a `Receipt` could lead to amplification loops.
+        if (crossMsg.kind == IpcMsgKind.Transfer || crossMsg.kind == IpcMsgKind.Receipt) {
+            return;
+        }
```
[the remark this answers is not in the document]
raulk: Follow-up: we added results for transfers, and this condition just matches on receipt. I still think this should be an assertion.

raulk:
IMO, this is ready for merge and further iteration on `main`. Will do on Friday.

contracts/src/lib/LibGateway.sol
```diff
@@ -386,20 +400,149 @@ library LibGateway {
...
+    function sendReceipt(IpcEnvelope memory crossMsg, OutcomeType outcomeType, bytes memory ret) internal {
+        if (crossMsg.isEmpty()) {
+            revert CannotCreateIpcReceipt();
+        }
+
+        // if we get a `Receipt` do nothing, no need to send receipts.
+        // - And sending a `Receipt` to a `Receipt` could lead to amplification loops.
+        if (crossMsg.kind == IpcMsgKind.Result) {
+            return;
+        }
```
raulk, line 454: I commented on the original PR, but recording here. How could we arrive here, and should this be an assertion instead?

contracts/src/lib/LibGateway.sol
```diff
@@ -386,20 +400,149 @@ library LibGateway {
...
+            reject = to.down(s.networkName).getActor().hasSupplyOfKind(SupplyKind.ERC20);
+        }
+        if (reject && crossMessage.kind == IpcMsgKind.Transfer) {
+            revert MethodNotAllowed("propagation of `Transfer` messages not suppported for subnets with ERC20 supply");
+        }
+
+        // If the directionality is top-down, or if we're inverting the direction
+        // because we're the LCA, commit a top-down message.
+        if (applyType == IPCMsgType.TopDown || isLCA) {
+            ++s.appliedTopDownNonce;
```
raulk, line 515: Alfonso pointed out that this might be wrong. We'll need to revisit this:

https://filecoinproject.slack.com/archives/C05J93G9G4F/p1705391013362889

contracts/src/lib/SupplySourceHelper.sol
```diff
@@ -54,31 +54,102 @@ library SupplySourceHelper {
...
-            ret = Address.functionCallWithValue({target: target, data: data, value: value});
+            (success, ret) = functionCallWithValue({target: target, data: data, value: value});
         } else if (supplySource.kind == SupplyKind.ERC20) {
             // Transfer the tokens first, _then_ perform the call.
-            IERC20(supplySource.tokenAddress).safeTransfer({to: target, value: value});
-            ret = Address.functionCall(target, data);
+            (success, ret) = ierc20Transfer(supplySource, target, value);
+            if (success) {
+                // Perform the call only if the ERC20 was successful.
+                (success, ret) = functionCallWithValue(target, data, 0);
```
raulk, line 108: Issue: the transfer + function call is not atomic. We need to figure out if putting these behind a function that uses revert semantics, and then calling it via delegatecall, would make these two actions atomic.

fendermint/vm/actor_interface/src/ipc.rs
```diff
@@ -88,7 +84,7 @@ lazy_static! {
                             abi: ia::subnet_actor_manager_facet::SUBNETACTORMANAGERFACET_ABI
                                 .to_owned(),
                         },
-                        EthFacet {
+EthFacet {
```
raulk, line 87: Strange that cargo fmt is not complaining here?

fendermint/app/Cargo.toml
```diff
@@ -72,16 +72,6 @@ quickcheck_macros = { workspace = true }
...
-
-# Load the same built-in actor bundle as the ref-fvm integration tests. We'll probably need built-in actors,
-# for example to deploy Solidity code. We can compile Wasm actors and deploy them too, but certain functions
-# in `ref-fvm` like looking up actor addresses depend on built-in actors like the `InitActor` maintaining state.
-# NOTE: The following would add this as a dependency. The problem is that this makes it more difficult to compile
-# the project because now when there is a new version of the FVM released, we also need a new version of the
-# actor project to be released. In prod, we'd just load it from a file, so let's see if that works.
-# We can build a bundle CAR with the Makefile.
-# actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", branch = "next" }
-
```
raulk, line 84: Was this necessary? cc @aakoshh

fendermint/app/Cargo.toml
```diff
@@ -72,16 +72,6 @@ quickcheck_macros = { workspace = true }
...
 # Using a single binary to run the application as well as to execute client commands.
 [[bin]]
```
raulk, line 76: Eh? What is this highlighting, GitHub?

ipc/api/src/evm.rs
```diff
@@ -81,99 +79,48 @@ macro_rules! cross_msg_types {
...
-                let msg_value = fil_to_eth_amount(&value.value)?;
-                let msg_fee = fil_to_eth_amount(&value.fee)?;
+            fn try_from(value: IpcEnvelope) -> Result<Self, Self::Error> {
+                let val = fil_to_eth_amount(&value.value)?;
 
-                log::info!(
-                    "storable message token amount: {:}, converted: {:?}",
-                    value.value.atto().to_string(),
-                    msg_value
-                );
```
raulk, line 95: @cryptoAtwill why remove this log line?

ipc/provider/src/checkpoint.rs
```diff
@@ -203,18 +199,22 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
                     .await?;
                 log::debug!("bottom up bundle: {bundle:?}");
 
-                todo!("implement submiet checkpoint")
```
raulk, line 206: Was this on `main`? 😨

raulk (approved):
Let's do it!

fendermint/app/Cargo.toml
aakoshh: I'm sorry, what do you mean exactly? 

So, originally I added a project reference to that specific branch, and then we had access to the type definitions. Then, we switched to loading the bundle CAR and having just what we needed copied to the `actor_interfaces` crate, but I left here this comment as a historical reminder. Now this comment has been removed. Wasn't necessary to remove it, but I have no problem with it. 

Which part are you asking?

fendermint/vm/actor_interface/src/ipc.rs
aakoshh: I think it's because we're inside a macro, `lazy_static!`. Within these macros we don't get any formatting, which is why the indent is inconsistent [here](https://github.com/consensus-shipyard/ipc/blob/c0bec4e3bee187fe78a6811ef985333927d0aba7/fendermint/app/src/cmd/key.rs#L49-L73).

fendermint/app/Cargo.toml
aakoshh: I wondered about this too, but it's not a bad thing. Easier to see than on https://docs.cometbft.com/v0.37/core/configuration where these have almost the same colour as a comment, which is why it took so long to figure out the right env var name for them.

contracts/src/lib/LibGateway.sol
```diff
@@ -386,20 +400,149 @@ library LibGateway {
...
+    /// @notice Sends a receipt from the execution of a cross-message.
+    /// Only `Call` messages trigger a receipt. Transfer messages should be directly
+    /// handled by the peer client to return the funds to the from address in the
+    /// failing network.
+    /// (we could optionally trigger a receipt from `Transfer`s to, but without
+    /// multi-level execution it would be adding unnecessary overhead).
+    function sendReceipt(IpcEnvelope memory crossMsg, OutcomeType outcomeType, bytes memory ret) internal {
+        if (crossMsg.isEmpty()) {
+            revert CannotCreateIpcReceipt();
+        }
```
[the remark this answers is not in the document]
cryptoAtwill: My take on this is a sanity check. Ideally `crossMsg` should not have been empty. If this happens that means validators have agreed to submit empty cross message. It will block checkpoint from being committed. Personally it's hard to tell if this is good or bad because if this happens, something is really wrong among the validators. But on the other hand, if validators all agree on empty cross msg, so be it, we can just return for the submission to go on.

contracts/src/lib/LibGateway.sol
```diff
@@ -235,21 +242,21 @@ library LibGateway {
...
 
         emit NewTopDownMessage({subnet: subnetId.getAddress(), message: crossMessage});
     }
 
     /// @notice Commits a new cross-net message to a message batch for execution
     /// @param crossMessage - the cross message to be committed
-    function commitBottomUpMsg(CrossMsg memory crossMessage) internal {
+    function commitBottomUpMsg(IpcEnvelope memory crossMessage) internal {
         GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
-        uint256 epoch = getNextEpoch(block.number, s.bottomUpMsgBatchPeriod);
```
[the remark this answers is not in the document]
cryptoAtwill: ok, I think I removed all of them in previous PRs, but somehow it might be lost in conflicts resolving. I will make another pass to remove them.

contracts/src/lib/LibGateway.sol
cryptoAtwill: This would a sanity check. But a potential case would be:
- In the parent, a contract `Transfer` some value to another contract in the child.
- Child rejects the fund due to no `fallback` nor `receive`.
- Receipt send back to parent with refund.
- Original contract rejects fund transfer as no `fallback` nor `receive`. 

The above should reach here.

ipc/provider/src/checkpoint.rs
cryptoAtwill: I think this was in between the BottomUpBatch and BottomUpCheckpoint refactoring, and the contract merged but relayer not ready yet, so it's unimplemented first.

ipc/api/src/evm.rs
cryptoAtwill: The caller should have logged it instead of in the conversion function.

PR #545, opened 2024-01-11
IPC-544: Make VoteRecord generic in its content

### LINKED ISSUE #544: Refactor the vote record to be generic in its payload

aakoshh:
Closes #544 

As the title says, it makes the IPLD `VoteRecord` generic in its content, with the intention that the whole application has an enum for all the kinds of votes that can come up in a subnet, all cast by validators. The generic parameter has been added to the `Service` and the `Client` as well, so serialization is taken care of in the library and we just deal with domain types.


ipld/resolver/src/vote_record.rs
```diff
@@ -38,24 +39,21 @@ impl<'de> Deserialize<'de> for ValidatorKey {
...
     pub public_key: ValidatorKey,
-    /// The subnet in which the vote is valid, to prevent a vote on the same CID
+    /// The subnet in which the vote is valid, to prevent a vote on the same subject
     /// in one subnet being replayed by an attacker on a different subnet.
     pub subnet_id: SubnetID,
-    /// The CID of the content the vote is about.
-    pub cid: Cid,
-    /// The claim of the vote, in case there can be votes about multiple facets
-    /// regarding the CID.
-    pub claim: String,
```
raulk, line 53: @aakoshh could you elaborate on these removals?

raulk (approved):

ipld/resolver/src/vote_record.rs
aakoshh: Sure, when I originally added this feature it the story was that we want to vote on the availability of some piece of data, represented by a CID. 

For example the subnet signals that they want to submit a checkpoint by sending a CID with a quorum to the parent. Then, the nodes on the parent have to resolve the CID of the checkpoint to its contents via the Resolver. But a validator faces exactly the same dilemma we have in our current story: when should they propose it in a parent block for execution? If they propose it too early and the majority hasn't yet resolved the CID, they will not vote for the block, risking liveness. 

The escape hatch was this voting: let's gossip that _"I have the CID, safe to propose"_, and only propose once we have a quorum to support it. I added a `claim` just in case we wish to differentiate what we are saying about it, but it was just an idea. 

This PR replaces the CID with a custom content, in this case a block hash (which is a CID in case the parent is Lotus) and a block height. I thought if we make the content flexible like that, we might as well remove the `claim` and use an `enum` for all the various statements we can vote on, such as `HaveCid | ParentFinality | ...`.

PR #549, opened 2024-01-12
DEBUG: Can we publish a different image?

aakoshh:
Trying to find out why publishing fails after https://github.com/consensus-shipyard/ipc/pull/534


aakoshh:
Configured `ipc` to be _Admin_ of the `fendermint` package of the `consensus-shipyard` org at https://github.com/orgs/consensus-shipyard/packages/container/fendermint/settings

PR #553, opened 2024-01-12
New bottomup

cryptoAtwill:
Merge `BottomUpMsgBatch` and `BottomUpCheckpoint` into one.

So now the overall bottom up flow is as follows:
- Send a `release` or some bottom up cross message
- Store the cross message in `BottomUpMsgBatch` of the next checkpoint period
- If the cross msg size exceeds max batch limit, make a batch cut and store the new batch at the current epoch.

Fendermint check at the end of every epoch to see if there is a new message batch. If there is, create a bottom up checkpoint with the message. Or check if the bottom up period has reached to create a checkpoint with empty cross messages.


contracts/src/lib/LibGateway.sol
```diff
@@ -259,38 +275,41 @@ library LibGateway {
...
-                });
-                uint256 msgLength = batch.msgs.length;
-                for (uint256 i; i < msgLength;) {
-                    newBatch.msgs[i] = batch.msgs[i];
-                    unchecked {
-                        ++i;
-                    }
+            return;
+        }
+
```
adlrocha, line 280:
```suggestion
```

contracts/src/lib/LibGateway.sol
```diff
@@ -259,38 +275,41 @@ library LibGateway {
...
+        // if the maximum size was already achieved emit already the event
```
adlrocha, line 281:
```suggestion
        // if the maximum size was already reached emit already the event
```

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -153,4 +96,28 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
...
+            revert MaxMsgsPerBatchExceeded();
+        }
+
+        // if the bottom up messages' length is max, we consider that epoch valid
+        if (checkpoint.msgs.length == s.maxMsgsPerBottomUpBatch) {
+            return;
+        }
+
+        // the max batch size not reached, we only support checkpoint period submission.
+        uint256 lastBottomUpCheckpointHeight = s.lastBottomUpCheckpointHeight;
```
adlrocha, line 115: I guess this is fine for now, but we may relaxing this requirement in the future and let children submit a checkpoint whenever they want or need to.

contracts/test/integration/GatewayDiamond.t.sol
```diff
@@ -1605,565 +1613,107 @@ contract GatewayActorDiamondTest is Test, IntegrationTestBase {
...
-        gwBottomUpRouterFacet.execBottomUpMsgBatch(batch);
-
-        // fail with no messages
-        batch = BottomUpMsgBatch({subnetID: subnetId, blockHeight: d, msgs: new CrossMsg[](0)});
-
-        vm.prank(caller);
-        vm.expectRevert(BatchWithNoMessages.selector);
-        gwBottomUpRouterFacet.execBottomUpMsgBatch(batch);
-    }
+    // function testGatewayDiamond_execMsgBatch_WithMessages() public {
```
adlrocha, line 1616: Let's uncomment and fix these issues. Thanks!

adlrocha (commented):
Overall, I think the change makes sense to me. Let's fix the test and figure out the best way to merge with the GMP PR. Thanks!

PR #559, opened 2024-01-12
IPC-550: Finality vote tally

### LINKED ISSUE #550: Create tally component to aggregate votes

aakoshh:
Closes #550 

Adds an STM enabled `VoteTally` component to `fendermint_vm_topdown`. 

The way it works is that it keeps track of the finalized parent chain as observed by this node, that is, a single vector of block hashes that is never expected to be rolled back. This is slightly controversial, as there can be reorgs on Lotus, however it's not any more controversial than the way we handle finality: by waiting until a block is sufficiently embedded, then executing the messages up to that point in our instant-finality forward-only ledger. The ledger itself couldn't handle reorgs either. 

What that in mind, it is expected that the top-down syncer will add new finalized blocks to this component, while it will also receive votes from the gossip channel from other validators. Some of these votes will be about blocks we have not observed, which is why the votes go into a map indexed by height and hash until the time comes to sum them up looking for a quorum. At that point we go from the latest block we observed back towards the last finalized block in the ledger and sum up the votes until we find a height which has quorum. Every vote is implicitly voting on all previous heights.

Note that even if we know about votes in future blocks, unless these are heights that we consider final, we don't count them. That is because there is no evidence that these are on the came chain as the one we see. On the other hand if we get to the same height and see the same hash, then necessarily all previous hashes must match as well. We trust the syncer to feed us a chain. So, even if the syncer knows more blocks that lead to this vote, unless they are final this component ignores them.

A conceptual illustration:

```
PC: B0 - B1 - B2 - B3 - B4
--------------------------
V1:           B2                     B7
V2:      B1   B2                 B6
V3:      B1             B4   B5
```

Say we observe `B0 - B4` on the parent chain, and below that we see the votes we received from validators `V1`, `V2` and `V3`. Say our last finalized block was `B0`. In this case we'll find quorum for `B2` because `V1` and `V2` explicitly voted for it, and `V3` voted for `B4`, and we see that `B2` was indeed an ancestor of `B4` in the `PC` chain. However, we don't consider `B5` final, because even though `V1` and `V2` have voted for `B7` and `B6`, we don't actually know whether there has been a network partition until we ourselves see these blocks being added to `PC`. 

Even if we proposed `B5`, we couldn't execute it until we have it in `PC`, so if the validators want to finalize these blocks they can do it on their own, at which point we'll be forced to wait until we have it, but that's a different story.

TODO:
- [x] Add a view of the parent chain from our perspective
- [x] Add votes coming from validators (including us)
- [x] Find a block that we think is final and has enough votes to pass quorum
- [x] Detect equivocations
- [x] Clear out finalized blocks
- [x] Update power table
- [x] Property based testing


raulk:
> In this case we'll find quorum for B3 because V1 and V2 explicitly voted for it

I think you mean to say B2, right?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+        // Check that we are extending the chain. We could also ignore existing heights.
+        if let Some((prev, _)) = chain.get_max() {
+            if block_height != prev + 1 {
+                return abort(Error::UnexpectedBlock(prev + 1, block_height));
+            }
+        }
+
+        chain.insert(block_height, block_hash);
+
+        self.chain.write(chain)?;
```
fridrik01, line 122: Are you using immutable datastructures due to using the stm library? Copying them should be fast, but what is the performance if we need to do multiple times, like it seems we are doing here? Did you consider just using mutable datastructure combined with simple locks as an alternative?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        if !self.known_validator(&validator_key)? {
+            return abort(Error::UnknownValidator(validator_key));
+        }
+
+        let mut votes = self.votes.read_clone()?;
+        let votes_at_height = votes.entry(block_height).or_default();
+
+        for (bh, vs) in votes_at_height.iter() {
+            if *bh != block_hash && vs.contains(&validator_key) {
```
fridrik01, line 152: ah, so `Equivocation` is for when a validator votes on a height with different hash than he did previously?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+        }
+
+        let votes_for_block = votes_at_height.entry(block_hash).or_default();
+
+        if votes_for_block.insert(validator_key).is_some() {
+            return Ok(false);
+        }
+
+        if *self.pause_votes.read()? {
+            retry()?;
```
fridrik01, line 169: hmm, what does `retry` do here when we are not in a transaction?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        for (block_height, block_hash) in chain.iter().rev() {
+            if block_height == finalized_height {
+                break;
+            }
+            // Skipping null blocks
+            if let Some(block_hash) = block_hash {
+                if let Some(votes_at_height) = votes.get(block_height) {
+                    if let Some(votes_for_block) = votes_at_height.get(block_hash) {
+                        for vk in votes_for_block {
```
fridrik01, line 208: nit: check for None instead to keep code less indented for readability

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+            return Ok(None);
+        };
+
+        let votes = self.votes.read()?;
+        let power_table = self.power_table.read()?;
+
+        let mut weight = 0;
+        let mut voters = im::HashSet::new();
+
+        for (block_height, block_hash) in chain.iter().rev() {
```
fridrik01, line 200: so you basically just look from oldest block to newest and add up the weights from all the voters until you exceed the power threshold, and at that height, we have found last common block that will pass validator proposal?

fridrik01 (approved):
Quite readable even though I lacked a lot of context here. I left some questions/nit but overall looks good!

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    #[error("unknown validator: {0:?}")]
+    UnknownValidator(K),
+
+    #[error(
+        "equivocation by validator {0:?} at height {1}; {} != {}",
+        hex::encode(.2),
+        hex::encode(.3)
+    )]
+    Equivocation(K, BlockHeight, V, V),
```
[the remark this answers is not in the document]
aakoshh: Oh, no worries, I can explain. I tried to quickly find; here's one in the CometBFT docs about [Evidence](https://docs.cometbft.com/v0.38/spec/consensus/evidence#equivocation).

It's (what a I think is a common) term in Proof-of-Stake consensus when a Byzantine validator supports two contradicting facts at the same time, where an honest validator can only support one. In this instance we are trying to find a finalized chain, which should be evident if one observes the parent chain; if a validator casts votes for two different block hashes on the same height in the parent chain (on the _finalized_ chain, not the tip where there is forking in the Longest Chain Filecoin), then they are not being honest, or they have misconfigured their node and their finality threshold is too low.

In any case we should ignore their votes because if everyone was allowed to behave this way we'd end up finalizing something random. Hopefully no more than 1/3rd of the voting power ever casts contradicting votes, and then we can still finalize blocks. 

It's something I put in, but didn't pursue to the end where if you equivocate at all then you are out: you can equivocate on height 4, but not on 5, and that will still count as legal on 5. Or you can support chain A on height 4, chain B on height 5, then chain A on height 6 again, and I won't see it as equivocation even when I know that 4 and 6 formed a chain. The potential is there, but currently it's only used for the most obvious case where on the same height there are two conflicting votes from the same validator.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/// Keep track of votes beging gossiped about parent chain finality
+/// and tally up the weights of the validators on the child subnet,
+/// so that we can ask for proposals that are not going to be voted
+/// down.
+pub struct VoteTally<K = ValidatorKey, V = BlockHash> {
```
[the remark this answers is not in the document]
aakoshh: No, it's just the first word that came to mind. To be fair `Collector` sounds general too (reminds of the `.collect::<Vec<_>>()` for example). I don't really like "tally", we can totally rename it before feature branch is merged :+1:

fendermint/vm/topdown/src/voting.rs
aakoshh: That's right, I use `im` because of the way the STM library works, by doing a `.clone()` when replacing values. Copying them is just makes another reference to the root, with modifications having logarithmic cost due to structural sharing. 

It's true that STM will not be as performant as a hand-built solution with mutable data structures and locking. It's a tradeoff for composability and the kind of API it supports, e.g. in https://github.com/consensus-shipyard/ipc/pull/590 I can combine queries that the component wasn't completely aware of, or in https://github.com/consensus-shipyard/ipc/pull/584 I can query two STM components and know they are consistent with each other, the updates happen in lockstep with the syncers own data structures. 

A non-locking alternative of using channels with mutable data structures would surely be faster, but wouldn't compose and could be inconsistent across these isolated islands of state.

fendermint/vm/topdown/src/voting.rs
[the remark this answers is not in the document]
fridrik01: Oh nice!, yeah I think I figured out what this would be used for when I read rest of the PR, but thanks about the details here, super useful!

fendermint/vm/topdown/src/voting.rs
aakoshh: STM is used prominently in [Cardano](https://github.com/replica-io/replica-io/issues/26#issuecomment-1806433718). At the time I was interested in what we could take on from that project and pursued it in Rust. I like using it, but maybe in Rust it's not as nice because for example of the need to clone the inputs. Maybe it's a bit of an overkill for us, but I recommended to Will we try it when he was doing traditional locking and I found some race conditions IIRC. I think it's less tedious then the pattern with channels and clients and response channels in events, like inside the IPLD resolver service and client, and it makes me think a bit. 

I figured in this use case where we have to vote on blocks that are produced every 30 seconds it should be fine. But if the parent would be Fendermint too with 1s block time, gossiping votes could be overwhelming and we should possibly reduce the voting frequency to every N-th block.

fendermint/vm/topdown/src/voting.rs
aakoshh: I'm not sure what you mean by "not in transaction". These methods can only be invoked inside a transaction. I don't expect you to know that, just saying how it is with this library: the transaction is in a thread-local variable and if it wouldn't be set during the execution of `.read()` it would panic. The alternative was to pass it around, which I thought was cumbersome. 

Anyway, what `retry()?;` does is to stop the execution of this transaction and make it subscribe to a change in any of the variables it has read. If any of them change, this transaction will be re-executed. Until then it asynchronously waits for something to happen that changes the equation, such as `pause_votes` becoming `true`. 

I might move it up from where it is so it subscribes to the minimum amount of notifications :thinking: I placed it here so that if it can go ahead _without_ retrying because the vote is a duplicate, it can, but perhaps it's better to do this first.

fendermint/vm/topdown/src/voting.rs
aakoshh: Yes, except it's looking from the newest block back to the oldest, so that it can find quorum for the highest height possible.

fendermint/vm/topdown/src/voting.rs
aakoshh: Changed it to `let Some(...) else { continue; }` to reduce indentation, let me know if this works for you.

PR #560, opened 2024-01-12
Fix issues 37-38

dnkolegov:
This PR packs tightly structs, cache array length, updates subnet keys, and updates storage artifacts.


dnkolegov:
Hi @aakoshh. 
Do you know why, after changing `ConstructorParams`, fendermint tests start failing, but they pass on my machine?

contracts/scripts/deploy-gateway.template.ts
```diff
@@ -117,16 +117,15 @@ export async function deploy(libs: { [key in string]: string }) {
     }
 
     const gatewayConstructorParams = {
+        msgFee: ethers.utils.parseUnits('10', 'gwei'),
+        bottomUpCheckPeriod: 10,
+        activeValidatorsLimit: 100,
+        majorityPercentage: 66,
```
aakoshh, line 123: It has to be `>` than this, right, not `>=`?

contracts/src/GatewayDiamond.sol
```diff
@@ -26,13 +26,13 @@ contract GatewayDiamond {
...
 
     struct ConstructorParams {
-        SubnetID networkName;
-        uint256 bottomUpCheckPeriod;
         // deprecated (for now): no `msgFee` currenlty charged for cross-net messages
         uint256 msgFee;
+        uint256 bottomUpCheckPeriod;
+        uint16 activeValidatorsLimit;
         uint8 majorityPercentage;
+        SubnetID networkName;
```
aakoshh, line 34: Maybe the day will come when we can rename this variable to `subnetId` :crossed_fingers:

aakoshh: Is there any practical reason for reordering these?

contracts/src/SubnetActorDiamond.sol
```diff
@@ -25,18 +25,18 @@ contract SubnetActorDiamond {
...
         uint256 minActivationCollateral;
+        uint256 minCrossMsgFee;
         uint64 minValidators;
         uint64 bottomUpCheckPeriod;
-        uint8 majorityPercentage;
+        address ipcGatewayAddr;
         uint16 activeValidatorsLimit;
-        uint256 minCrossMsgFee;
+        uint8 majorityPercentage;
+        ConsensusType consensus;
```
aakoshh, line 35: Why is this `ConsensusType` required?

contracts/src/GatewayDiamond.sol
dnkolegov: Just to save more space in memory and in the storage.

contracts/src/lib/LibGatewayActorStorage.sol
```diff
@@ -73,6 +46,39 @@ struct GatewayActorStorage {
     bool checkpointRelayerRewards;
     /// @notice Determines if relayers should be rewarded for cross-net message execution
     bool crossMsgRelayerRewards;
+    // == Structs ==
+    /// @notice The current membership of the child subnet
+    Membership currentMembership;
+    /// @notice The last membership received from the parent and adopted
```
aakoshh, line 52: I find this "adopted" confusing. It suggests it should be the current one, but there is another variable for that. 

I know this has nothing to do with the PR, just saying.

contracts/src/SubnetActorDiamond.sol
dnkolegov: Do you mean why we need this field if we have only one consensus protocol?

aakoshh (approved):
LGTM!

contracts/src/lib/LibGatewayActorStorage.sol
dnkolegov: How would you call them? I will create an issue to not forget

contracts/src/SubnetActorDiamond.sol
aakoshh: The only value it has is `Fendermint`; I was just wondering what it really affects, and whether it's right to have such an enum. I don't know how to extend it. 

Sorry, I was lazy to go look; if it's "complicated" that's okay, I thought I'd ask.

contracts/src/lib/LibGatewayActorStorage.sol
aakoshh: Maybe `nextMembership`, if it's the membership "in waiting"?

Wait, I just checked and found [the assignment](https://github.com/consensus-shipyard/ipc/blob/5c840dcfa85176641ca2241d21e7eb8267c40786/contracts/src/lib/LibGateway.sol#L165) didn't even work in the direction I thought it would. 

I thought `currentMembership` was the _currently active_ one and `lastMembership` was for what we have last received an update, but haven't activated yet. But it looks like it's the other way around. 

Does that mean [this](https://github.com/consensus-shipyard/ipc/blob/5c840dcfa85176641ca2241d21e7eb8267c40786/fendermint/vm/interpreter/src/fvm/state/ipc.rs#L157-L162) is completely wrong? It's what we base checkpoints on.

aakoshh: Ah, I see what's happening, the `currentMembership` is being overwritten by something, but before it's assigned to `lastMembership`. Then I would call it `prevMembership`, as in, the one prior to the current one.

aakoshh: But you see how this:
> The current membership of the child subnet
vs what this says:
> The last membership received from the parent and adopted

Is completely confusing. I'm trying to understand what is the difference between them, and while the first sounds obvious, the second talks about receiving things from the parent and adopting them. Well the first one was received from the parent and adopted too :shrug:

PR #564, opened 2024-01-15
Merge bottom up msg batch with checkpoint

cryptoAtwill:
Merge bottom up msg batch with checkpoint mechanism.


contracts/test/integration/SubnetActorDiamond.t.sol
```diff
@@ -807,76 +811,6 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
         }
     }
 
-    function testSubnetActorDiamond_submitBottomUpMsgBatch_basic() public {
```
adlrocha, line 810: I don't know if this is currently but if not, it would be good to have a test message for the happy path of message execution and with some of the failure scenarios that we had before for the message batch code.

adlrocha (approved):
As mentioned in my previous review, overall this LGTM. I agree that many of the unit tests for msgBatch can go, but it would be great to still the happy and a few test scenarios as we had before (e.g. if there is a checkpoint trying to be submitted out of the checkpoint period where the number of messages hasn't reached the limit, or one where the number of messages is over the limit, etc.). Can you have a look at this before merging? Thanks!

Issue #508, opened 2024-01-03
Implement FendermintExterns::get_tipset_cid


aakoshh:
Implement [get_tipset_cid](https://github.com/consensus-shipyard/ipc-monorepo/blob/0bae28b402a9bfef072e70048480d22e8d70e59c/fendermint/fendermint/vm/interpreter/src/fvm/externs.rs#L42-L44) so any contract using the `BLOCKHASH` Solidity function doesn't crash the node.

Things to consider:
* Operators can configure a limited size application history, after which not even CometBFT remembers block headers, so we can't look up block identity uniformly unless the limit is part of consensus.
* Snapshots only export actor state, so after a snapshot restore we have no block history at all, or at least I don't think CometBFT promises that we do. 
* For these reasons I recommend that we add a new built-in actor with an `AMT` and schedule it to be executed by the `cron` actor and store a configurable (in genesis) number of block hashes in the actor state itself, so that snapshot export/import restores the pre-agreed length of history, and then we implement `get_tipset_cid` based on this actor.

raulk:
For EVM compatibility, we only need a lookback of 256 blocks. It may be possible to ask CometBFT to sync X blocks from the network on node init. From then onwards, we assume CometBFT would maintain sync and inform us of new blocks so we can maintain the lookback live.

If it's possible to guarantee that CometBFT can sync N blocks, on init we could find a way to suspend Fendermint init until the lookback cache has been filled. However, @akosh.farkash pointed out that this may not be easy. It also couples IPC further to CometBFT, which ain't good.

As @akosh.farkash suggested, the alternative is to create a "chain history actor" to store the lookback in state. Resolving this syscall would then involve making a transient call to this actor.

Alternatively, that actor could be an extended version of the system actor; in which case we'd have to fork builtin-actors (bad), or replace the system actor with our custom version at genesis (better). I'd choose this path if we wanted to prevent system actor sprawl, but I also don't have strong opinions.

If we generalise this actor to something like a "chain metadata actor", it comes broader in scope and more useful down the line as a good home for future chain-dependent/specific functionality.

raulk:
In conclusion, I think @akosh.farkash has it right and this shouldn't be hard too implement. From my POV:

- [ ] Create a new "chain metadata" actor the state data model Akosh suggested above. Take a look at the suggested interface below.
- [ ] Bundle this actor in a separate bundle with its own manifest.
[...]

PR #568, opened 2024-01-15
Implement get_tipset_cid

fridrik01:
See https://github.com/consensus-shipyard/ipc/issues/508

TODO:
 - [x] Create a new "chain metadata" actor the state data model Akosh suggested in  https://github.com/consensus-shipyard/ipc/issues/508 and using the the suggested interface in the task.
- [x] Bundle this actor in a separate bundle with its own manifest.
- [x] Load the new bundle at genesis.
- [x] Create new actor at genesis, constructing it with a len determined by a code constant.
- [x] Figure out how to obtain the blockhash at the right time within the ABCI pipeline so we can feed it in (take into account the case where the contract calls BLOCKHASH 0, i.e. the *current* block, this happens in immediate execution chains like CometBFT/Ignite, unlike Filecoin)
- [x] Figure out if we need to fork cron, or if we should have a dedicated cron, or if we should call `push_blockcid` individually.
- [x] Implement the `get_tipset_cid` syscall by calling `ChainMetadata::blockcid()` 

Follow up tasks
- [ ] https://github.com/consensus-shipyard/ipc/issues/615 
- [ ] https://github.com/consensus-shipyard/ipc/issues/616
- [ ] https://github.com/consensus-shipyard/ipc/issues/617
- [ ] https://github.com/consensus-shipyard/ipc/issues/618
- [ ] https://github.com/consensus-shipyard/ipc/issues/619


fridrik01:
Mostly marking this as ready to review to trigger CI as this feature is not fully implemented and I expect tests to fail, need to figure out how to save state in actor (get `ipld::block_create -- cannot put block:  bafy2bzacecm52f6bp... not reachable (6: resource not found)`

fendermint/actors/chainmetadata/src/actor.rs
```diff
@@ -0,0 +1,75 @@
...
+fil_actors_runtime::wasm_trampoline!(Actor);
+
+pub struct Actor;
+
+impl Actor {
+    fn constructor(rt: &impl Runtime, params: ConstructorParams) -> Result<(), ActorError> {
+        rt.validate_immediate_caller_is(std::iter::once(&SYSTEM_ACTOR_ADDR))?;
+
+        let state = State {
+            blockhashes: VecDeque::new(),
```
aakoshh, line 24: Have you considered storing this in an AMT? It would allow structural sharing, whereas this will always copy the whole list of hashes.

aakoshh: I think it supports popping values via [delete](https://github.com/filecoin-project/ref-fvm/blob/8a93ecdf43892cc15ba997d9e2c742b272a48fd7/ipld/amt/src/amt.rs#L329). A short example of using this type is in the docstring in https://github.com/consensus-shipyard/ipc/blob/main/ipc/types/src/amt.rs

docs/fendermint/running.md
```diff
@@ -375,6 +375,7 @@ configuration will look for it at `~/.fendermint/bundle.car`, so we might as wel
 ```shell
 make actor-bundle
```
aakoshh, line 376: Nice, but then it looks like [this](https://github.com/consensus-shipyard/ipc/blob/6c83a7c2ec549ae4dfac11415372827323b2507e/fendermint/Makefile#L101) should also depend on the `$(ACTORS_BUNDLE)`.

fendermint/actors/chainmetadata/src/shared.rs
```diff
@@ -0,0 +1,44 @@
...
+pub struct PushBlockParams {
+    pub epoch: ChainEpoch,
+    pub block: Cid,
+}
+
+#[derive(FromPrimitive)]
+#[repr(u64)]
+pub enum Method {
+    Constructor = METHOD_CONSTRUCTOR,
+    PushBlock = 2,
```
aakoshh, line 41: Why is this method a hardcoded `2` when the rest use `frc42_dispatch`?

fendermint/actors/chainmetadata/src/shared.rs
```diff
@@ -0,0 +1,44 @@
...
+// The state is a stores `blockhashes` in an AMT containing the blockhashes of the
+// last `lookback_len` epochs
+#[derive(Serialize_tuple, Deserialize_tuple)]
+pub struct State {
+    // the AMT root cid of blockhashes
+    pub blockhashes: Cid,
+
+    // the maximum size of blockhashes before removing the oldest epoch
+    pub lookback_len: u64,
+}
```
aakoshh, line 18: Maybe this could also have a shared constructor like [this one](https://github.com/consensus-shipyard/ipc/blob/904705346b7788e79920526c95c0e62bee5869a3/fendermint/vm/actor_interface/src/init.rs#L53-L54) to take care of the empty array initialisation.

fendermint/actors/src/manifest.rs
```diff
@@ -0,0 +1,78 @@
...
+use cid::Cid;
+use fvm_ipld_blockstore::Blockstore;
+use fvm_ipld_encoding::CborStore;
+use std::collections::HashMap;
+
+const CHAINMETADATA_ACTOR_NAME: &str = "chainmetadata";
+
+/// A mapping of internal actor CIDs to their respective types.
+pub struct Manifest {
+    pub chainmetadata_code: Cid,
```
aakoshh, line 13: Is it necessary to separate the builtin-actor manifest from the custom actor manifest? I don't see `is_chainmetadata_actor` being used anywhere, but I wonder if there is a way to have a unified approach?

fendermint/actors/src/manifest.rs
```diff
@@ -0,0 +1,78 @@
...
+    by_id: HashMap<u32, Cid>,
+    by_code: HashMap<Cid, u32>,
+}
+
+pub const CHAINMETADATA_ACTOR_CODE_ID: u32 = 1;
+pub const CHAINMETADATA_ACTOR_ID: u64 = 48;
```
aakoshh, line 19: Should this actor have its own module under `actor_interface` where this is declared? I'm already regretting not putting all those IDs into the `lib.rs` of that crate, so it's easier to see at a glance which IDs are free, but this for sure isn't a place I'd be looking for further IDs.

fendermint/actors/src/manifest.rs
aakoshh: I see that the original `Manifest` assigns `code_id` based on this sequence so if we added more entries to the root CID somehow, they could be unstable if the builtin actors are expanded by them in subsequent versions.

What I'm thinking is maybe we don't need a known code ID for this actor. I'm not even sure what the usecase for it is, but for example the IPC actors don't have it, just the identity IDs, which is enough to send messages to a well-known location. Then we could put our own actors at the end of a common list and use a single manifest. 

It's a bit strange anyway that actors have a name and a CID, and then they get an arbitrary ID based on their appearance, which also gets hardcoded in the code as a stable `_CODE_ID` yet if we decided to delete an actor from the middle of our list it would all go haywire. Why not just use the name if we have to look up its CID?

Perhaps this is a pattern we don't have to perpetuate?

fendermint/app/config/default.toml
```diff
@@ -11,6 +11,8 @@ snapshots_dir = "snapshots"
 contracts_dir = "contracts"
 # Builtin actor bundle CAR file.
 builtin_actors_bundle = "bundle.car"
+# In-repo actor bundle CAR file.
+actors_bundle = "actors_bundle.car"
```
aakoshh, line 15: Maybe something like `custom_actors_bundle` would help highlight the difference between the two.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -83,6 +83,31 @@ where
             anyhow::bail!("failed to apply block cron message: {}", err);
         }
 
+        // Push the current block hash to the chainmetadata actor
+        //
+        let params = fvm_ipld_encoding::RawBytes::serialize(
+            fendermint_actor_chainmetadata::PushBlockParams {
+                epoch: height,
+                block: fendermint_vm_message::cid(&state.block_hash().unwrap()).unwrap(),
```
aakoshh, line 91: `fendermint_vm_message::cid` serializes its input and hashes it to produce a CID. That's a one-way conversion so if we wanted to use what we stored to actually identify blocks, say, to retrieve them from the API, we'd be in trouble because while this value would be unique, it wouldn't be recognised by the CometBFT API. 

I think we could _pretend_ that the `[u8; 32]` returned by `block_hash()` is a CID. It is after all a SHA256 (I just don't know if't it's Sha2 or Sha3; I have a `tx_hash` method but it's unclear) of the block, as given to us by CometBFT, and it's certainly possible for a CID to store this.

Then we can unwrap it to get to the raw digest if we have to use it with the CometBFT API.

fendermint/vm/interpreter/src/fvm/genesis.rs
```diff
@@ -228,6 +230,26 @@ where
...
+        let empty_blockhashes_cid =
+            Array::<(), _>::new_with_bit_width(state.store(), BLOCKHASHES_AMT_BITWIDTH).flush()?;
+        state
+            .create_actor(
+                fendermint_actors::CHAINMETADATA_ACTOR_CODE_ID,
+                fendermint_actors::CHAINMETADATA_ACTOR_ID,
+                &fendermint_actor_chainmetadata::State {
+                    blockhashes: empty_blockhashes_cid,
+                    lookback_len: fendermint_actor_chainmetadata::DEFAULT_LOOKBACK_LEN,
+                },
```
aakoshh, line 247: If the `State` had a constructor then this `Array` magic could stay an implementation detail.

fendermint/vm/interpreter/src/fvm/genesis.rs
```diff
@@ -228,6 +230,26 @@ where
...
+        // STAGE 1b: Then we initialize the in-repo actors.
+        //
+
+        // Initialize the chain metadata actor which handles saving metadata about the chain
+        // (e.g. block hashes) which we can query.
+        let empty_blockhashes_cid =
+            Array::<(), _>::new_with_bit_width(state.store(), BLOCKHASHES_AMT_BITWIDTH).flush()?;
+        state
+            .create_actor(
+                fendermint_actors::CHAINMETADATA_ACTOR_CODE_ID,
```
aakoshh, line 242: Thinking of how users could expand the system with their own actors, I really think a unique name would be a better UX than some auto-assigned ID.

fendermint/vm/interpreter/src/fvm/genesis.rs
```diff
@@ -228,6 +230,26 @@ where
...
+                fendermint_actors::CHAINMETADATA_ACTOR_ID,
+                &fendermint_actor_chainmetadata::State {
+                    blockhashes: empty_blockhashes_cid,
+                    lookback_len: fendermint_actor_chainmetadata::DEFAULT_LOOKBACK_LEN,
```
aakoshh, line 246: This could be moved into the genesis file, even if all we do is assign the default there and let them edit it manually.

fendermint/actors/chainmetadata/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright 2021-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+mod actor;
+mod shared;
```
aakoshh, line 4: I don't fully understand how this one is actually _shared_ with the normal Rust code, it all looks like one crate, whereas for example the `evm` actor case [shared](https://github.com/filecoin-project/builtin-actors/tree/master/actors/evm/shared) has its own `Cargo.toml` file.

fendermint/vm/interpreter/src/fvm/exec.rs
fridrik01: Updated to store only the blockhash digest in the state

fendermint/actors/chainmetadata/src/lib.rs
fridrik01: yeah, maybe it was split up in the `evm` actor to reduce code compilation?

fendermint/vm/interpreter/src/fvm/genesis.rs
fridrik01: @aakoshh What about having something like a `custom_actors` field in the genesis file which allows us to more easily configure this (and future) actors more easily, something like:


```
"custom_actors": {
  "chainmetadata": {
     "lookback_len": 256,
     ...
  },
  ...
}
```

But maybe we can do that in a separate PR as this one is getting quite large and we can use the fixed value until then?

fendermint/actors/chainmetadata/src/lib.rs
aakoshh: Yes, if it's split it doesn't (shouldn't?) need reference to certain things like the actor runtime, so we can import just the code to deal with the DTOs and the state for example, but not all the other stuff that the actual actor implementation can require.

fendermint/vm/interpreter/src/fvm/genesis.rs
aakoshh: Sure thing, a separate PR is always fine. 

In the future I imagine that we would want to make the genesis file itself generic so that users can inject some custom bits of handling, have their own actors initialised in the genesis process. I am not sure what's the most ergonomic way of doing that: to make it like a `Genesis<T: DeserializeOwned>` where a certain field is parsed from JSON into their type and given to their handler, or maybe we store stuff in a `BTreeMap<ActorName, Json>` and give that to them to parse, and one of the entires is our `chainmetadata` actor. 

For sure our actor _has_ to be in the genesis file because without it we can't implement the EVM actor completely. With the `Genesis<T>` approach it would have to live separately, with the map it can be one of many. 

Okay, I like that map-like approach in that it forces us to work out the code to support this use case and then it's easier to make it available for users as well.

fendermint/actors/chainmetadata/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+fvm_ipld_blockstore = { workspace = true }
+num-derive = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_tuple = { workspace = true }
+num-traits = { workspace = true }
+frc42_dispatch = { workspace = true }
+anyhow = { workspace = true }
+
+[features]
+fil-actor = []
```
aakoshh, line 26: I don't fully understand this feature. 

At first glance I thought that because of [this conditional](https://github.com/consensus-shipyard/ipc/pull/568/files#diff-74a1adc014625e31a29cc808774f2b2351cef9055d0c1e65e5680c1782c32187R8-R11) thing it will only require the actor runtime if the target is Wasm, which would be cool. But then wouldn't `fil_actors_runtime` have to be an `optional` dependency activated by this feature, similar to how it's in the [testing](https://github.com/consensus-shipyard/ipc/blob/8f165c6db1eec7898ad5725a541f431bc41f0437/fendermint/testing/Cargo.toml#L35) crate where there are lots of optional stuff you might not need, or the [rocksdb](https://github.com/consensus-shipyard/ipc/blob/8f165c6db1eec7898ad5725a541f431bc41f0437/fendermint/rocksdb/Cargo.toml#L29-L30) crate optionally implementing blockstore and kvstore?

fendermint/actors/chainmetadata/src/actor.rs
```diff
@@ -0,0 +1,106 @@
...
+use fil_actors_runtime::ActorDowncast;
+use fil_actors_runtime::ActorError;
+use fil_actors_runtime::Array;
+use fvm_shared::clock::ChainEpoch;
+use fvm_shared::error::ExitCode;
+
+use crate::BlockHash;
+use crate::{ConstructorParams, Method, PushBlockParams, State};
+
+#[cfg(feature = "fil-actor")]
```
aakoshh, line 18: Maybe this can be applied in the `lib.rs` on top of the whole module, which would deactivate the entire `use fil_actors_runtime` thing too, really no implementation at all.

fendermint/actors/chainmetadata/Cargo.toml
aakoshh: I'm not sure if you don't need to declare an empty `default-features` too to make sure by default this is off.

fendermint/actors/chainmetadata/src/actor.rs
```diff
@@ -0,0 +1,106 @@
...
+fil_actors_runtime::wasm_trampoline!(Actor);
+
+pub struct Actor;
+
+impl Actor {
+    fn constructor(rt: &impl Runtime, params: ConstructorParams) -> Result<(), ActorError> {
```
aakoshh, line 24: Just to make sure I understand: I don't think we ever call this constructor, right? We insert the state prepared in the genesis interpreter and move on, unlike with the EVM actors.

fendermint/actors/chainmetadata/src/shared.rs
```diff
@@ -0,0 +1,113 @@
+// Copyright 2021-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use cid::{
+    multihash::{Code, MultihashDigest},
+    Cid,
+};
+use fil_actors_runtime::Array;
```
aakoshh, line 8: Why do we need this `Array` type, is it an alias? Can we use the `fvm_ipld_amt` crate to not need this runtime?

aakoshh: IIRC it provides some helper methods, but we shouldn't need those that badly.

fendermint/actors/chainmetadata/src/shared.rs
```diff
@@ -0,0 +1,113 @@
...
+                    "failed to get blockhash at epoch {}, error: {}",
+                    epoch,
+                    err
+                ));
+            }
+        };
+
+        Ok(Some(Cid::new_v1(
+            DAG_CBOR,
+            Code::Blake2b256.digest(blockhash),
```
aakoshh, line 82: This is going to take the `Blake2b256` hash of the block hash, ie. double hash it. 

What you need to use I believe is `Code::Sha2_256::wrap` instead, or it might be `Sha3_256`. Oh wait, that's if the hash is from CometBFT, if it's from Lotus then it is `Blake2b256`. It looks like we lose that information here, the `code`. 

Hm, well, IIRC `Blake2b256` is the only one supported for IPLD so maybe we can use this here to avoid any surprises, or we can store the `code` with the hash. 

In either case it's the `wrap` function you want, to just put the hash you already have into the data structure.

fendermint/actors/chainmetadata/src/shared.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct ConstructorParams {
+    pub lookback_len: u64,
+}
+
+pub type BlockHash = [u8; 32];
+
+#[derive(Default, Debug, Serialize_tuple, Deserialize_tuple)]
+pub struct PushBlockParams {
+    pub epoch: ChainEpoch,
+    pub block: BlockHash,
```
aakoshh, line 103: Hah, okay I see that the input is a 256 byte hash, not a `Cid`. In that case I'd suggest that the way to resolve this conundrum of how to wrap it into a `Cid` in `State::get_block_cid` is not to do it at all, but rather rename it to `get_block_hash` and return it as-is. We can do the wrapping on the host side. 

Alternatively this should be a `Cid` too, and then this actor should be responsible for taking out the digest and storing it as data. In that case it has the responsibility of dealing with the `code` too, or rejecting if it's not `Blake2b256`, or whatever. 

In any case it should be symmetrical in its input and output.

fendermint/actors/src/manifest.rs
aakoshh: I think we should try to move these IDs to the `fendermint_vm_actor_interface` package along with all the others, and forget the `CODE_ID` part, just make it look like one of the [ipc actors](https://github.com/consensus-shipyard/ipc/blob/8f165c6db1eec7898ad5725a541f431bc41f0437/fendermint/vm/actor_interface/src/ipc.rs#L30-L31): it should have a stable ID and address, but not a code ID. 

The code ID seems like an unstable identifier that might work for a fixed list of builtin actors, but looks like a timebomb for custom actors that might be merged with user defined ones and depend on instantiation order.

fendermint/actors/src/manifest.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Returns the code CID for an actor, given the actor's ID.
+    pub fn code_by_id(&self, id: u32) -> Option<&Cid> {
+        self.by_id.get(&id)
+    }
+
+    /// Returns the the actor code's "id" if it exists. Otherwise, returns 0.
+    pub fn id_by_code(&self, code: &Cid) -> u32 {
+        self.by_code.get(code).copied().unwrap_or(0)
+    }
```
aakoshh, line 72: Let's remove these if we can and use the name instead when creating this actor.

fendermint/app/src/app.rs
```diff
@@ -187,6 +191,7 @@ where
             state_store: Arc::new(state_store),
             multi_engine: Arc::new(MultiEngine::new(1)),
             builtin_actors_bundle: config.builtin_actors_bundle,
+            custom_actors_bundle: config.custom_actors_bundle,
```
aakoshh, line 194: At some point we could follow up with a PR that allows a list here, so users can load their own bundles, all identified by name.

fendermint/vm/interpreter/Cargo.toml
```diff
@@ -43,6 +45,7 @@ fvm_shared = { workspace = true }
 fvm_ipld_blockstore = { workspace = true }
 fvm_ipld_encoding = { workspace = true }
 fvm_ipld_car = { workspace = true }
+fil_actors_runtime = { workspace = true }
```
aakoshh, line 48: Yeah so this shouldn't be needed as we are not running an actor inside the interpreter.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -83,6 +83,36 @@ where
...
+            let params = fvm_ipld_encoding::RawBytes::serialize(
+                fendermint_actor_chainmetadata::PushBlockParams {
+                    epoch: height,
+                    block: block_hash,
+                },
+            )?;
+
+            let msg = FvmMessage {
+                from: system::SYSTEM_ACTOR_ADDR,
+                to: fvm_shared::address::Address::new_id(fendermint_actors::CHAINMETADATA_ACTOR_ID),
```
aakoshh, line 98:
```suggestion
                to: chainmetadata::CHAINMETADATA_ACTOR_ADDR,
```
Since it's a well known one it should conform to the established pattern IMO.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -83,6 +83,36 @@ where
...
+                sequence: height as u64,
```
aakoshh, line 99: How odd, I have no idea why I passed `height` for a nonce to the `cron` actor :thinking:

fendermint/vm/interpreter/src/fvm/externs.rs
```diff
@@ -29,10 +60,60 @@ impl Consensus for FendermintExterns {
...
+impl<DB> Chain for FendermintExterns<DB>
+where
+    DB: Blockstore + Clone + 'static,
+{
+    // for retreiving the tipset_cid, we load the chain metadata actor state
+    // at the given state_root and retrieve the blockhash for the given epoch
+    fn get_tipset_cid(&self, epoch: ChainEpoch) -> anyhow::Result<Cid> {
+        // create a read only state tree from the state root
+        let bstore = ReadOnlyBlockstore::new(&self.blockstore);
+        let state_tree = StateTree::new_from_root(&bstore, &self.state_root)?;
```
aakoshh, line 72: If this was lazily loaded with the `RefCell<Option<...>>` then it would only have to read the root of the state tree once. We can do it as an optimisation later, though.

fendermint/vm/interpreter/src/fvm/genesis.rs
```diff
@@ -228,6 +228,24 @@ where
             )
             .context("failed to create reward actor")?;
 
+        // STAGE 1b: Then we initialize the in-repo custom actors.
+        //
+
+        // Initialize the chain metadata actor which handles saving metadata about the chain
+        // (e.g. block hashes) which we can query.
+        const LOOKBACK_LEN: u64 = 256;
```
aakoshh, line 236: I believe this has already been declared in the actor's module.

fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -97,11 +100,41 @@ where
...
+        let (custom_actor_manifest_version, custom_actor_manifest_data_cid): (u32, Cid) =
+            match store.get_cbor(custom_actors_bundle_root)? {
+                Some(vd) => vd,
+                None => {
+                    return Err(anyhow!(
+                        "no manifest information in custom actor bundle root {}",
+                        bundle_root
+                    ))
+                }
+            };
```
aakoshh, line 124: Can this and the load of the previous manifest root be moved into a utility function?

fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -174,6 +206,35 @@ where
...
+        code_id: u32,
+        id: ActorID,
+        state: &impl Serialize,
+        balance: TokenAmount,
+        delegated_address: Option<Address>,
+    ) -> anyhow::Result<()> {
+        // Retrieve the CID of the actor code by the numeric ID.
+        let code_cid = *self
+            .custom_actor_manifest
+            .code_by_id(code_id)
```
aakoshh, line 223:
```suggestion
            .code_by_name(actor_name)
```

aakoshh (commented):
Nice one! That `Clone` was a viral change, wasn't it.

So my main observations are:
* Let's create an entry under the `actor_interface` to fit in with the others, just a re-export of stuff. 
* Let's forget the `code_id` for these, it's just an index in a random list of names - just use the names. We can verify the names are unique when constructing the manifest. 
* Let's not import the actor runtime into the host. 

This could really benefit from a test contract and an integration test similar to the `Greeter.sol` one that deploys and calls the contract - it could have a method that returns the hash of a block, which could be compared against what the CometBFT API returns. That would highlight the double hashing.

fendermint/actors/chainmetadata/src/actor.rs
fridrik01: Right, we actually only call `PushBlock` method of this actor, everything else (constructing, and lookup up blockhash by epoch in the get_tipset_cid) is done from the outside.

fendermint/actors/chainmetadata/src/shared.rs
fridrik01: its an alias, should be easy to replace

fendermint/actors/chainmetadata/Cargo.toml
fridrik01: I made the `fil_actors_runtime` a conditional dependency when fil-actor is set

fendermint/vm/interpreter/src/fvm/externs.rs
fridrik01: Will create a follow up task

fendermint/actors/src/manifest.rs
fridrik01: I created a `chainmetadata` module in the `fendermint_vm_actor_interface` package where I call `define_id` macro to generate the CHAINMETADATA_ID/CHAINMETADATA_ACTOR.

Funnily, I accidentally set its id ti 64, which caused the graph-test to fail with a `failed to create checkpoint` error which took me a while to figure out as I had changed so several other things at the same time. 

So +1 for having all of these IDs in `lib.rs` so to not accidentally use the same ID twice, which when happens silently gets accepted.

fendermint/actors/chainmetadata/src/shared.rs
fridrik01: Fixed by returning a blockhash in this method and renaming it to `get_block_hash`

fendermint/actors/chainmetadata/src/shared.rs
fridrik01: Yeah, that makes sense. I'll rename the method to `get_block_hash` and let it return a `BlockHash`, and then do the conversion to `Cid` in the `tip_tipset_cid`. I verified via manual testing that the blockhash at given epoch in `get_tipset_cid` extern was the same which was added in `begin`. But it still needs proper testing

fendermint/vm/interpreter/src/fvm/exec.rs
fridrik01: Should I set this to 0?

fridrik01:
@aakoshh thanks for the detailed review! I think I have addressed most/all you comments

fendermint/actors/src/manifest.rs
```diff
@@ -0,0 +1,55 @@
...
+        let mut code_by_name = HashMap::new();
+        for (name, code_cid) in iter.into_iter() {
+            code_by_name.insert(name.into(), code_cid);
+        }
+
+        // loop over required actors and ensure they are present
+        for &name in REQUIRED_ACTORS.iter() {
+            let _ = code_by_name
+                .get(name)
+                .context(format!("manifest missing required actor {}", name))?;
```
aakoshh, line 46:
```suggestion
                .with_context(|| format!("manifest missing required actor {}", name))?;
```
That version only produces the interpolated string if there is an actual error.

fendermint/vm/interpreter/src/fvm/exec.rs
aakoshh: It's fine, I don't know why I set this, but it should work either way. I don't think it's checked for implicit calls.

fendermint/vm/interpreter/src/fvm/externs.rs
```diff
@@ -29,10 +63,60 @@ impl Consensus for FendermintExterns {
...
+                    return Err(anyhow!(
+                        "failed to get chain metadata actor ({}) state, error: {}",
+                        CHAINMETADATA_ACTOR_ID,
+                        err
+                    ));
+                }
+            };
+
+        match actor_state.get_block_hash(&bstore, epoch) {
+            Ok(Some(v)) => Ok(Cid::new_v1(DAG_CBOR, Code::Blake2b256.digest(&v))),
```
aakoshh, line 115:
```suggestion
            Ok(Some(v)) => Ok(Cid::new_v1(DAG_CBOR, Code::Blake2b256.wrap(&v))),
```

`.digest` would take the hash of the hash.

fridrik01: Ahhh

fendermint/vm/interpreter/src/fvm/externs.rs
```diff
@@ -29,10 +63,63 @@ impl Consensus for FendermintExterns {
...
+                    return Err(anyhow!(
+                        "failed to get chain metadata actor ({}) state, error: {}",
+                        CHAINMETADATA_ACTOR_ID,
+                        err
+                    ));
+                }
+            };
+
+        match actor_state.get_block_hash(&bstore, epoch) {
+            Ok(Some(v)) => match Code::Blake2b256.wrap(&v) {
```
aakoshh, line 115: Maybe worth leaving a comment that this hash might not be `Blake2b256`. Like in our case it's actually `Sha2_256`, because it's a Tendermint block hash. We just leave it as it is because we don't know if it would upset the FVM if the code was something it did not expect.

fridrik01:
Working on fixing CI, there was some issue when rebasing, and now getting cargo errors failing to select a version for a package

fendermint/testing/contracts/Greeter.sol
```diff
@@ -15,7 +15,10 @@ contract Greeter {
 
     //This function returns the current value stored in greeting variable
     function greet() public view returns (string memory) {
-        return greeting;
+        //uint256 number = uint256(blockhash(block.number - 1));
+        bytes32 myHash = blockhash(block.number - 1);
+
+        return string(abi.encodePacked(myHash));
```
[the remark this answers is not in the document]
fridrik01: sorry, this was accidentally committed, it was just what I used for testing

Cargo.toml
```diff
@@ -186,9 +192,8 @@ cid = { version = "0.10.1", default-features = false, features = [
   "std",
 ] }
 
-# Depending on the release cycle, this dependency might want an earlier version of the FVM.
-# We can work around it by hardcoding the method hashes; currently there is only one.
-# frc42_dispatch = "3.2"
+# We need a version of frc42_dispatch that has been updated to use FVM 4.1
+frc42_dispatch = { git = "https://github.com/fridrik01/filecoin.git", branch = "update-fvm-4.1" }
```
fridrik01, line 196: Note, that I needed to fork this repo to update it to FVM 4.1. I have a PR for merging this into the official repo (https://github.com/helix-onchain/filecoin/pull/231)

fendermint/testing/contracts/Greeter.bin
```diff
@@ -1 +1 @@
...
...ffffffff826103f6565b61043b86836103f6565b95508019841693508086168417925050509392505050565b5f819050919050565b5f819050919050565b5f61047f61047a61047584610453565b61045c565b610453565b9050919050565b5f819050919050565b61049883610465565b6104ac6104a482610486565b848454610402565b825550505050565b5f90565b6104c06104b4565b6104cb81848461048f565b505050565b5b818110156104ee576104e35f826104b8565b6001810190506104d1565b5050565b601f82111561053357610504816103d5565b61050d846103e7565b8101602085101561051c578190505b610530610528856103e7565b8301826104d0565b50505b505050565b5f82821c905092915050565b5f6105535f1984600802610538565b1980831691505092915050565b5f61056b8383610544565b9150826002028217905092915050565b610584826102de565b67ffffffffffffffff81111561059d5761059c610173565b5b6105a782546103a5565b6105b28282856104f2565b5f60209050601f8311600181146105e3575f84156105d1578287015190505b6105db8582610560565b865550610642565b601f1984166105f1866103d5565b5f5b82811015610618578489015182556001820191506020850194506020810190506105f3565b868310156106355784890151610631601f891682610544565b8355505b6001600288020188555050505b50505050505056fea264697066735822122016f2baf93879d1f50483aa9b26b1323ba354fd1a58899568601512441cffe1c364736f6c63430008170033
```
aakoshh, line 1: Should these `.bin` files change? CI passes so maybe it's just a compiler version thing, but just making sure it's not your test version being committed without the source changes which you removed from the PR?

aakoshh (approved):
Great! I'm not sure why `Greeter.bin` changes, but as long as CI passes, let's go!

fendermint/testing/contracts/Greeter.bin
fridrik01: I have solc version 0.8.23 which seems to result in a slightly different artifacts, but yeah its compiled from the original contract so it should be good

PR #569, opened 2024-01-15
CHORE: Fix after IPLD update

aakoshh:
Some fixes after `libipld` to upgraded in https://github.com/consensus-shipyard/ipc/pull/542


fridrik01 (approved):
LGTM, 

So this basically was a compile error when running the smoke test in ipld/resolver, I am  wondering why this was not caught by CI in https://github.com/consensus-shipyard/ipc/pull/542, did I just miss it? or  maybe it didn't run because now we use workspace dependencies so bumping versions there do not trigger file changes in subcrates based on the GH workflows. If so, we probably need to reevaluate when to run github actions.

aakoshh:
@fridrik01 yes that's right, it wasn't executed because nothing changed inside `ipld/resolver`, only the top level `Cargo.toml`. It _should_ be rare that it causes anything to break and it will obviously be caught on `main` where everything runs all the time, so it's a tradeoff between earlier warning and longer builds whenever one of the crates adds a new dependency or something similarly benign happens. You can always check by running a `make lint` in the root.

PR #571, opened 2024-01-16
Update bottomup creation

cryptoAtwill:
This PR adds the bottom up checkpoint creation with message batching. 

The idea is as follows, when there is a new bottom up message created at the child, it will be committed into the message batch of the NEXT BOTTOM UP CHECKPOINT PERIOD. If the batch is full, i.e. more than a certain threshold, there is a batch cut at the current block height and a new batch will be created at NEXT BOTTOM UP CHECKPOINT PERIOD.

Previously, when a bottom up checkpoint is created, it checks if the checkpoint period has reached. But now, this will miss certain batch cut. In this case, the new flow should be:
- Check if there is a msg batch at the current height in the child
- Check if the bottom up checkpoint period has reached
The rest of the process follows the previous version, hence no change.


fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -63,70 +63,85 @@ where
...
-
-                debug_assert_eq!(next_power_configuration_number, next_configuration_number);
-
-                power_diff(curr_power_table, next_power_table)
-            };
-
-            Ok(Some((checkpoint, power_updates)))
+    let batch = gateway.bottom_up_msg_batch(state, height.into())?;
+
+    if batch.block_height.as_u64() == 0 {
```
maciejwitowski, line 68: Nit: after this change the comment on this function `Construct and store a checkpoint if this is the end of the checkpoint period.` should probably be updated.

PR #573, opened 2024-01-16
fix: short option names must be unique in fendermint cli

fridrik01:
This PR fixes a s collision in our cli parser where the '-l' short option was duplicated between `log_dir` and `log_level`. I renamed the short option of the former to `-d`

Test

```
> fendermint -h
thread 'main' panicked at /home/fridrik/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.14/src/builder/debug_asserts.rs:112:17:
Command fendermint_app_options: Short option names must be unique for each argument, but '-l' is in use by both 'log_dir' and 'log_level'
```

### DIFF
--- a/fendermint/app/options/src/lib.rs
+++ b/fendermint/app/options/src/lib.rs
@@ -76,7 +76,7 @@ pub struct Options {
     pub home_dir: PathBuf,
 
     /// Set a custom directory for ipc log files.
-    #[arg(short = 'l', long, env = "FM_LOG_DIR")]
+    #[arg(short = 'd', long, env = "FM_LOG_DIR")]
     pub log_dir: Option<PathBuf>,
 
     /// Optionally override the default configuration.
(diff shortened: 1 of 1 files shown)


aakoshh:
Great, thanks! 

Maybe this would be a good opportunity to add a test for it too. [There is one](https://github.com/consensus-shipyard/ipc/blob/57099cf0b04dda988030162b4b22e47d7aada548/fendermint/app/options/src/lib.rs#L145-L151) that parses the `GlobalOptions` with `--help`, but as noted in the comment if we try it with `Options` it would quit the process, so the test should be forked, or I'm not sure how else it could be conducted, but it would be great if we didn't rely on someone trying this occasionally, now that we see it can sneak up on us.

aakoshh:
Ah, easy, there should be a [try_parse_from](https://github.com/clap-rs/clap/blob/e648e086f3934afb40b121b5999b9e23657ddc28/clap_builder/src/derive.rs#L60-L73) which doesn't exit the test framework.

fendermint/app/options/src/lib.rs
```diff
@@ -76,15 +76,15 @@ pub struct Options {
     pub home_dir: PathBuf,
 
     /// Set a custom directory for ipc log files.
-    #[arg(short = 'l', long, env = "FM_LOG_DIR")]
+    #[arg(short = 'd', long, env = "FM_LOG_DIR")]
```
aakoshh, line 79: This is already used for `home_dir`

fridrik01: oh wow, I am blind I missed it completely :D

aakoshh (approved):
Cheers :pray:

PR #574, opened 2024-01-16
BUILD: Run tests if Cargo.toml changes

aakoshh:
Changes `ci.yaml` to run all Rust workflows if `Cargo.toml` changes, so that if we change some dependency version we don't have to wait until the PR is merged into `master` to see any build failures, which might be only due to `clippy`, as it happened in https://github.com/consensus-shipyard/ipc/pull/569

This is a tradeoff between doing more work with potentially no reason vs fixing stuff on `main`.


fridrik01 (approved):
Nice!

PR #577, opened 2024-01-16
IPC-576: Initialise vote tally

### LINKED ISSUE #576: Initialize tally with current committee

aakoshh:
Closes #576 

The vote tally will be needed in the `ChainInterpreter` similar to how we use the IPC finality provider. We might not need it for queries (haven't thought about how to update the provider yet), but we do have to pass the updated power table to it. Since the `ChainInterpreter` already had two extra pieces in its input compared to what it passed on to the inner interpreter (which is only interested in the execution state), I created a new `ChainEnv` struct to hold these pieces, and extended it with the vote tally. 

The vote tally has a similar initialisation story as the IPC finality provider: it needs the latest finalized block, but it can't get it from the ledger until `init_chain` (ie. genesis) is completed. For this reason I added an extra query to the startup routine of the syncer and set these together as an atomic operation. The syncer will have to maintain parts of this tally anyway, so I think it's fine to group these operations together.

Until the tally is initialised with the last finalized block it will reject attempts to extend the chain.


fridrik01 (approved):
:+1:

PR #584, opened 2024-01-16
IPC-583: Use the vote tally in prepare proposal

### LINKED ISSUE #583: Refactor prepare_proposal to consult the vote tally

aakoshh:
Closes #583 

Changes `ChainInterpreter::prepare` so that it looks at both the vote tally and the current parent finality provider: 
1. asks the vote tally if there is a quorum over any height in the parent chain
2. if there is, asks the provider for its next proposal
3. proposes the minimum of the two

This way we go as little as possible forward in the chain: only as much as there is quorum for, but potentially less if some of the existing limits are hit.

This way we don't have to change the logic of the existing provider, we are strictly just gatekeeping it from proposing too early.

### DIFF
--- a/fendermint/vm/interpreter/src/chain.rs
+++ b/fendermint/vm/interpreter/src/chain.rs
@@ -121,6 +121,9 @@ where
         });
 
         // Prepare top down proposals.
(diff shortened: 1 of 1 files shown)


fridrik01 (approved):
lgtm

PR #587, opened 2024-01-17
IPC-586: Subscribe to votes and add them to the tally

### LINKED ISSUE #586: Subscribe to the voting channel

aakoshh:
Closes #586 

Add subscription and event handling.

### DIFF
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use anyhow::{anyhow, bail, Context};
+use async_stm::atomically_or_err;
 use fendermint_abci::ApplicationService;
 use fendermint_app::ipc::{AppParentFinalityQuery, AppVote};
 use fendermint_app::{App, AppConfig, AppStore, BitswapBlockstore};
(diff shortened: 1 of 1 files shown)


fridrik01:
@aakoshh I think the PR includes some commits from previous PRs

aakoshh:
> I think the PR includes some commits from previous PRs
Yes, I had to rebase after merging https://github.com/consensus-shipyard/ipc/pull/577 which this PR was targetting. 

The remaining ones are independent :relieved:

fendermint/app/src/cmd/run.rs
```diff
@@ -375,3 +387,60 @@ fn to_address(sk: &SecretKey, kind: &AccountKind) -> anyhow::Result<Address> {
...
+
+async fn dispatch_vote(
+    vote: VoteRecord<AppVote>,
+    parent_finality_votes: &VoteTally,
+    topdown_enabled: bool,
+) {
+    match vote.content {
+        AppVote::ParentFinality(f) => {
+            if !topdown_enabled {
+                tracing::debug!("ignoring vote; topdown disabled");
```
fridrik01, line 423: is this maybe too verbose as its inside a loop?

fridrik01 (approved):
:+1:

fendermint/app/src/cmd/run.rs
aakoshh: Yeah, if we saw this it would be because:
1. We are the root subnet, but we are using the IPLD resolver to listen to things from one of the subnets we are interested in
2. Yes, somehow we are receiving votes about the finality of of our non-existing parent. Can this be someone trying to flood the system with spam?

I certainly didn't want to give the ability to someone to generate warnings and keep everyone up, but I thought if these arrive we should know because maybe we misconfigured something and we are ignoring these votes by mistake (more likely). It doesn't give any indication of who we should ban, so it's more like a "hey, this is very unexpected!" flag, with `debug` level in case we are already trying to figure out what's going on.

All I can think of is some kind of global bool to only warn once, but then it could be drowned out by other stuff. Maybe this way, if our node misses a handful of votes every 30 seconds, we can have an indication of why things aren't working. 

If they turn out to be spammy we should remove them.

PR #594, opened 2024-01-18
CHORE: Substitute canonical apache licence

jsoares:
The current apache licence is not in canonical form and is not recognised by GH for repo licensing purposes:

image

This just replaces the text with the standard one.


aakoshh (approved):
Thanks :judge:

PR #599, opened 2024-01-19
IPC-532: Implementation of abstract IPC Contract + various fixes

### LINKED ISSUE #532: GMP: `IpcContract` abstract contract and `IpcEntrypoint` core implementation
Implement the abstract contract and common logic for all IPC-enabled contracts. This will be used as the "sdk" for contracts looking to support the execution of cross-net message. Design details [here](https://www.notion.so/pl-strflt/General-Message-Passing-7686edb5b65c4f9086fa4d4288c35906?pvs=4#8047dfc79bcb4e0cb8ea12f2e4fe2bdd).

adlrocha:
Closes #532 

This PR implements the IPC abstract contract that should be implemented by IPC-enabled contracts. This implementation lacks unit tests. Ideally, we should already implement an example contract to showcase how to use this contract and make a contract IPC-enabled.

See also the changes introduced in #606.


contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+    // in-flight messages were received.
+    mapping(uint256 => bytes32[]) public inFlightMsgHeights;
+    // last height where there messages were pruned
+    uint256 private lastGcBlock;
+
+    // The adderss of the gateway in the subnet.
+    address public gatewayAddr;
+    /// Number of blocks that the contract will wait for a receipt
+    /// before garbage collecting the in-flight message.
+    uint256 public timeoutBlocks;
```
raulk, line 30: - A timeout is a simple solution but it's dangerous in the context of distributed transactions.
- IPC as a framework enforces no timeouts, so it's possible for us to drop a message here (and potentially roll back any local actions, e.g. return funds to the user) only for an ACK to be delivered the epoch later. This would throw the contract/dapp into an inconsistent state, with potential double-booking of funds and no way to amend the state.
- Dropping a message on timeout may require a local rollback action, which currently this interface does not support.
- State ballooning, whilst undesirable, only affects long-term gas costs in FEVM. Freeing state does not refund gas in FEVM (unlike the EVM). While it _does_ make the amortized cost of writing lower over time (due to lower HAMT depth), we can regard that purely as an optimization. In fact, we need to store a whole entirely mapping _only_ to do timeout tracking, so I'll go out on a limb and say that timeout tracking may even be counterproductive cost-wise.
- The GC logic, as currently implemented, puts the cost of clearing the map on the first entrypoint call for a non-empty epoch, which is unfair.

Therefore:
1. I am going to take the liberty to remove the timeout machinery here.
2. Instead, I'll add a method for the owner to explicitly _drop_ messages by hash. This allows the owner to run whatever local logic they want if they really care about dropping potentially expired messages.
3. By making this explicit, it also forces the contract author to think about rollback effects that may be associated by declaring a message dead.

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+import {IpcEnvelope, ReceiptMsg, IpcMsg, IpcMsgKind} from "../src/structs/CrossNet.sol";
+import {EMPTY_BYTES} from "../src/constants/Constants.sol";
+import {IGateway} from "../src/interfaces/IGateway.sol";
+
+// Interface that needs to be implemented by IPC-enabled contracts.
+// This is really convenient to call it from other contracts.
+interface IfaceIpcContract {
+    /// @notice Entrypoint for cross-net messages in IPC-enabled contracts.
+    // solhint-disable-next-line func-name-mixedcase
+    function IpcEntrypoint(IpcEnvelope calldata crossMsg) external payable returns (bytes memory ret);
```
raulk, line 13: Why not use lower camel case to follow Solidity conventions?

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+        // in case any timeout has triggered already.
+        // timeout == 0 means that no in-flight expiraton is set for receipts.
+        if (timeoutBlocks != 0) {
+            inFlightGC();
+        }
+
+        // only the gateway address is allowed to deliver cross-net messages.
+        if (msg.sender != gatewayAddr) {
+            revert CallerIsNotGateway();
+        }
```
raulk, line 56: Should extract to a modifier.

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+
+        // internal dispatch of the cross-net message to the right method.
+        if (crossMsg.kind == IpcMsgKind.Call) {
+            IpcMsg memory callMsg = abi.decode(crossMsg.message, (IpcMsg));
+
+            (bool success, bytes memory ret) = address(this).delegatecall(
+                abi.encodeWithSelector(callMsg.method, callMsg.params)
```
raulk, line 63: This is not good, we're back to square one where the callee has lost information about the call (e.g. sender address).

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+        // only the gateway address is allowed to deliver cross-net messages.
+        if (msg.sender != gatewayAddr) {
+            revert CallerIsNotGateway();
+        }
+
+        // internal dispatch of the cross-net message to the right method.
+        if (crossMsg.kind == IpcMsgKind.Call) {
+            IpcMsg memory callMsg = abi.decode(crossMsg.message, (IpcMsg));
+
+            (bool success, bytes memory ret) = address(this).delegatecall(
```
raulk, line 62: Instead of a delegate call, we should just dispatch to an abstract IPC-aware handler method.

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+                abi.encodeWithSelector(callMsg.method, callMsg.params)
+            );
+            if (!success) {
+                revert IpcMsgCallFailed();
+            }
```
raulk, line 67: This occludes the actual error!

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+            IpcEnvelope memory originalMsg = inFlightMsgs[ackMsg.id];
+            if (
+                originalMsg.message.length == 0 ||
+                keccak256(abi.encode(crossMsg.from)) != keccak256(abi.encode(originalMsg.to))
+            ) {
+                revert UnrecognizedReceipt();
+            }
+            // process ACK message and remove from in-flight if it succeeds.
+            if (AckEntrypoint(ackMsg)) {
+                delete inFlightMsgs[ackMsg.id];
```
raulk, line 86: Calling this ACK is not correct. It carries information (return data), and it can be an error.

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        return EMPTY_BYTES;
+    }
+
+    /// @notice Function that needs to be called internally by the contract to send a
+    /// cross-net message.
+    /// TODO: Consider a more Solidity-friendly interface instead of having to pass
+    /// the full envelope?
+    function XnetCall(IpcEnvelope memory crossMsg) internal virtual {
```
raulk, line 99: I don't understand why this is `virtual`?

raulk (dismissed):

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,151 @@
...
+            // sketchy.
+            IpcEnvelope memory originalMsg = inFlightMsgs[ackMsg.id];
+            if (
+                originalMsg.message.length == 0 ||
+                keccak256(abi.encode(crossMsg.from)) != keccak256(abi.encode(originalMsg.to))
+            ) {
+                revert UnrecognizedReceipt();
+            }
+            // process ACK message and remove from in-flight if it succeeds.
+            if (AckEntrypoint(ackMsg)) {
```
raulk, line 85: So here we would have correctly correlated the response, but handling the result can fail. Even in that case, we should remove the entry from the correlation map because we won't receive a second response.

contracts/sdk/IpcContract.sol
raulk: Solved in #606, merged here.

contracts/examples/contracts/cross-token/ERC20TokenMessenger.sol
raulk, line 1: NOTE: I think this is not being compiled in CI, that's why we didn't notice that the previous code would not compile.

raulk:
Moving out of draft to run the full CI.

snissn (approved):
looks good to me!

contracts/sdk/IpcContract.sol
```diff
@@ -0,0 +1,112 @@
...
+    /// NOTE: This must not revert as doing so will leave the correlation map in an inconsistent state.
+    /// (IPC will consider the result delivery attempted, and will not repeat it again).
+    function _handleIpcResult(IpcEnvelope storage original, IpcEnvelope memory result, ResultMsg memory resultMsg) internal virtual;
+
+    /// @notice Method the implementation of this contract can invoke to perform an IPC call.
+    function performIpcCall(IPCAddress calldata to, CallMsg calldata callMsg, uint256 value) internal virtual {
+        // Queue the cross-net message for dispatch.
+        IpcEnvelope memory envelope = IGateway(gatewayAddr).sendContractXnetMessage{value: value}(IpcEnvelope({
+            kind: IpcMsgKind.Call,
+            from: to, // TODO: will anyway be replaced by sendContractXnetMessage.
```
cryptoAtwill, line 80: just a thought, maybe we can create a `callXnetContract` method that hides `kind`, `from` and `nonce`, so that don't need the TODO comment?

contracts/src/gateway/GatewayMessengerFacet.sol
```diff
@@ -20,39 +21,59 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
...
+        // Will revert if the message won't deserialize into a CallMsg.
+        abi.decode(envelope.message, (CallMsg));
+
+        committed = IpcEnvelope({
+            kind: IpcMsgKind.Call,
+            from: IPCAddress({subnetId: s.networkName, rawAddress: FvmAddressHelper.from(msg.sender)}),
+            to: envelope.to,
+            value: msg.value,
+            message: envelope.message,
+            nonce: 0
```
cryptoAtwill, line 64:
```suggestion
            nonce: 0   // nonce will be updated by LibGateway.commitCrossMessage
```

cryptoAtwill (approved):

contracts/sdk/IpcContract.sol
raulk: 100% — I captured this as a TODO in #613. I already got started on this but the PR was becoming too heavy. We can address later.

raulk: (We can also remove the `value`)

PR #600, opened 2024-01-19
IPC-541: Gossip finality votes

### LINKED ISSUE #541: Gossip votes about top-down finality
Currently top-down finality proposals work by a validator proposing to the others what they think is final on the parent chain, and then hope that they see it as final as well. If they don't, they won't vote on the proposed block, which negatively affects CometBFT liveness.

Our other ace in the sleeve was to use the voting mechanism added to the IPLD Resolver to first agree on the finality off-chain, and then propose knowing that it will pass the bar. We could add a quorum certificate to the proposal, but the minimal solution is to keep the block voting as it is, and just use the IPLD Resolver to prevent introducing proposals that are bound to fail.

Votes would be cast on `(BlockHeight, BlockHash)` pairs. This needs an extension in the library to allow generic voting records; currently it's just a CID you can vote on.

Votes will be cast on different heights by different nodes; to make it easier to find common ground, we will try to find the Last Common Ancestor among the different votes; in other words a vote on a block is implicitly a vote on of its ancestors, so if Alice votes on B100 as final, and Bob votes on B105 as final, and Alice knows about blocks 101-105, just doesn't see them as deep enough yet, she can still use Bob's vote as support for B100, because she knows it's an ancestor of B105.

The goal here is to refactor the top-down proposals to be based on this voting mechanism going on in the background, and to add the necessary calls to cast votes.

Nodes should only vote on things they truly perceive as final; there shouldn't be a reason to revoke votes due to reorgs for example, nor is there a mechanism to do so.

```
### Tasks
- [ ] https://github.com/consensus-shipyard/ipc/issues/544
- [ ] https://github.com/consensus-shipyard/ipc/issues/550
- [ ] https://github.com/consensus-shipyard/ipc/issues/576
- [ ] https://github.com/consensus-shipyard/ipc/issues/586
- [ ] https://github.com/consensus-shipyard/ipc/issues/585
- [ ] https://github.com/consensus-shipyard/ipc/issues/583
- [ ] https://github.com/consensus-shipyard/ipc/issues/579
- [ ] https://github.com/consensus-shipyard/ipc/issues/578
- [ ] https://github.com/consensus-shipyard/ipc/issues/612
```

aakoshh:
Closes #541 

(This is the feature branch, the culmination of many smaller PRs in the epic. It's easier to review them individually then this one).

Closes ENG-509.
Closes ENG-559.


fendermint/app/src/cmd/run.rs
```diff
@@ -367,3 +418,60 @@ fn to_address(sk: &SecretKey, kind: &AccountKind) -> anyhow::Result<Address> {
...
+    loop {
+        match rx.recv().await {
+            Ok(event) => match event {
+                ResolverEvent::ReceivedPreemptive(_, _) => {}
+                ResolverEvent::ReceivedVote(vote) => {
+                    dispatch_vote(*vote, &parent_finality_votes, topdown_enabled).await;
+                }
+            },
+            Err(RecvError::Lagged(n)) => {
+                tracing::warn!("the resolver service skipped {n} gossip events")
```
raulk, line 436: I think eventually we might want to deliver this to the consumer of the IPLD resolver, since what to do in this case might be business-specific.

fendermint/app/src/cmd/run.rs
```diff
@@ -367,3 +418,60 @@ fn to_address(sk: &SecretKey, kind: &AccountKind) -> anyhow::Result<Address> {
...
+            })
+            .await;
+
+            match res {
+                Ok(_) => {}
+                Err(e @ VoteError::Equivocation(_, _, _, _)) => {
+                    tracing::warn!(error = e.to_string(), "failed to handle vote");
+                }
+                Err(e) => {
+                    tracing::debug!(error = e.to_string(), "failed to handle vote")
```
raulk, line 472: Why is this debug? Are general errors normal here?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,375 @@
...
+            let Some(votes_for_block) = votes_at_height.get(block_hash) else {
+                continue; // We could detect equovicating voters here.
+            };
+
+            for vk in votes_for_block {
+                if voters.insert(vk.clone()).is_none() {
+                    // New voter, get their current weight; it might be 0 if they have been removed.
+                    weight += power_table.get(vk).cloned().unwrap_or_default();
+                }
+            }
```
raulk, line 260: Do validators gossip about every single block they've observed? What if a validator goes down or has their parent sync loop interrupted, and then syncs back to parent HEAD in a batch? e.g. going for epoch 100 to 200 immediately. Will it send 100 gossip messages? (which may be considered spamming) Or will sending a vote for the latest implicitly vote for the rest of the chain? I think we should do the latter, but it's not clear to me if that's what's implemented here.

raulk (approved):

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,375 @@
...
+            let has_power = vote_tally.has_power(&validator_key)?;
+
+            if has_power {
+                // Add our own vote to the tally directly rather than expecting a message from the gossip channel.
+                // (I'm not messages published by this node would be delivered to it, so this might be the only way).
+                if let Err(StmError::Control(c)) =
+                    vote_tally.add_vote(validator_key.clone(), next_height, next_hash.clone())
+                {
+                    // We should not see any other error from this as we just checked that the validator had power.
+                    return Err(c);
```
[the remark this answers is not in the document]
raulk: IUC this will stop the loop, and will degrade the system and shortly thereafter it'll stall. Is this completely unrecoverable? And if so, should we panic/abort instead?

fendermint/app/src/cmd/run.rs
aakoshh: Yeah this is a grey area, I should add a comment. I didn't want to handle all possible errors in the enum to keep the noise down here, but maybe that would save some head cratching as well as force us to pick levels for any future errors:
* `Unintitialized` we can safely ignore, there will be other votes once the system is up and running; unless it keeps repeating and we're wondering what's happening, in which case during debug it is helpful to realise that initialisation has failed
* `UnpoweredValidator` can happen if we get a vote from someone who isn't a validator according to our view, but they could have been one in a previous block, or they might be one in the next block; or they might truly just publish trolling votes. Unless there is a way to identify vandals, I wouldn't give them the power to spam the logs, but in debug it can help to see this is happening.
* `UnexpectedBlock` will not be returned by this method.

fendermint/app/src/cmd/run.rs
aakoshh: Could be; for now I just wanted us to know that we are missing events because something is very slow.

fendermint/vm/topdown/src/voting.rs
aakoshh: Great question; the mechanism that should prevent us from spamming everyone with votes on blocks in the past is the the [TendermintAwareSyncer](https://github.com/consensus-shipyard/ipc/blob/335748d945066123629f60285e2890e34bfbaa82/fendermint/vm/topdown/src/sync/tendermint.rs#L30-L31) which checks if CometBFT is syncing with the network. If it is, it skips syncing with the parent in that round altogether, expecting to see final decisions in the blocks instead. 

Because the parent syncer doesn't run, there are no blocks added by it to the tally either, so we won't vote on anything.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,375 @@
...
+        if !self.has_power(&validator_key)? {
+            return abort(Error::UnpoweredValidator(validator_key));
+        }
+
+        let mut votes = self.votes.read_clone()?;
+        let votes_at_height = votes.entry(block_height).or_default();
+
+        for (bh, vs) in votes_at_height.iter() {
+            if *bh != block_hash && vs.contains(&validator_key) {
+                return abort(Error::Equivocation(
```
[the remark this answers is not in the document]
aakoshh: That's exactly right, this equivocation indicates either that the sender is malicious or that they (and maybe everyone else) have configured their system with too low embedding threshold. It's impossible to say currently. 

I also mentioned to Fridrik that I added this as a low hanging fruit of a clear evidence of a Byzantine node, however there are other forms of equivocations I ignore: for example when looking for quorum I could deduce that you are supporting a block which is not an ancestor of a block with a higher block number you already voted for in my view, that is, you are voting for different forks, albeit at different heights. 

There is potential but unless we want to do something with these, I left it at this obvious one to warn about the clearest of all cases.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,375 @@
...
+        let votes = self.votes.read()?;
+        let power_table = self.power_table.read()?;
+
+        let mut weight = 0;
+        let mut voters = im::HashSet::new();
+
+        for (block_height, block_hash) in chain.iter().rev() {
+            if block_height == finalized_height {
+                break;
+            }
```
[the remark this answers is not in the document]
raulk: We could also skip one before we reverse the iterator, but that might be spooky.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,375 @@
...
+            Some(weight) => Ok(*weight > 0),
+        }
+    }
+
+    /// Calculate the minimum weight needed for a proposal to pass with the current membership.
+    pub fn quorum_threshold(&self) -> Stm<Weight> {
+        let total_weight: Weight = self.power_table.read().map(|pt| pt.values().sum())?;
+
+        Ok(total_weight * 2 / 3)
+    }
```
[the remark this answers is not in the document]
aakoshh: I'll check how CometBFT works, which is ultimately the bar we have to pass with our proposal that this lets through.

ipc/api/src/subnet_id.rs
```diff
@@ -214,13 +214,29 @@ impl Default for SubnetID {
...
+
+        let segments: Vec<&str> = id.split('/').skip(1).collect();
+
+        let root = segments[0][1..]
+            .parse::<u64>()
+            .map_err(|_| Error::InvalidID(id.into(), "invalid root ID".into()))?;
+
+        let mut children = Vec::new();
+
+        for addr in segments[1..].iter() {
```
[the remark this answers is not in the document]
aakoshh: Thanks, I'll add a test (and also move the others tests I added to the settings, they really belong here, but they are from the time when this was in a separate repo). This is how it was before, so I was hoping that `1..` just skips one and returns an empty iterator, but I am not actually sure. It's definitely not an error not to have children.

fendermint/vm/topdown/src/voting.rs
aakoshh: If the parent were CometBFT with a 1s block time, we should probably avoid gossiping on each height!

> syncs back to parent HEAD in a batch? e.g. going for epoch 100 to 200 immediately. Will it send 100 gossip messages?

That depends on whether CometBFT itself lost its sync with the subnet chain: https://docs.cometbft.com/v0.37/core/block-sync

It has fast sync, but on the bottom of the page that going _back_ to fast sync is an open issue.

cryptoAtwill (approved):

ipc/api/src/subnet_id.rs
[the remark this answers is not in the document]
aakoshh: The test to parse just a root ID [exists](https://github.com/consensus-shipyard/ipc/blob/a406f692b3d455c79bf13f2f7e4f36ee337ac0c7/ipc/api/src/subnet_id.rs#L259-L262), so it should be fine.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -0,0 +1,375 @@
...
+            let next_hash = match vote_tally.block_hash(next_height)? {
+                Some(next_hash) => next_hash,
+                None => retry()?,
+            };
+
+            let has_power = vote_tally.has_power(&validator_key)?;
+
+            if has_power {
+                // Add our own vote to the tally directly rather than expecting a message from the gossip channel.
+                // (I'm not messages published by this node would be delivered to it, so this might be the only way).
```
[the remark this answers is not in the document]
aakoshh: https://linear.app/interplanetary-consensus/issue/ENG-622/investigate-gossip-behaviour-do-we-get-our-own

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -215,6 +231,7 @@ where
             .get_top_down_msgs(height)
             .await
             .map_err(|e| Error::CannotQueryParent(e.to_string(), height))?;
+
```
cryptoAtwill, line 234: if we detected a chain reorg, do we have to clear existing votes?

fendermint/vm/topdown/tests/smt_voting.rs
```diff
@@ -0,0 +1,499 @@
...
+}
+
+/// Model state of voting
+#[derive(Clone)]
+pub struct VotingState {
+    /// We have a single parent chain that everybody observes, just at different heights.
+    /// There is no forking in this test because we assume that the syncing component
+    /// only downloads blocks which are final, and that reorgs don't happen.
+    ///
+    /// The tally is currently unable to handle reorgs and rejects equivocations anyway.
```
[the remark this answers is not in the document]
aakoshh: https://linear.app/interplanetary-consensus/issue/ENG-623/handle-equivocations-in-vote-gossiping

fendermint/vm/topdown/src/voting.rs
aakoshh: https://linear.app/interplanetary-consensus/issue/ENG-624/throttle-vote-gossiping

fendermint/vm/topdown/src/sync/syncer.rs
aakoshh: Thanks for highlighting this; the tally doesn't handle reorgs in any way at the moment. I wasn't sure how well it worked with this component either, so I ignored it. I thought it might be a remnant of when we fetched up to the tip of the chain where it was volatile, but as I understand currently it only fetches what is supposed to be final, and then it votes for that too, so if everyone votes it gets finalized and no reorg can happen, or at least we can't handle it.

fendermint/vm/topdown/src/voting.rs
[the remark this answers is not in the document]
aakoshh: It looks like this is the the formula in CometBFT as well: https://github.com/cometbft/cometbft/blob/a8991d63e5aad8be82b90329b55413e3a4933dc0/types/vote_set.go#L307

Issue #598, opened 2024-01-19
hardhat test simple case


snissn:
Hey can anyone reproduce this test case where a very simple hardhat test project is failing with this error:
```
mikeseiler@Mikes-MacBook-Air ipc-hardhat-test3 % npx hardhat --network local run scripts/deploy.js
Compiled 1 Solidity file successfully (evm target: paris).
ProviderError: unexpected transaction type
    at HttpProvider.request (/Users/mikeseiler/dev/snissn/ipc-hardhat-test3/node_modules/hardhat/src/internal/core/providers/http.ts:88:21)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at HardhatEthersSigner.sendTransaction (/Users/mikeseiler/dev/snissn/ipc-hardhat-test3/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
    at ContractFactory.deploy (/Users/mikeseiler/dev/snissn/ipc-hardhat-test3/node_modules/ethers/src.ts/contract/factory.ts:111:24)
    at main (/Users/mikeseiler/dev/snissn/ipc-hardhat-test3/scripts/deploy.js:15:16)
```

I've created a repro repo https://github.com/snissn/ipc-hardhat-test . please install with npm install and not yarn .  I am testing IPC by running cargo make --makefile ./infra/fendermint/Makefile.toml testnode

snissn:
when running this command with ipc i have the following http output:

```
mikeseiler@Mikes-MacBook-Air ~ % socat -v TCP-LISTEN:8555,fork,reuseaddr TCP:localhost:8545


> 2024/01/19 12:39:28.000930581  length=208 from=0 to=207
POST / HTTP/1.1\r
host: 127.0.0.1:8555\r
connection: keep-alive\r
Content-Type: application/json\r
User-Agent: hardhat 2.19.4\r
content-length: 63\r
\r
{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}< 2024/01/19 12:39:28.000964776  length=169 from=0 to=168
HTTP/1.1 200 OK\r
content-type: application/json-rpc;charset=utf-8\r
content-length: 43\r
date: Fri, 19 Jan 2024 22:39:28 GMT\r
\r
{"jsonrpc":"2.0","result":"0x12c09","id":1}> 2024/01/19 12:39:28.000975739  length=3367 from=0 to=3366
POST / HTTP/1.1\r
[...]

PR #602, opened 2024-01-20
Bugfix for hardhat incorrectly assuming that IPC is a legacy network

snissn:
Hardhat queries eth_feeHistory to construct its transaction and understand what gas format to use. Hardhat assumes that baseFeePerGas returned from this endpoint has an array length of 2 and ends up accidentally sending a malformed legacy transaction.

```

 mikeseiler@Mikes-MacBook-Air ipc-hardhat-test % npx hardhat --network local run scripts/deploy.js
 response {
   baseFeePerGas: [ '0x3e8' ],
   gasUsedRatio: [ 0 ],
   oldestBlock: '0x21287',
   reward: [ [ '0x0' ] ]
 }

 mikeseiler@Mikes-MacBook-Air ipc-hardhat-test % npx hardhat --network ganache  run scripts/deploy.js
 response {
   oldestBlock: '0x9',
   baseFeePerGas: [ '0x123d6551', '0x1002692e' ],
   gasUsedRatio: [ 0.0108704 ],
   reward: [ [ '0x4d84ee9' ] ]
 }

```
in hardhat -  [internal/core/providers/gas-provider.js](https://github.com/NomicFoundation/hardhat/blob/b626dc15b047773603964a66f7efa9982d188bda/packages/hardhat-core/src/internal/core/providers/gas-providers.ts#L290) this is where they create the error. 

Attached in this PR is code that simply doubles the array if it has a length of 1.

Related issue https://github.com/consensus-shipyard/ipc/issues/598


aakoshh:
Strange, the code you link to shows that they are asking for 1 block in the results, yet they expect 2 items. 

Since you are touching that part of the code, could you please move [this statement](https://github.com/consensus-shipyard/ipc/blob/904705346b7788e79920526c95c0e62bee5869a3/fendermint/eth/api/src/apis/eth.rs#L277) inside the `if` above? That is, if there is no result for the latest block, we should not decrement the block count. 

It wouldn't help with this issue, but I think it would be possible to get 0 results even if somebody asks for 1.

snissn:
I have been able to test this code change can confirm that it enables hardhat to properly connect with fendermint eth

aakoshh:
Look, there is some [special casing](https://github.com/filecoin-project/lotus/blob/v1.25.2/node/impl/full/eth.go#L721-L731) for this in Lotus, the array doesn't start empty, there is an initial element. They have a comment which I don't understand, and conclude that they will duplicate the newest element, but it's unconditional, not just if there is 1 block requested. 

Should we do the same?

aakoshh:
I pushed a commit which tries to fetch the next available base fee to put it into the array at the beginning.

snissn:
@aakoshh  I have tested and can confirm that your latest updates fix the incompatibility with hardhat! The tests run correctly now!

also here is the output from the above eth api call in the newest commit 0f87ed498e6b7b895e1ddac36a20564064fa48c6 

```
response {
  baseFeePerGas: [ '0x3e8', '0x3e8' ],
  gasUsedRatio: [ 0 ],
  oldestBlock: '0x534d7',
  reward: [ [ '0x0' ] ]
}
```

aakoshh (approved):
It's been tested by Mike, so let's go!

PR #603, opened 2024-01-22
Topdown receipt

cryptoAtwill:
Handing top down message receipt insertion. The idea is as follows:

When top down messages are executed, the contract exposes a `applyMsgWithRet` method that returns (bool hasReturn, ApplyRet ret) tuple. Some message does not have return receipts, so we don't handle receipt insertion. If there are return receipts, we will call `sendReceipt` method in the contract to insert the receipt. It is also possible the message execution fails at the vm, which will not be caught by the contract. In this case, it is handled directly by fendermint.

Note that for bottom up, the receipt is not handled at all, at least at the current stage. This is because the parent is running lotus, which we cannot insert the receipt at the vm level. Hence, we are still using the previous contract execution receipt handling logic.


contracts/src/structs/CrossNet.sol
```diff
@@ -104,3 +104,12 @@ struct ReceiptMsg {
     // TODO: In the future we may include here events and other
     // feedback information.
 }
+
+struct ApplyMsgRet {
+    /// The ret bytes from message execution
```
dnkolegov, line 109: What is ret?

I think naming can be improved here. For me, even the comment "The ret bytes from message execution" says not much. Probably, ret is returned?

contracts/src/gateway/router/XnetMessagingFacet.sol
```diff
@@ -29,4 +29,28 @@ contract XnetMessagingFacet is GatewayActorModifiers {
     function applyCrossMessages(IpcEnvelope[] calldata crossMsgs) external systemActorOnly {
         LibGateway.applyMessages(s.networkName.getParentSubnet(), crossMsgs);
     }
+
+    /// @notice Applies top-down cross-net message locally one by one.
```
dnkolegov, line 33: This description does not tell what `ret` is but `ret` is in the name of the method.

dnkolegov (commented):
My main concern here is naming and descriptions of the functions.

contracts/src/structs/CrossNet.sol
cryptoAtwill: sounds good, will do so. I used `ret` because it's in the source code. I will make it more expressive.

contracts/binding/src/lib.rs
```diff
@@ -2,25 +2,33 @@
 #[macro_use]
 mod convert;
 #[allow(clippy::all)]
-pub mod i_diamond;
+pub mod checkpointing_facet;
 #[allow(clippy::all)]
-pub mod diamond_loupe_facet;
+pub mod cross_msg_helper;
```
aakoshh, line 7: @cryptoAtwill something on your end still formats this code into alphabetical ordering. Is that because you manually open and save this file, or is there something else at play? Could it be something that explains why we couldn't get a deterministic build?

contracts/src/gateway/router/XnetMessagingFacet.sol
```diff
@@ -29,4 +29,28 @@ contract XnetMessagingFacet is GatewayActorModifiers {
...
+    ///         This is invoked by IPC nodes when drawing messages from their parent
+    ///         subnet for local execution. That's why the sender is restricted to the system sender,
+    ///         because this method is implicitly invoked by the node during block production.
+    /// @dev It requires the caller to be the system actor.
+    /// @param crossMsg The cross-network message to be applied.
+    function applyMsgWithRet(
+        IpcEnvelope calldata crossMsg
+    ) external systemActorOnly returns (bool hasReturn, ApplyMsgRet memory) {
```
aakoshh, line 41: Just so I understand: `hasReturn` indicates that the second member is not a default value?

contracts/binding/src/lib.rs
cryptoAtwill: what I did was `make compile-abi`, `make rust-binding` and `cargo fmt`. Is it because the contract changed that's causing this?

contracts/src/gateway/router/XnetMessagingFacet.sol
```diff
@@ -29,4 +29,28 @@ contract XnetMessagingFacet is GatewayActorModifiers {
...
+    /// @notice Sends a receipt from the execution of a cross-message.
+    /// Only `Call` messages trigger a receipt. Transfer messages should be directly
+    /// handled by the peer client to return the funds to the from address in the
+    /// failing network. This should be only called for top-down message execution.
+    function sendReceipt(
+        IpcEnvelope calldata originalCrossMsg,
+        bool success,
+        bytes memory ret
+    ) external systemActorOnly {
+        LibGateway.sendReceipt(originalCrossMsg, success, ret);
```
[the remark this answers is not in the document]
cryptoAtwill: that's the original implementation, but that wont catch the vm errors, which is not handled by the contract.

contracts/src/lib/LibGateway.sol
```diff
@@ -416,6 +416,100 @@ library LibGateway {
...
+    /// This function assumes that the relevant funds have been already minted or burnt
+    /// when the top-down or bottom-up messages have been queued for execution.
+    /// This function is not expected to revert. If a controlled failure happens, a new
+    /// cross-message receipt is propagated for execution to inform the sending contract.
+    /// `Call` cross-messages also trigger receipts if they are successful.
+    /// @dev this method differs from `applyMsg` as it returns the return data to the caller
+    ///      instead of saving it. The vm will insert the ret back into the contract.
+    ///      At the current stage, we only call this method in top-down
+    ///      message execution. In bottom-up, the execution of cross net messages happens in 
+    ///      the parent, which most likely, we have no access to its vm execution
```
aakoshh, line 429: Why can't we handle it the same way in both cases? How does it work with Lotus?

contracts/src/lib/LibGateway.sol
```diff
@@ -416,6 +416,100 @@ library LibGateway {
...
+            applyRet = ApplyMsgRet({
+                crossMsg: crossMsg,
+                success: false,
+                ret: abi.encodeWithSelector(InvalidCrossMsgDstSubnet.selector)
+            });
+            return (true, applyRet);
+        }
+
+        // The first thing we do is to find out the directionality of this message and act accordingly,
+        // incrasing the applied nonces conveniently.
```
aakoshh, line 445:
```suggestion
        // increasing the applied nonces conveniently.
```

contracts/src/lib/LibGateway.sol
cryptoAtwill: There are cases of "vm" errors which are not thrown in the contract, so receipt will not be available in the contract and cannot be committed.

contracts/src/lib/LibGateway.sol
```diff
@@ -416,6 +416,100 @@ library LibGateway {
...
+            // The value carried in top-down messages locally maps to the native coin, so we pass over the
+            // native supply source.
+            supplySource = SupplySourceHelper.native();
+        }
+
+        // If the crossnet destination is NOT the current network (network where the gateway is running),
+        // we add it to the postbox for further propagation.
+        // Even if we send for propagation, the execution of every message
+        // should increase the appliedNonce to allow the execution of the next message
+        // of the batch (this is way we have this after the nonce logic).
```
aakoshh, line 490:
```suggestion
        // of the batch (this is why we have this after the nonce logic).
```

contracts/src/lib/LibGateway.sol
```diff
@@ -416,6 +416,100 @@ library LibGateway {
...
+        if (!crossMsg.to.subnetId.equals(s.networkName)) {
+            bytes32 cid = crossMsg.toHash();
+            s.postbox[cid] = crossMsg;
```
aakoshh, line 493: How do you know what to pick up from the postbox later?

contracts/src/lib/LibGateway.sol
```diff
@@ -416,6 +416,100 @@ library LibGateway {
...
+            return (
+                false,
```
aakoshh, line 495: If this is `false`, why not return the default `applyRet`?

contracts/src/structs/CrossNet.sol
```diff
@@ -104,3 +104,12 @@ struct ReceiptMsg {
...
+    bytes ret;
+    /// Whether the execution is successful
+    bool success;
+    /// The cross net message executed
+    IpcEnvelope crossMsg;
```
aakoshh, line 114: Is this returned so as not to have to clone the message when calling this?

cryptoAtwill: reason for this is that there are inplace modification to the crossmsg, so I cannot just use the original one. Maybe I should add a comment for this.

fendermint/vm/interpreter/src/fvm/state/fevm.rs
```diff
@@ -64,6 +64,10 @@ pub struct ContractCallerReturn<T> {
 
 impl<T: Detokenize> ContractCallerReturn<T> {
     pub fn into_decoded(self) -> anyhow::Result<T> {
+        self.decoded_ret()
+    }
+
+    pub fn decoded_ret(&self) -> anyhow::Result<T> {
```
aakoshh, line 70:
```suggestion
        self.decoded()
    }

    pub fn decoded(&self) -> anyhow::Result<T> {
```

If we already call it `into_decoded` and the only difference is whether it takes `self` or `&self` then I don't see a reason to introduce a different name what this can be turned into.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -263,16 +275,50 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
+
+    fn apply_cross_msg(
+        &self,
+        state: &mut FvmExecState<DB>,
+        msg: xnet_messaging_facet::IpcEnvelope,
+        final_ret: &mut TopDownApplyRetAggregator,
+    ) -> anyhow::Result<()> {
+        let (msg, success, ret) = match self
+            .xnet
+            .call_with_return(state, |c| c.apply_msg_with_ret(msg.clone()))
```
aakoshh, line 297: If you have to clone here anyway so that you can still access `msg` in the error case, then I really don't see why `cross_msg` is returned with the result. Can it be different from the input?

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -323,3 +369,51 @@ pub fn tokens_to_burn(msgs: &[gateway_getter_facet::IpcEnvelope]) -> TokenAmount
...
+
+    is_appending_events: bool,
+    is_adding_gas: bool,
+}
+
+impl TopDownApplyRetAggregator {
+    pub fn default(ret: FvmApplyRet) -> Self {
+        Self {
+            final_return: ret,
+            is_appending_events: false,
```
aakoshh, line 386: Why is this `false` by default/

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -263,16 +275,50 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
+                final_ret.agg_return(v.into_return());
+
+                if !has_return {
+                    return Ok(());
+                }
+
+                (apply_ret.cross_msg, apply_ret.success, apply_ret.ret)
+            }
+            Err(e) => {
+                let ret = ethers::abi::encode(&[Token::String(e.to_string())]);
```
aakoshh, line 311: This `e` is the failure case of the `anyhow::Result`, right? I don't think you should use `e.to_string()` for sending back via ledger what happened to the contract. It won't be deterministic for many reasons:
* It's whatever `anyhow` decides to format the errors as
* Another non-Rust implementation, or Rust but not using this library, will have something completely different
* The concrete value returned by `call_with_return` (ie. `bail!("failed to execute contract call to {}:\ncode: {}\nerror: {:?}\ninfo: {}"`) was meant for a readable error in the logs, not as a programmatically accessible error. 

If you want to send back something, you should look at `try_call_with_ret` which returns a `CallError` if things go wrong, which contains the exit code and failure info, which should be deterministic. This is wrapped in `Ok(Err(...))` - if the outer `Ok` is an `Err` then the whole invocation failed and there's nothing to send back.

So, I reckon what you send back has to be similar to what you receive from the FVM: an exit code, a (potentially empty) returned bytes, and the separate failure info. Don't do anything to these if you want them to be deterministic.

aakoshh: This is why it would be good if the host wasn't involved in the _ack_ handling: if it's all handled within Solidity then there is little that can be non-deterministic about it.

contracts/binding/src/lib.rs
aakoshh: Aha, probably the `cargo fmt` part does it. It's not done by `make` at any point (my editor does it on save), and the generated source is deliberately [ignored](https://github.com/consensus-shipyard/ipc/blob/main/rustfmt.toml#L4) during format checks. 

Yes if you have any change in the contracts then you will get a completely fresh `lib.rs`; on my machine it's left as-is, the modules are in the order of appearance in `build.rs`. So when I build stuff, it will undo your changes, and because I don't use `cargo fmt`, we'll flip between ordered and unordered modules.

contracts/src/gateway/router/XnetMessagingFacet.sol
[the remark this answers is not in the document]
aakoshh: What's a VM error, why does it need catching? If the FVM fails, we don't want to send receipts anyway AFAIU. 

Here's how I thought the Solidity method would work:
* We have a single Solidity method to execute a whole batch of messages (or just one with a MMR proof, whatever)
* The Solidity method does the delegation or whatever trick it needs to do to actually process the message and gets the result. 
* It emits an event about which message was executed, to make it easy for clients to monitor the execution and outcome of cross messages.
* It can also insert an ack message if that's how you want to send results back. 

This way you don't have to aggregate multiple `ApplyRet` into a single one to sum up the gas costs, nor do you have to accumulate events: there is 1 call to the FVM all succeeding or failing together on the VM level (eg. if the relayer itself cannot cover the gas cost), with as many events about the outcomes as there are messages in the batch. 

You may or may not want to count the gas cost of the application of finality.

contracts/src/lib/LibGateway.sol
aakoshh: I have some recollection that we discussed once that it was tricky to reason about how to handle VM errors when a topdown "checkpoint" is handled as 3 independent transactions:
* the commitment of the parent finality
* the storing of the validator changes
* the execution of top-down messages

That is, it's not at all clear what it should mean if the first two succeed but the third fails.

As I understand this PR wants to skip failing messages one by one, even if they fail for unknown reasons. Another approach which IIRC I recommended was to have a single method to do everything, and fail or succeed atomically; if any of the three aspects fail, the entire proposed top-down execution fails, and we can try again later once we fixed the bug or something. 

It ain't pretty, but better than a partial execution which cannot be retried.

contracts/src/lib/LibGateway.sol
```diff
@@ -416,6 +416,100 @@ library LibGateway {
...
+        } else if (applyType == IPCMsgType.TopDown) {
+            // Note: there is no need to load the subnet, as a top-down application means that _we_ are the subnet.
+            if (s.appliedTopDownNonce != crossMsg.nonce) {
+                return (true, ApplyMsgRet({
+                    crossMsg: crossMsg,
+                    success: false,
+                    ret: abi.encodeWithSelector(InvalidCrossMsgNonce.selector)
+                }));
+            }
+            s.appliedTopDownNonce += 1;
```
aakoshh, line 479: It's still not clear to me what these "vm" errors that we're supposed to handle by doing things in the host should be, and how they should work for example with this nonce. 

For example I see that we adjust the nonce, then execute, returning whether the Solidity sub-transaction was successful or not. This is good, the next message can be handled and it should match the next nonce, all is well. However, what happens if we encounter this "vm" error? I assume it means the nonce was _not_ executed in those cases, so even if the host sends the next message, it will fail the nonce check, no?

raulk (changes requested):
Pardon me if I get some of this wrong, but:

1. We shouldn't be creating a new Solidity method. What is it needed for? Unless I'm mistaken, nothing continues using `applyMsg`.
2. The call should be handled entirely _inside_ Solidity -- IMO there is no reason to pull this out to the client. Inside the former `applyMsg`, you would need to use a [low-level Solidity call](https://medium.com/@anunobi31/understanding-low-level-calls-in-solidity-call-50100c725592) to the destination, packing the calldata to ensure you're calling the 
**IPC entrypoint method**. The low-level call already returns whether the call was successful or not. If the call is unsuccessful, the return data will contain the error compliant with the Solidity call convention. You can aggregate receipts in Solidity and commit them for dispatch.

I'm reading in review threads that the justification to do all this work in the client is to deal with some mysterious "vm error". Could you explain what you're referring to?

raulk:
I was curious if there are any edge cases or obscure corners where the low-level Solidity call itself would revert, so I asked ChatGPT. Pasting the answer here for everyone's benefit as it does an oustanding job in explaining the more general mechanics of this mechanism:

> # Low-Level Solidity Calls Revert
> 
> #### You:
> Can a low-level Solidity call revert?
> 
> #### ChatGPT:
> Yes, a low-level Solidity call can revert. In Solidity, a low-level call refers to using the `.call()`, `.delegatecall()`, or `.staticcall()` methods. These methods interact with other contracts or addresses in a more manual and less abstracted way compared to high-level contract interactions.
> 
> Here are some key points about low-level calls and reverts:
> 
> 1. **Control Over Reverting**: When you make a low-level call (like `someAddress.call(abi.encodeWithSignature("someFunction", arg1, arg2))`), you have more control over whether to revert the transaction or not. A low-level call returns two values: a boolean indicating success or failure, and the returned data from the call. You can decide to revert the transaction based on the boolean value.
> 
> 2. **No Automatic Reverts**: Unlike high-level calls (which automatically revert the entire transaction when the called function fails), low-level calls do not automatically revert the transaction. If the call fails, it's up to you to handle the failure in your contract code, for example, by using `require(success, "Call failed")`.
> 
> 3. **Error Handling**: Low-level calls are often used for more advanced error handling in Solidity. Since they allow you to directly access the success or failure of the call, you can implement custom logic depending on the outcome of the interaction.
> 
> 4. **Use Cases**: These calls are useful in scenarios where you might want to interact with a contract that might not exist, or if you want to handle failures in a specific way without reverting the entire transaction.
> 
> 5. **Gas Forwarding**: Low-level calls allow you to control how much gas is forwarded to the called contract. This can be important in some scenarios, especially when dealing with unknown contracts or executing complex operations.
> 
> 6. **Security Considerations**: Using low-level calls requires careful consideration and understanding of the implications, as they can lead to security vulnerabilities if not used properly, such as reentrancy attacks or unexpected failures.
> 
> Overall, low-level calls in Solidity offer more control and flexibility at the cost of requiring more careful management and understanding of potential risks and behaviors.
> 
> 
> #### You:
> You open your answer stating a firm "yes". You then provide scenarios where a low-level Solidity call can allow me to _handle_ reverts from the callee. Unfortunately, you do not outline any scenarios in which the low-level Solidity call **itself** would revert. Please correct your answer accordingly.
> 
> #### ChatGPT:
> You are correct in pointing out the need for clarification. In the context of whether the low-level Solidity call itself can revert, the answer is nuanced:
> 
> 1. **Direct Cause of Reversion**: A low-level call itself (using `.call`, `.delegatecall`, or `.staticcall`) does not inherently cause a revert. These methods are designed to return a boolean value indicating success or failure, along with the return data. They are specifically used because they don't automatically throw an exception (revert the transaction) if the called function fails or doesn't exist.
> 
> 2. **Handling Failure**: If the low-level call fails (e.g., the called function doesn't exist, or there's not enough gas), the call returns `false` as the success indicator. It is then up to the caller to decide how to handle this. The caller can choose to revert the transaction based on this failure by using a statement like `require(success, "Call failed")`. In this scenario, it is not the low-level call that reverts, but rather the caller's decision to revert based on the failure.
> 
> 3. **Circumstances Leading to Revert**: While the low-level call itself does not revert, certain circumstances in the broader context of the call can lead to a revert. For example, if the low-level call results in an out-of-gas situation, or if there's an invalid operation in the called code, the Ethereum Virtual Machine (EVM) will revert the transaction. However, these are effects of the execution environment and the called code, not the low-level call mechanism itself.
> 
> 4. **Gas Limitation**: If you send a low-level call with an explicit gas limit (like `someAddress.call.gas(gasAmount)(...)`) and the execution requires more gas than the amount provided, the call will fail, but it won't revert unless you handle the failure and explicitly revert.
> 
> In summary, a low-level Solidity call, in and of itself, is designed not to revert but to return a success flag that must be checked by the caller. The caller can then decide to revert based on this result. Reverts associated with low-level calls are generally due to how they are handled in the calling contract or due to execution issues like out-of-gas within the EVM during the call's execution.
>

raulk:
(Sorry, I closed accidentally!)

contracts/src/lib/LibGateway.sol
cryptoAtwill: good call, I copied the previous implementation, probably this is a separate issue.

cryptoAtwill:
@raulk The "vm errors" are errors from the environment or the vm, like out of gas. There is also a slack thread that discussed this as well, I'll share it here for context: https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1705656349698669

raulk:
Top-down messages are executed as an implicit messages with unlimited gas. There is currently nothing that's setting a per-message gas limit. (Obviously this is not the final solution). I don't see why one can get an out of gas under this circumstance?

raulk:
Re: preflight checks, these are related to the sender, balance, inclusion gas, balance, etc. None of these are relevant in our case, so I think the original assumption was misguided.

PR #605, opened 2024-01-22
eth block count bugfix

snissn:
if there is no result for the latest block, we should not decrement the block count. co-authored by @aakoshh

see https://github.com/consensus-shipyard/ipc/pull/602#issuecomment-1903818475


aakoshh (approved):
Thanks!

PR #606, opened 2024-01-23
Address review comments for #599 + refactors

raulk:
## In the SDK contract

- Removes timeout features.
- Dispatches to internal virtual methods.
- Adds Ownable.
- Various refactors.

## In IPC contracts

- Return the committed envelope when sending a xnet message so we can compute the message ID (hash).
- In `sendContractXnetMessage`, set the sender implicitly instead of trusting whatever is passed.
- Turn `CallMsg#method` into a general `bytes` type so it can support calls to built-in actors in the future (uint64 method number).

## TODO

> To be handled in follow-up PRs.

- [x] Remove the usage of `CallMsg` as a message payload for `IpcMsgKind#Transfer`; it's entirely redundant.
- [ ] In `sendContractXnetMessage`, stop using `IpcEnvelope` as the parameter and instead take `to` and `callMsg` only, using the passed-in value.
- [ ] Consolidate the many `InvalidCrossMsg*` errors into a single `InvalidXnetMessage(string reason)`.
- [ ] Move the `IpcHandler` interface to a `shared` package, so that the IPC contracts are not accessing the SDK.
- [x] Adjust tests that are sending `IpcMsgKind#Transfer` message kinds through `sendContractXnetMessage`.
- [ ] Rename `IpcMsgKind#Transfer` to `IpcMsgKind#SystemTransfer`, to make it clear that these are system generated messages.
- [ ] Rename `IpcMsgKind#Call` to `IpcMsgKind#ActorCall` to represent that this is a call originating in an actor.


contracts/src/lib/CrossMsgHelper.sol
```diff
@@ -172,42 +171,13 @@ library CrossMsgHelper {
...
-    ) public pure returns (IpcEnvelope memory ret) {
-        if (crossMsg.kind == IpcMsgKind.Call || crossMsg.kind == IpcMsgKind.Transfer) {
-            crossMsg.message = abi.encode(message);
-            return crossMsg;
-        }
-
-        // Cannot set IPCMsg for the wrong kind
-        revert InvalidCrossMsgKind();
-    }
-
```
raulk, line 210: These methods are only used in tests, so I moved them to the tests. The lesser unused code we have in production contracts, the lesser the attack surface.

contracts/src/structs/CrossNet.sol
```diff
@@ -70,37 +70,37 @@ struct IpcEnvelope {
...
     // uint256 gasLimit;
 }
 
 /// @notice Message format used for `Transfer` and `Call` messages.
-struct IpcMsg {
-    /// @dev abi.encoded method being called by the contract.
-    bytes4 method;
+struct CallMsg {
+    /// @dev Target method. A bytes4 function selector for EVM/Solidity targets, or a uint64 for Wasm actors.
+    bytes method;
```
raulk, line 91: Note this.

contracts/test/integration/GatewayDiamond.t.sol
```diff
@@ -1161,9 +1140,9 @@ contract GatewayActorDiamondTest is Test, IntegrationTestBase {
 
         IpcEnvelope[] memory msgs = new IpcEnvelope[](10);
         for (uint64 i = 0; i < 10; i++) {
-            msgs[i] = TestUtils.newTransferCrossMsg(
+            msgs[i] = TestUtils.newXnetCallMsg(
                 IPCAddress({subnetId: subnetId, rawAddress: FvmAddressHelper.from(caller)}),
-                IPCAddress({subnetId: gwGetter.getNetworkName(), rawAddress: FvmAddressHelper.from(caller)}),
+                IPCAddress({subnetId: gwGetter.getNetworkName(), rawAddress: FvmAddressHelper.from(vm.addr(100 + i))}),
```
raulk, line 1145: This change was needed because we were previously transferring to the subnet actor. We are now _calling_ instead of _transferring_, which calls the target address on the `handleIpcCall` method. Since the subnet actor didn't have that method, it failed. We now call an Eth account, and everyone is happy.

contracts/test/integration/L2GatewayDiamond.t.sol
```diff
@@ -41,9 +41,9 @@ contract L2GatewayActorDiamondTest is Test, L2GatewayActorDiamond {
 
         IpcEnvelope[] memory topDownMsgs = new IpcEnvelope[](n);
         for (uint64 i = 0; i < n; i++) {
-            topDownMsgs[i] = TestUtils.newTransferCrossMsg(
+            topDownMsgs[i] = TestUtils.newXnetCallMsg(
                 IPCAddress({subnetId: id, rawAddress: FvmAddressHelper.from(address(this))}),
-                IPCAddress({subnetId: id, rawAddress: FvmAddressHelper.from(receipient)}),
+                IPCAddress({subnetId: gwGetter.getNetworkName().getParentSubnet(), rawAddress: FvmAddressHelper.from(receipient)}),
```
raulk, line 46: Xnet calls to the same subnet are invalid.

contracts/src/gateway/GatewayMessengerFacet.sol
```diff
@@ -20,39 +21,59 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
...
+        IpcEnvelope calldata envelope
+    ) external payable returns (IpcEnvelope memory committed) {
         if (!s.generalPurposeCrossMsg) {
             revert MethodNotAllowed(ERR_GENERAL_CROSS_MSG_DISABLED);
         }
 
-        // we prevent the sender from being an EoA.
+        // TODO: consolidate all these error types into an error InvalidXnetMsg(string reason).
+        //
+        // We prevent the sender from being an EoA.
```
dnkolegov, line 42: To my knowledge, it is not possible to reliably prevent this.

contracts/sdk/IpcContract.sol
```diff
@@ -1,151 +1,112 @@
...
-            IpcEnvelope memory originalMsg = inFlightMsgs[ackMsg.id];
+        if (envelope.kind == IpcMsgKind.Call) {
+            CallMsg memory call = abi.decode(envelope.message, (CallMsg));
+            return _handleIpcCall(envelope, call);
+        } else if (envelope.kind == IpcMsgKind.Receipt) {
+            ResultMsg memory result = abi.decode(envelope.message, (ResultMsg));
+
+            // Recover the original message.
+            // If we were not tracking it, or if some details don't match, refuse to handle the receipt.
+            IpcEnvelope storage orig = inflightMsgs[result.id];
```
dnkolegov, line 48: There is no ay input validation for ID. ID is controlled by the sender.
Is it sufficient to check `keccak256(abi.encode(envelope.from)) != keccak256(abi.encode(orig.to))` ?

dnkolegov (approved):

contracts/sdk/IpcContract.sol
raulk: What kind of input validation would make sense here? This method can only be invoked by the owner.

contracts/src/gateway/GatewayMessengerFacet.sol
raulk: In which cases the check below would pass if the sender was a contract? (Note that the sender must exist, so it's either an Eth Account, EVM smart contract, or another built-in actor).

dnkolegov: That is a very interesting and arguable question in the community. The classic example is code in the constructor.

contracts/sdk/IpcContract.sol
dnkolegov: I thought about removing arbitrary messages

contracts/src/gateway/GatewayMessengerFacet.sol
raulk: I suspect you're referring to the case where a contract A _under construction_ calls another contract B in its constructor, which then enquires if A is an EOA or a contract based on A's code length. In that case, I think the EVM will report code length = 0 because the EVM bytecode is still being generated. This will give B the impression that A is an EOA. However, I don't see that edge case affecting us here in deleterious ways. We can document the exception that a contract under construction will be prevented from sending xnet messages?

contracts/sdk/IpcContract.sol
raulk: Could you elaborate further? I don't follow you.

raulk:
I've removed the `CallMsg` payload for `IpcMsgKind.Transfer` messages. 
- It was invariably being set to `CallMsg { method: METHOD_SEND (0), params: EMPTY_BYTES }`.
- It carried no additional information than what the envelope carries already (it was entirely redundant). Both the method number and the params were ignored everywhere. In fact, the method number was _actually incorrect_, since performing a Solidity transfer does not call method 0 on the chain, but actually InvokeEVM.
- We now have a clean 1:{0,1} mapping between message different message kinds and inner message types.
	- Transfer => none
	- Call => CallMsg
	- Result => ResultMsg

PR #609, opened 2024-01-23
adding tsconfig whos output is generated by `npx tsc --init`

snissn:
This fixes a problem that I was having where `npx hardhat` was incredibly slow to start up and seemed to be exploring my entire hard drive.

I ran the command

`npx tsc --init`

which creates the tsconfig.json file that is added to this commit.

Having the tsconfig file in this repo reduces my hardhat start up time from minutes to less than 1 second. I measured this through

`time npx hardhat`


snissn:
This speed up works just as well if tsconfig.json only have the value `{}` in it..

raulk (approved):

PR #514, opened 2024-01-05
add query for built-in actors; fix eth_getCode and eth_getStorageAt.

raulk:
Aims to fix #506 in a more stable way than #511. Needs tests and cleanup.

PR #610, opened 2024-01-24
Properly fix eth_getCode and eth_getStorageAt

mb1896:
Based on Raul's draft [PR](https://github.com/consensus-shipyard/ipc/pull/514/files) 

Manually verified with

getCode:
* Non existent address: return `0x0`
* EVM account: return `0x0`
* EVM contract: return expected binary code

getStorageAt
* Non existent address: return `0x0`
* EVM account: return `0000000000000000000000000000000000000000000000000000000000000000`
* EVM contract: return expected storage

Integration test covered by existing assertions in `fendermint/eth/api/examples/ethers.rs` running as `smoke-test`.

### DIFF
--- a/fendermint/eth/api/src/apis/eth.rs
+++ b/fendermint/eth/api/src/apis/eth.rs
@@ -18,14 +18,14 @@ use fendermint_rpc::response::{decode_fevm_invoke, decode_fevm_return_data};
 use fendermint_vm_actor_interface::eam::{EthAddress, EAM_ACTOR_ADDR};
 use fendermint_vm_actor_interface::evm;
 use fendermint_vm_message::chain::ChainMessage;
-use fendermint_vm_message::query::{ActorState, FvmQuery, FvmQueryHeight};
+use fendermint_vm_message::query::FvmQueryHeight;
 use fendermint_vm_message::signed::SignedMessage;
 use futures::FutureExt;
 use fvm_ipld_encoding::RawBytes;
 use fvm_shared::address::Address;
 use fvm_shared::bigint::BigInt;
 use fvm_shared::crypto::signature::Signature;
-use fvm_shared::{chainid::ChainID, error::ExitCode, ActorID};
+use fvm_shared::{chainid::ChainID, error::ExitCode};
 use jsonrpc_v2::Params;
 use rand::Rng;
 use tendermint::block::Height;
@@ -35,6 +35,8 @@ use tendermint_rpc::{
     endpoint::{block, block_results, broadcast::tx_sync, consensus_params, header},
     Client,
 };
+
+use fil_actors_evm_shared::uints;
 
 use crate::conv::from_eth::to_fvm_message;
 use crate::conv::from_tm::{self, msg_hash, to_chain_message, to_cumulative, to_eth_block_zero};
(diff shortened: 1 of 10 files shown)


fendermint/eth/api/src/apis/eth.rs
```diff
@@ -794,8 +794,6 @@ where
 
     // Return empty if not an EVM actor.
     if data.get_actor_type(&address, height).await? != ActorType::Known("evm".to_string()) {
-        tracing::info!("jiejie: SHIT SHIT");
```
raulk: 😆

mb1896: Sorry..all cleaned up.

mb1896:
@aakoshh Can you please review this PR? Thanks!

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -738,6 +740,23 @@ pub async fn get_storage_at<C>(
...
+            data.to_big_endian(&mut bz);
+        }
+        // The client library expects hex encoded string.
+        Ok(hex::encode(bz))
+    };
+
+    let height = data.query_height(block_id).await?;
+
+    // If not an EVM actor, return empty.
+    if data.get_actor_type(&address, height).await? != ActorType::Known("evm".to_string()) {
```
aakoshh, line 755: I think it would be worth creating `const` instances of known types [such as these](https://github.com/filecoin-project/ref-fvm/blob/daefd6bcb4eefe3e8c29454d73a295c3fbe81bfc/shared/src/error/mod.rs#L50-L57), and then we should be able to reuse it as `ActorType::Evm` and not have to repeat this, always guessing if it's upper or lower case, etc.

fendermint/eth/api/src/state.rs
```diff
@@ -395,6 +407,30 @@ where
...
+            ActorState {
+                code: actor_type_cid,
+                ..
+            },
+        )) = self.client.actor_state(&addr, height).await?.value
+        else {
+            return Ok(ActorType::Inexistent);
+        };
+        let registry = self.client.builtin_actors(height).await?.value.registry;
+        let ret = match registry.into_iter().find(|(_, cid)| cid == &actor_type_cid) {
```
aakoshh, line 428: Have you considered what I commented about caching these lookups?

aakoshh (approved):
This looks correct, thank you for following up on some of my comments on the draft.

My only remaining suggestions are to:
1. Remove the need for the repetition of the `"evm"` magic string.
2. Cache the results of API calls. CID-to-type is an obvious one - if found, it's going to be the same most likely forever. But also addr-to-type could be considered (with a size limit) in case somebody uses `eth_getStorageAt` to traverse the state of an actor, which I'm not sure is something they do.

Could be follow-up PRs if you agree with any of these.

raulk:
I agree with caching -- it seems wasteful to round trip to get environmental info that won't change in 99% of the cases. However, this opens up the can of worms of cache invalidation. Because Fendermint doesn't support upgrade schedules, it may not be an actual problem, but it all depends on how we plan to carry out upgrades going forward.

fendermint/eth/api/src/apis/eth.rs
mb1896: Fixed as you suggested. Thanks!

fendermint/eth/api/src/state.rs
mb1896: Added a [TODO in Linear](https://linear.app/interplanetary-consensus/issue/ENG-616/caching-the-result-of-get-actor-type) for that.

PR #611, opened 2024-01-24
Fix create bottomup checkpoint

cryptoAtwill:
Fix the CI error in `create_bottom_up_checkpoint`: perform the gateway enabled check before doing contract queries.

Also fixing a minor issue in  submit checkpoint in relayer.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -63,9 +63,7 @@ where
         .block_hash()
         .ok_or_else(|| anyhow!("block hash not set"))?;
 
-    let (msgs, subnet_id) = if let Some(v) = should_create_checkpoint(gateway, state, height)? {
-        v
-    } else {
+    let Some((msgs, subnet_id)) = should_create_checkpoint(gateway, state, height)? else {
         return Ok(None);
     };
 
@@ -325,10 +323,6 @@ where
             "bottom up checkpoint period reached height"
         );
     } else {
-        tracing::debug!(
-            height = height.value(),
-            "height has no message batch nor reached checkpoint period "
-        );
         return Ok(None);
     }
 
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -63,30 +63,11 @@ where
         .block_hash()
         .ok_or_else(|| anyhow!("block hash not set"))?;
 
-    let batch = gateway.bottom_up_msg_batch(state, height.into())?;
-
-    if batch.block_height.as_u64() == 0 {
-        tracing::debug!(height = height.value(), "no bottom up msg batch at height");
-
-        // we need to check if the checkpoint period has reached
-        if should_create_checkpoint(gateway, state, height)?.is_none() {
-            tracing::debug!(
-                height = height.value(),
-                "not bottom up checkpoint period height"
-            );
-            return Ok(None);
-        }
-
-        tracing::debug!(
-            height = height.value(),
-            "bottom up checkpoint period height reached"
-        );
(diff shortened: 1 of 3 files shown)


fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -322,23 +297,47 @@ fn should_create_checkpoint<DB>(
...
+    } else if height.value() % gateway.bottom_up_check_period(state)? == 0 {
+        tracing::debug!(
+            height = height.value(),
+            "bottom up checkpoint period reached height"
+        );
+    } else {
+        tracing::debug!(
+            height = height.value(),
+            "height has no message batch nor reached checkpoint period "
+        );
```
aakoshh, line 331: This will be a bit spammy on debug level at every block height.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -322,23 +297,47 @@ fn should_create_checkpoint<DB>(
...
+    let id = gateway.subnet_id(state)?;
+    let is_root = id.route.is_empty();
+
+    if is_root {
+        return Ok(None);
+    }
+
+    let batch = gateway.bottom_up_msg_batch(state, height.into())?;
+
+    if batch.block_height.as_u64() != 0 {
```
aakoshh, line 317: Have you considered some minimum batch size here?

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -63,30 +63,11 @@ where
...
+    let (msgs, subnet_id) = if let Some(v) = should_create_checkpoint(gateway, state, height)? {
+        v
     } else {
-        tracing::debug!(
-            height = height.value(),
-            "bottom up msg batch found at height"
-        );
-    }
+        return Ok(None);
+    };
```
aakoshh, line 70: You can use the new [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) syntax here:

```rust
let Some((msgs, subnet_id)) = should_create_checkpoint(...) else {
  return Ok(None);
}
```

ipc/provider/src/checkpoint.rs
```diff
@@ -199,18 +199,22 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
                     .await?;
                 log::debug!("bottom up bundle: {bundle:?}");
 
-                todo!("implement submit checkpoint")
```
aakoshh, line 202: This change look unrelated from the fix. Just to be clear this is called by the relayer, right?

aakoshh (approved):
Looks correct, just minor comments.

ipc/provider/src/checkpoint.rs
cryptoAtwill: yeah, it's a minor thing, should have enabled, but too small for another PR.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
cryptoAtwill: yeah, probably the batch full log message can stay, this one can go

fendermint/vm/interpreter/src/fvm/checkpoint.rs
cryptoAtwill: From the contract, there is no minimum batch size, it's either non-exists or some number of messages. If it's zero, that means the batch does not exists.

aakoshh: I know, I meant: should it create a new checkpoint even if there is just 1 message in the batch, or should it wait until there is some, but not too many; e.g. make a checkpoint whenever there are at least 10 outgoing messages _or_ 1000 blocks have been created.

PR #629, opened 2024-01-25
Adding receipt for Transfer

cryptoAtwill:
Triggering receipt for transfer but only when the Transfer execution fails.

Fixes ENG-593.


contracts/src/lib/LibGateway.sol
```diff
@@ -447,10 +447,14 @@ library LibGateway {
...
-        // - Transfers don't need to be acknowledged.
+        // if we get a `Receipt` do nothing, no need to send receipts.
         // - And sending a `Receipt` to a `Receipt` could lead to amplification loops.
-        if (crossMsg.kind == IpcMsgKind.Transfer || crossMsg.kind == IpcMsgKind.Result) {
+        if (crossMsg.kind == IpcMsgKind.Result) {
+            return;
+        }
+
+        // if we get a `Transfer` and the execution is ok, we dont send the receipt.
+        if (crossMsg.kind == IpcMsgKind.Transfer && outcomeType == OutcomeType.Ok) {
```
raulk, line 457: @cryptoAtwill what case were you trying to catch here? We can only get outcomes in Result messages, not in Transfer messages.

cryptoAtwill: This is creating the ResultMsg for Transfer. My thinking was if the message is a Transfer and the execution is successful, then we dont really need to create the receipt.

raulk: Ah, sorry, I misread! I thought we were looking into the outcome of the crossMsg itself :-/ Yes, let's remove this because we do want to relay an OK as well when a deposit/withdrawal (transfer) succeeds. Otherwise a contract will never learn that it succeeded and it can stop tracking an inflight operation.

raulk (approved):

PR #630, opened 2024-01-25
[HOLD] Add an example custom kernel defining a custom syscall

fridrik01:
DO NOT MERGE

This is PR is part of a pluggable syscall tutorial which we will share later and contains the final source code after following the tutorial.

**Context**
This PR replaces the current Fendermint FVM kernel with a new custom kernel which implements a new syscall and shows how you can use this syscall from a new custom actor which we load at genesis.

More specifically, this PR:
- Implement a new `CustomKernelImpl FVM kernel which extends the default kernel.
- Implement a new syscall called `my_custom_syscall` on the new kernel
- Switches out the DefaultKernel used in Fendermint with this new custom kernel
- Creates a new `customsyscall` actor which demonstrates calling the new syscall
- Updates fendermint to load this new actor at genesis and run it at each epoch

**Test plan**
You can test this new kernel/syscall/actor by running one of our integration tests.

First you need to have docker installed, then compile a local docker image with this PR checked out:

```
cd fendermint
make docker-build
```
After the fendermint docker image has been built, you can run one of the integration tests

```
cd fendermint/testing/smoke-test
# creates the docker containers
cargo make setup
# runs the integration test
cargo make test
```

View fendermint logs and see the output generated by calling the `customsyscall` actor in each epoch:

```
docker ps
CONTAINER ID   IMAGE                       COMMAND
8da423d8bb1e   fendermint:latest           "fendermint --networ…"
...
```

View the docker logs:

```
docker logs 8da423d8bb1e
...
customsyscall actor returned: 21
```

You can now run `cargo make teardown` to stop the container.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
+++ b/fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
@@ -1,0 +1,201 @@
+// Copyright 2021-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+use fvm::call_manager::CallManager;
+use fvm::gas::Gas;
+use fvm::kernel::filecoin::{DefaultFilecoinKernel, FilecoinKernel};
+use fvm::kernel::prelude::*;
+use fvm::kernel::Result;
+use fvm::kernel::{
+    ActorOps, CryptoOps, DebugOps, EventOps, IpldBlockOps, MessageOps, NetworkOps, RandomnessOps,
+    SelfOps, SendOps, SyscallHandler, UpgradeOps,
+};
+use fvm::syscalls::Linker;
+use fvm_shared::clock::ChainEpoch;
+use fvm_shared::consensus::ConsensusFault;
+use fvm_shared::piece::PieceInfo;
+use fvm_shared::randomness::RANDOMNESS_LENGTH;
+use fvm_shared::sector::{
+    AggregateSealVerifyProofAndInfos, RegisteredSealProof, ReplicaUpdateInfo, SealVerifyInfo,
+};
+use fvm_shared::sys::out::network::NetworkContext;
+use fvm_shared::sys::out::vm::MessageContext;
+use fvm_shared::{address::Address, econ::TokenAmount, ActorID, MethodNum};
+
+use ambassador::Delegate;
+use cid::Cid;
+
+// we define a single custom syscall which simply doubles the input
+pub trait CustomKernel: Kernel {
+    fn my_custom_syscall(&self, doubleme: i32) -> Result<i32>;
+}
+
+// our custom kernel extends the filecoin kernel
+#[derive(Delegate)]
+#[delegate(IpldBlockOps, where = "C: CallManager")]
+#[delegate(ActorOps, where = "C: CallManager")]
+#[delegate(CryptoOps, where = "C: CallManager")]
+#[delegate(DebugOps, where = "C: CallManager")]
+#[delegate(EventOps, where = "C: CallManager")]
+#[delegate(MessageOps, where = "C: CallManager")]
+#[delegate(NetworkOps, where = "C: CallManager")]
+#[delegate(RandomnessOps, where = "C: CallManager")]
+#[delegate(SelfOps, where = "C: CallManager")]
+#[delegate(SendOps<K>, generics = "K", where = "K: CustomKernel")]
+#[delegate(UpgradeOps<K>, generics = "K", where = "K: CustomKernel")]
+pub struct DefaultCustomKernel<C>(pub DefaultFilecoinKernel<C>);
+
+impl<C> CustomKernel for DefaultCustomKernel<C>
+where
+    C: CallManager,
+    DefaultCustomKernel<C>: Kernel,
+{
+    fn my_custom_syscall(&self, doubleme: i32) -> Result<i32> {
+        Ok(doubleme * 2)
+    }
+}
+
+impl<C> DefaultCustomKernel<C>
+where
+    C: CallManager,
+{
+    fn price_list(&self) -> &PriceList {
+        (self.0).0.call_manager.price_list()
+    }
+}
+
+impl<C> Kernel for DefaultCustomKernel<C>
+where
+    C: CallManager,
+{
+    type CallManager = C;
+    type Limiter = <DefaultFilecoinKernel<C> as Kernel>::Limiter;
+
+    fn into_inner(self) -> (Self::CallManager, BlockRegistry)
+    where
+        Self: Sized,
+    {
+        self.0.into_inner()
+    }
+
+    fn new(
+        mgr: C,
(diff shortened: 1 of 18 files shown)


fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
raulk, line 1: @fridrik01 could you cook up a quick test that demonstrates how one would plug this custom kernel through the Fendermint configuration/constructors?

fridrik01: @raulk we don't have the code yet to set kernels based on configuration, but to do this manually we only need to changes this one line I recorded in this PR summary to switch kernels. Is that what you meant, or are you thinking about writing actual tests which test the custom syscall? There are already integration tests in the ref-fvm repo which do that though

fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
```diff
@@ -0,0 +1,201 @@
...
+#[delegate(CryptoOps, where = "C: CallManager")]
+#[delegate(DebugOps, where = "C: CallManager")]
+#[delegate(EventOps, where = "C: CallManager")]
+#[delegate(MessageOps, where = "C: CallManager")]
+#[delegate(NetworkOps, where = "C: CallManager")]
+#[delegate(RandomnessOps, where = "C: CallManager")]
+#[delegate(SelfOps, where = "C: CallManager")]
+#[delegate(SendOps<K>, generics = "K", where = "K: CustomKernel")]
+#[delegate(UpgradeOps<K>, generics = "K", where = "K: CustomKernel")]
+pub struct DefaultCustomKernel<C>(pub DefaultFilecoinKernel<C>);
```
mikevoronov, line 45: The `DefaultCustomKernel` name makes the difference between `CustomKernel` not so obvious, maybe it'd be better to name it `CustomKernelImpl`?

fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
```diff
@@ -0,0 +1,201 @@
...
+};
+use fvm_shared::sys::out::network::NetworkContext;
+use fvm_shared::sys::out::vm::MessageContext;
+use fvm_shared::{address::Address, econ::TokenAmount, ActorID, MethodNum};
+
+use ambassador::Delegate;
+use cid::Cid;
+
+// we define a single custom syscall which simply doubles the input
+pub trait CustomKernel: Kernel {
```
mikevoronov, line 28: Am I right that if one wants to register a syscall only for an actor, then it's not necessary to implement a new kernel trait and it'd be enough to implement just a new kernel impl (which is called here `DefaultCustomKernel`) and corresponding `SyscallHandler` for it.

fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
fridrik01: yeah, that is probably a better naming for what it is, I kept with the naming convention as `ref-fvm` with `Kernel` and`DefaultKernel` but open to change it

fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
fridrik01: You don't register a syscall for an individual actor, but for a whole kernel which then exposes that syscall to all actors.

If you want to register a _new_ syscall, then you have to define a new kernel trait with a new kernel implementation and corresponding SyscallHandler. 

However, If you want to override an existing syscall then you don't _need_ a new trait but just a separate implementation/corresponding SyscallHandler.

mikevoronov: > If you want to register a new syscall, then you have to define a new kernel trait with a new kernel implementation and corresponding SyscallHandler.

I tried to understand why, following the source code of Fendermint and came up with only idea that it's really needed if one want to change some types like `Limiter`.

Could you please explain which drawbacks are in [this](https://github.com/consensus-shipyard/ipc/compare/main...fluencelabs:ipc:fluence#diff-2f5413c018d995ea1371b7c3a906f41bfe06fdd65077103969efd1a8d5857a1cR33) approach where I define a new kernel impl and then impl `SyscallHandler` with a new syscall for it?

fridrik01: Hmm, maybe I am misunderstanding but you need to somehow invoke the new syscall on the new kernel, I don't see the implementation of the `fluence_fvm_syscall::run_randomx`, where is the actual kernel implementation of the syscall? 

See similar to the [link_syscall function](https://github.com/filecoin-project/ref-fvm/blob/40d396edcd486ae5185f649ab59d52e8815975a3/testing/integration/src/custom_kernel/mod.rs#L200) which calls the new syscall on the kernel [here](https://github.com/filecoin-project/ref-fvm/blob/40d396edcd486ae5185f649ab59d52e8815975a3/testing/integration/src/custom_kernel/mod.rs#L53)

mikevoronov: Right, it's being invoked in directly this [function](https://github.com/fluencelabs/fluence-fendermint/blob/main/fluence-fendermint-syscall/src/lib.rs#L27). I don't see why do we need to call it from a kernel...

fridrik01: Yeah, if you can implement the functionality outside of the kernel, than that should be fine!

fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn gas_available(&self) -> Gas {
+        self.0.gas_available()
+    }
+
+    fn charge_gas(&self, name: &str, compute: Gas) -> Result<GasTimer> {
+        self.0.charge_gas(name, compute)
+    }
+}
+
+impl<C> FilecoinKernel for DefaultCustomKernel<C>
```
longfeiWan9, line 117: @fridrik01 since are customizing syscall for IPC, is it necessary to implement FilcoinKernel for DefaultCustomKernel?

fridrik01: nope, will update!

fendermint/vm/interpreter/src/fvm/examples/mycustomkernel.rs
maciejwitowski: I remember we discussed users at this stage won't be able to plug a custom kernel via Fendermint configuration/constructor by rather will need to fork the Fendermint, right?

fridrik01:
This PR can be closed as its for documentation purposes

Issue #628, opened 2024-01-25
Multi-subnet Solidity tests


raulk:
## Problem

Currently our Solidity tests are only testing behaviours within a single subnet. When we expect an interaction (xnet call) to impact another subnet, our tests can only assert on the behaviour of the local Gateway actor, i.e the outputs from the subnet: checkpoints, mesages, etc. However, this is only one side of the picture, and we could get a ton more coverage if we were able to build multi-subnet tests.

## How this would work

Create a new integration test base that instantiates two subnets (parent and child). The parent would instantiate the gateway actor and the subnet actor representing the subnet. The subnet would only have a single gateway.

The framework should "simulate" Fendermint. This sounds complex, but it's actually quite easy.
- The test base would offer methods to generate a bottom up checkpoint and commit it in the parent. 
- The test base would offer methods to draw a parent finality into the child subnet by pulling all messages.

## What we unlock

This would unlock test scenarios like:

1. Deploy IPC contracts in the parent and the subnet that perform xnet calls, make a call from the child to the parent (callee), assert on the parent contract, send a result, and assert the result was received in the child contract (caller).
2. Perform an xnet call from a contract in one subnet to an account in another subnet and verify that it simply transfers the value and moves on.
3. etc.

## Future

In the future, this could evolve into a "IPC minibox" framework that our users/adopters can use to write their IPC-aware contracts without having to deploy an actual subnet.

PR #631, opened 2024-01-25
Multisubnet test

dnkolegov:
Fixes https://github.com/consensus-shipyard/ipc/issues/628.


contracts/test/IntegrationTestBase.sol
```diff
@@ -152,6 +152,50 @@ contract TestSubnetActor is Test, TestParams {
...
+            bottomUpCheckPeriod: DEFAULT_CHECKPOINT_PERIOD,
+            majorityPercentage: DEFAULT_MAJORITY_PERCENTAGE,
+            activeValidatorsLimit: DEFAULT_ACTIVE_VALIDATORS_LIMIT,
+            powerScale: DEFAULT_POWER_SCALE,
+            permissionMode: PermissionMode.Collateral,
+            supplySource: SupplySource({kind: SupplyKind.ERC20, tokenAddress: tokenAddress})
+        });
+
+        return params;
+    }
```
raulk, line 197: Can't we consolidate these two funcitons and just take a SupplySource as a parameter? The caller can use `SupplySourceHelper.native()` for a shortcut.

contracts/test/IntegrationTestBase.sol
```diff
@@ -853,6 +909,7 @@ contract IntegrationTestBase is Test, TestParams, TestRegistry, TestSubnetActor,
 
     function registerSubnetGW(uint256 collateral, address subnetAddress, GatewayDiamond gw) public {
         GatewayManagerFacet manager = GatewayManagerFacet(address(gw));
+        GatewayGetterFacet getter = GatewayGetterFacet(address(gw));
```
raulk, line 912: I think this is starting to become a bit pedantic. I would like to remove these local variables going forward. How about we create a `GatewayFacetsHelper` library with a bunch of functions that take `GatewayDiamond` or `address` and return the right facet? We can then attach it to the type with `using` and it becomes much nicer.

```
library GatewayFacetsHelper {
    function getter(address gw) returns GatewayGetterFacet;
    function manager(address gw) returns GatewayManagerFacet;
}

// binding
using GatewayFacetsHelper for address;

// usage
gw.getter().getNetworkName()
```

This is a lot easier to read since you always use the name of the entity (gateway) so it's easier to follow what thing this is the getter to.

contracts/test/integration/MultiSubnet.t.sol
```diff
@@ -0,0 +1,684 @@
...
+
+import "forge-std/console.sol";
+
+contract MultiSubnetTest is Test, IntegrationTestBase {
+    using SubnetIDHelper for SubnetID;
+    using CrossMsgHelper for IpcEnvelope;
+
+    GatewayDiamond public rootGateway;
+    GatewayGetterFacet public rootGatewayGetter;
+    GatewayManagerFacet public rootGatewayManager;
```
raulk, line 46: With the sugar I proposed above, we can just have the rootGateway here, and methods can do: rootGateway.getter()... Much nicer!

contracts/test/integration/MultiSubnet.t.sol
```diff
@@ -0,0 +1,684 @@
...
+
+    GatewayDiamond public tokenSubnetGateway;
+    GatewayDiamond public nativeSubnetGateway;
+
+    address[] public nativeSubnetPath;
+    address[] public tokenSubnetPath;
+
+    SubnetID rootSubnetName;
+    SubnetID nativeSubnetName;
+    SubnetID tokenSubnetName;
```
raulk, line 59: For each subnet we have a:
- gateway address
- subnet actor (if not root net)
- name
- path

Instead of repeating all variables, how about we create a struct:

```solidity
struct TestSubnetDefinition {
    address gateway;
    address subnetActor;
    SubnetID id;
    // address[] path << can't we get this from the SubnetID?
}
```

we can then have much better scoping and namespacing:

```
// state variables
TestSubnetDefinition public root;
TestSubnetDefinition public native;
TestSubnetDefinition public token;

// usage
root.gateway.getter()
```

raulk (commented):
This looks really good, great job figuring this out. I just did a first pass and will do a second one later.

@dnkolegov if you agree with my suggestions above, it'll be best to address in a separate PR, as there's good opportunity to simplify all Solidity tests.

raulk:
@dnkolegov let's do it!

raulk (approved):

PR #633, opened 2024-01-26
Fix cannot join with init amount

cryptoAtwill:
After the subnet is bootstrapped and the `initial_balance` parameter is set, it will call `pre_fund` which will thrown an error as `pre_fund` only works if the subnet is not bootstrapped. 

This PR checks if the subnet is bootstrapped before sending the `initial_balance`.

It is still possible that after sending the `pre_fund` call and during the call, the subnet is bootstrapped. This should be rare and the CLI will ask the user to retry.

Closes ENG-815


ipc/cli/src/commands/subnet/join.rs
```diff
@@ -29,11 +30,17 @@ impl CommandLineHandler for JoinSubnet {
...
-                .pre_fund(subnet.clone(), from, f64_to_token_amount(initial_balance)?)
-                .await?;
+            if !provider.bootstrapped(&subnet).await? {
+                log::info!("pre-funding address with {initial_balance}");
+                provider
+                    .pre_fund(subnet.clone(), from, f64_to_token_amount(initial_balance)?)
+                    .await
+                    .map_err(|e| anyhow!("cannot perform prefund due to {e}, probably subnet is already initialized."))?;
+            } else {
+                log::info!("subnet already initialized, no need pre-funding");
```
aakoshh, line 42: Shouldn't it then do the regular fund? Isn't the point of the command that the validators intends to have some initial balance on the subnet, e.g. to be able to sign and broadcast their checkpoint signature? 

It will be difficult to automate this CLI if one has to look in the logs to see whether they ended up with a balance or have to do a followup.

On a different note, I'm wondering why the initial balances in the examples are so low, how do validators pay for checkpoint signatures with 1 atto?

ipc/provider/src/lib.rs
```diff
@@ -290,6 +290,16 @@ impl IpcProvider {
             .await
     }
 
+    pub async fn bootstrapped(&mut self, subnet: &SubnetID) -> anyhow::Result<bool> {
+        let parent = subnet.parent().ok_or_else(|| anyhow!("no parent found"))?;
+        let conn = match self.connection(&parent) {
+            None => return Err(anyhow!("target parent subnet not found")),
+            Some(conn) => conn,
+        };
```
aakoshh, line 298: I see this error handling appear ~11 times in this file. Can it be factored out?

ipc/cli/src/commands/subnet/join.rs
cryptoAtwill: This command is using one interface to both `join` and `preFund` from the contract. If the subnet is already boostrapped, ideally we should panic if `--initial-balance` is provided. The original intent did not throw an error, just performs a skip which is what this PR does.

I dont think one needs to look into the logs to see whether they ended up with a balance. As long as `join` succeeds, they will have some balance.

aakoshh: You may be right, I'm just trying to understand it from a perspective of someone who isn't intimately familiar with all the commands. 

As I understand the user calls something like `ipc-cli join --subnet-id /r123/t410a1...e9 --collateral 1000 --initial-balance 100 --public-key alice.sk`. 

Just looking at this what I imagine the end state should look like is that my account, Alice, will become a validator (candidate) on the subnet with 1000 collateral and 100 balance. 

Yet from the flow it looks like I may or may not have 100 balance, because if the subnet is bootstrapped it won't send the funds, only the collateral. 

As a user I would find that confusing. I know I need a balance on the subnet to pay for signing checkpoints, so I want that 100 balance there. And it looks like what we are saying is if that's true then I should go and call `fund` separately. 

Am I wrong? You say "they will have some balance". From where, how much is "some"?

cryptoAtwill: yeah, I think you are right. Maybe we should deprecated the `initial-balance` and just use `collateral`. Then in the CLI, we determine if the subnet is bootstrapped. If yes, then we call `join` and if not, then we call `pre_fund`, so that the user does not really need to know the details.

cryptoAtwill: I take a look at the contract, `preFund` and `join` actually are quite different in terms of user experience. I'm not sure why it's merged into one CLI command in the first place, but I think maybe we can break this into two command, `join` just join the subnet regardless of subnet state. `preFund` command just handles `pre-bootstrapped` balance.

aakoshh: I think the reason they are one is because from a user perspective the single command can express the intent quite clearly. If validators can join without funds and get the subnet bootstrapped, if they haven't got any funds on the subnet they won't be able to sign their checkpoints. It makes sense to say how much collateral and balance they want to commit when they join. I think this pre-fund / fund distinction is irrelevant to them. I don't know how it works in the contracts, but ideally it should be transparent as the desired end-state is quite clear.

cryptoAtwill: Ok, I see. I think the current way also works, with `pre_fund` and `join` separated. Now validators can call `pre_fund` repetitively to adjust the genesis fund and call `join` to determine the collateral. It's clearer than `--collateral` and `--initial-balance` in the sense that `pre_fund` actually adds the value provided instead of setting it, which the `--initial-balance` seems to suggest.

aakoshh: What does pre-fund do if the subnet is already bootstrapped? Does it turn transparently into a `fund`? 

Ultimately as a validator I just want to end up with collateral _and_ balance on the subnet, I don't want errors like "Sorry Bob, you should have called fund instead of prefund, because the other prefund sent by Alice already bootstrapped the subnet. Try again please. Oh, but you _are_ joined with 0 balance so you are part of the validator set, you just can't sign checkpoints, so do hurry up or you might be slahsed for inactivity!"

cryptoAtwill: Maybe we should merge `preFund` and `fund` into one function, depending on the bootstrap state.

cryptoAtwill:
@raulk There is no state diagram, but the idea is simple:
- If the address wants to `fund` but subnet is not bootstrapped, then one needs to call `preFund` and adds to `genesis`.
- Otherwise just call `fund`.

maciejwitowski:
Just a note that if we are thinking to modify the `fund`/`preFund` API duo, we should change `release`/`preRelease` as well.

ipc/cli/src/commands/subnet/join.rs
```diff
@@ -147,3 +138,50 @@ pub struct UnstakeSubnetArgs {
...
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(name = "pre-fund", about = "Pre fund a subnet")]
+pub struct PrefundArgs {
+    #[arg(long, short, help = "The address that joins the subnet")]
```
maciejwitowski, line 178: Should it say `joins` here?

ipc/cli/src/commands/subnet/join.rs
```diff
@@ -147,3 +138,50 @@ pub struct UnstakeSubnetArgs {
...
+    pub from: Option<String>,
+    #[arg(long, short, help = "The subnet to join")]
```
maciejwitowski, line 180: Should it say `join` here?

ipc/cli/src/commands/subnet/join.rs
maciejwitowski: > Maybe we should merge preFund and fund into one function, depending on the bootstrap state.
@cryptoAtwill you mean here merging `preFund` and `join`, right?

jsoares:
I don't want to keep a complete and relatively simple PR on hold, though I understand the UX questions raised. 

@raulk @aakoshh @cryptoAtwill We can carry this over to next week but let's try to agree on whether (i) this is a problem; (ii) this is the solution -- or close if the answer is no.

Whenever possible, I think it's best to look at it from the user perspective. Akosh & Willes's suggestion above makes sense, i.e. make it transparent to the user, possibly by merging `preFund` and `fund` into one function that checks for bootstrap state. Do we see any shortcomings with that route?

jsoares:
After discussing with Raúl, we're leaning towards closing and folding this into a bigger re-architecture, also tying in with the ERC-20 prefunding. Other ideas welcome, though.

Issue #615, opened 2024-01-24
Write an integration test which tests calling blockhash from a solidity contract 


fridrik01:
See https://github.com/consensus-shipyard/ipc/pull/568

Context: https://github.com/consensus-shipyard/ipc/pull/568#pullrequestreview-1841200112

PR #636, opened 2024-01-26
Add integration test for testing the chainmetadata actor

fridrik01:
See https://github.com/consensus-shipyard/ipc/issues/615

This PR adds a new QueryBlockhash contract which calls the blockhash() method (which ends up calling get_tipset_cid in fendermint which is stored from the actor state) and compares that to what Tendermint RPC returns.

Fixes ENG-596.


fendermint/eth/api/examples/query_blockhash.rs
```diff
@@ -0,0 +1,168 @@
...
+            .send_with_receipt()
+            .await
+            .context("failed to send deployment")?;
+
+    tracing::info!(addr = ?contract.address(), "QueryBlockhash deployed");
+
+    let contract = QueryBlockhash::new(contract.address(), contract.client());
+
+    // check that the blockhash returned by the contract matches the one returned by tendermint
+    for epoch in 1..=5 {
```
fridrik01, line 130: Is there possibility that this could be flaky?

fendermint/eth/api/examples/query_blockhash.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// Secret key used to deploy the contract.
+    ///
+    /// Assumed to exist with a non-zero balance.
+    #[arg(long, short)]
+    pub secret_key: PathBuf,
+
+    /// Path to write the contract metadata to.
+    #[arg(long, short)]
+    pub out: Option<PathBuf>,
```
aakoshh, line 59: You probably don't need this. 

I think with the `flatten` attribute in `clap` we could reuse most of these common options. Not sure what the syntax is now, but it should be possible.

fendermint/eth/api/examples/query_blockhash.rs
aakoshh: How do you mean?

fendermint/eth/api/examples/query_blockhash.rs
```diff
@@ -0,0 +1,168 @@
...
+        //
+        let b = request(
+            "eth_getBlockByNumber w/o txns",
+            provider
+                .get_block(BlockId::Number(BlockNumber::Number(U64::from(epoch))))
+                .await,
+            |b| b.is_some() && b.as_ref().map(|b| b.number).flatten() == Some(U64::from(epoch)),
+        )?;
+        let bh = b.unwrap().hash.expect("hash should be set");
+        tracing::info!("blockhash from actor:    {:?}", bh);
```
aakoshh, line 153:
```suggestion
        tracing::info!("blockhash from API:    {:?}", bh);
```

fendermint/testing/smoke-test/Makefile.toml
```diff
@@ -30,8 +28,11 @@ EOF
 
 [tasks.test]
 clear = true
-dependencies = ["simplecoin-example", "ethers-example"]
-
+dependencies = [
+  "simplecoin-example",
+  "ethers-example",
+  "query_blockhash-example",
```
aakoshh, line 34:
```suggestion
  "query-blockhash-example",
```

fendermint/testing/smoke-test/Makefile.toml
```diff
@@ -50,3 +51,11 @@ cargo run -p fendermint_eth_api --release --example ethers -- \
   --secret-key-from testing/smoke-test/test-data/keys/emily.sk \
   --secret-key-to   testing/smoke-test/test-data/keys/eric.sk
 """
+
+[tasks.query_blockhash-example]
```
aakoshh, line 55:
```suggestion
[tasks.query-blockhash-example]
```

aakoshh (approved):
Looks good :+1: 

Is it not working?

fridrik01:
> Looks good 👍
> 
> Is it not working?

Yes, it passes the tests locally, and hopefully on CI.. will see soon

fendermint/eth/api/examples/query_blockhash.rs
fridrik01: I mean could it happen that when this test runs, those epochs have not happened?

aakoshh:
For a moment I thought the reason could be that epoch 1 is the genesis block and it doesn't have `begin_block`, just `init_chain`, but I don't actually remember. Maybe `init_chain` produces a state that appears in the header of block 1 :thinking: At least that's how I remember it from the snapshot tests :dizzy_face:

fendermint/eth/api/examples/query_blockhash.rs
aakoshh: I doubt it because we have `cometbft-wait` that surely gives time for 5 blocks to be created, but as I commented further down the contract might have forgotten the first 5.

fridrik01:
yeah, the CI has some slow machines which caused the test to fail as the chainmetadata actor had removed them from the history, but this should be fixed now

fendermint/testing/contracts/QueryBlockhash.sol
```diff
@@ -0,0 +1,10 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
+
+contract QueryBlockhash {
+    constructor() {}
+
+    function getBlockhash(uint blockNumber) public view returns (bytes32) {
```
raulk, line 7: How about we make the argument an offset, call NUMBER, add the offset, then call BLOCKHASH? This would make the test less time sensitive. And you'd pass in a negative block number to look back (which means this argument type would have to be int)

fendermint/eth/api/examples/common/mod.rs
```diff
@@ -116,3 +117,34 @@ where
 
     Ok(call)
 }
+
+pub trait CheckResult {
+    fn check_result(&self) -> anyhow::Result<()>;
+}
+
+impl CheckResult for bool {
+    fn check_result(&self) -> anyhow::Result<()> {
```
raulk, line 126: Fun trick: this could be less verbose chaining then_some() and ok_or().

fendermint/testing/contracts/QueryBlockhash.sol
```diff
@@ -0,0 +1,10 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
+
+contract QueryBlockhash {
```
maciejwitowski, line 4: @fridrik01 If I'm not mistaken, this is your first Solidity contract, right? 🆒

fridrik01: yebb

fendermint/testing/contracts/QueryBlockhash.sol
fridrik01: I prefer absolute number here, it make comparing this result to the API simpler

PR #637, opened 2024-01-26
FIX: Fix paths in infra. Create the .env file.

aakoshh:
Fixes some paths in the infra scripts and adds the creation of the common `.env` file in the testnet, and the passing of persistent peers via env var in docker-compose.


raulk (approved):

aakoshh:
I know it's confusing that we are setting the peers in two different ways. In my opinion we should get rid of the `sed` approach and leave the `config.toml` well alone, setting everything through env vars.

@dnkolegov if you agree then I can add a follow-up PR to do it.

Issue #613, opened 2024-01-24
Follow-ups from GMP cleanup


raulk:
- [ ] In `sendContractXnetMessage`, stop using `IpcEnvelope` as the parameter and instead take `to` and `callMsg` only, using the passed-in value.
- [x] Consolidate the many `InvalidCrossMsg*` errors into a single `InvalidXnetMessage(string reason)`.
- [ ] Move the `IpcHandler` interface to a `shared` package, so that the IPC contracts are not accessing the SDK.
- [ ] Rename `IpcMsgKind#Transfer` to `IpcMsgKind#SystemTransfer`, to make it clear that these are system generated messages.
- [ ] Rename `IpcMsgKind#Call` to `IpcMsgKind#ActorCall` to represent that this is a call originating in an actor.

PR #640, opened 2024-01-29
consolidate cross message error types 

snissn:
consolidate all Invalid cross message error types to an error InvalidXnetMsg(string reason).

https://github.com/consensus-shipyard/ipc/issues/613


dnkolegov:
I did not understand the goal of this consolidation. Is it to decrease the number of lines in the code?
After this aggregation, it is tough to distinguish why the code fails. The changed tests in the PR show exactly that. 
So, for now, to reliably test the code, it is also necessary to check the returned string somehow.

@raulk, what do you think?

snissn:
I had similar feelings to @dnkolegov and @cryptoAtwill while coding this

raulk:
## Code

These errors are thrown from `sendContractXnetMessage`, a method that's part of our public API for IPC developers. This is how a try/catch block would look for a dev using this API, pretty unergonomic:

```solidity
try {
    gateway.sendContractXnetMessage(...);
} catch InvalidCrossMsgSender() {
   // ... do something
} catch InvalidCrossMsgDstSubnet() {
   // ... do something
} catch InvalidCrossMsgNonce() {
   // ... do something
} catch InvalidCrossMsgValue() {
   // ... do something
} catch InvalidCrossMsgKind() {
   // ... do something
}
```

Why force the dev to write 5 catch blocks to deal with a single circumstance: "IPC rejected your message because it failed validation"? There is nothing fundamentally different in the way you'd handle each of those errors -- the variants simply convey the problematic field -- a purely informational hint.

## Gas

Separately there is the question of gas efficiency. The implementation in this PR is not optimal from a gas perspective:

1. The "Invalid Cross Msg" string is redundant. The error _type_ is already named "InvalidCrossMsg". Why do we need to repeat this in the error string? At least, we should drop that substring.
2. Is there another way we can squeeze further gas efficiency? Yes: using an enum to signal the field we failed to validate.

I _think_ enums encode as uint8 in ABI encoding (worth checking), so by merging these error types into a single one with an enum field to supply the extra information, we are adding exactly one byte to the return value, and we're helping the dev save gas on contract deployment (shorter bytecode) and on execution (fewer conditions to evaluate).

raulk:
Hope this convinces you that this is a good idea @dnkolegov @cryptoAtwill @snissn!

contracts/src/errors/IPCErrors.sol
```diff
@@ -26,11 +26,7 @@ error InvalidBatchSource();
 error InvalidSubnetActor();
 error InvalidCollateral();
 error InvalidConfigurationNumber();
-error InvalidCrossMsgSender();
-error InvalidCrossMsgDstSubnet();
-error InvalidCrossMsgNonce();
-error InvalidCrossMsgValue();
-error InvalidCrossMsgKind();
+error InvalidXnetMessage(string reason);
```
dnkolegov, line 29: Should not we store all the reason strings here and then use them in tests and in the code?

dnkolegov (approved):

raulk:
In the interest of progress, I've made the changes I explained above. Looks like @dnkolegov found them convincing.

PR #645, opened 2024-01-30
Remove short names from all ipc-cli commands

fridrik01:
Issue: https://github.com/consensus-shipyard/ipc/issues/660

Noticed this error when following the quickstart calibration guide. Lets just remove the short names as we should always prefer the long names anyway for clarity

```bash
fridrik@workpc:~/workspace4/ipc-monorepo$ cargo run --bin ipc-cli subnet create --parent /r314159 --min-validators 3 --min-validator-stake 1 --bottomup-check-period 30
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/ipc-cli subnet create --parent /r314159 --min-validators 3 --min-validator-stake 1 --bottomup-check-period 30`
thread 'main' panicked at /home/fridrik/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.14/src/builder/debug_asserts.rs:112:17:
Command create: Short option names must be unique for each argument, but '-m' is in use by both 'min_validator_stake' and 'min_cross_msg_fee'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

### DIFF
--- a/docs/ipc/quickstart-calibration.md
+++ b/docs/ipc/quickstart-calibration.md
@@ -58,14 +58,14 @@ You'll need to create a set of wallets to spawn and interact of the subnet. Plea
 
 * Create the three different wallets
 ```bash
-ipc-cli wallet new -w evm
-ipc-cli wallet new -w evm
-ipc-cli wallet new -w evm
+ipc-cli wallet new -wallet-type evm
+ipc-cli wallet new -wallet-type evm
+ipc-cli wallet new -wallet-type evm
 ```
 
 * You can optionally set one of the wallets as your default so you don't have to use the `--from` flag explicitly in some of the commands:
 ```bash
-ipc-cli wallet set-default --address <DEFAULT_ETH_ADDR> -w evm
+ipc-cli wallet set-default --address <DEFAULT_ETH_ADDR> -wallet-type evm
 ```
 
 <!-- * Convert the 0x addresses to f4 addresses for later usage (OWNER_1_F4, OWNER_2_F4, and OWNER_3_F4)  -->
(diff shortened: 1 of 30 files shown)


raulk:
How about we remove all shorts all over the place? The UX of ipc-cli hasn't been thought out thoroughly, short flags are scarce, and we can't guarantee that more important options will emerge down the line that should take precedence for getting a short flag over the existing ones.

fridrik01:
> How about we remove all shorts all over the place? The UX of ipc-cli hasn't been thought out thoroughly, short flags are scarce, and we can't guarantee that more important options will not emerge down the line that should take precedence for getting a short flag over the existing ones.

@raulk Just so you are aware, that would change over 120 different commands in the codebase, let me know if you want to do that (its fine by me)

raulk:
Yep, I expect a big change here. @aakoshh WDYT?

fridrik01:
> Yep, I expect a big change here. @aakoshh WDYT?

Ok, I see it colliding in other commands, for example running `ipc-cli wallet export ` gives me:

```bash
Command export: Short option names must be unique for each argument, but '-h' is in use by both 'hex' and 'help' (call `cmd.disable_help_flag(true)` to remove the auto-generated `--help`)
```

I will go ahead and update this PR to remove short opts in all commands

jsoares:
I think we mostly moved away from short opts in the docs, but we should anyhow review things to make sure they don't break.

cc @trruckerfling

raulk:
> I think we mostly moved away from short opts in the docs, but we should anyhow review things to make sure they don't break.
> 
> cc @trruckerfling

https://linear.app/interplanetary-consensus/issue/GTM-26/review-docs-for-usage-of-short-flags-with-ipc-cli

docs/ipc/quickstart-calibration.md
```diff
@@ -58,14 +58,14 @@ You'll need to create a set of wallets to spawn and interact of the subnet. Plea
 
 * Create the three different wallets
 ```bash
-ipc-cli wallet new -w evm
-ipc-cli wallet new -w evm
-ipc-cli wallet new -w evm
+ipc-cli wallet new -wallet-type evm
+ipc-cli wallet new -wallet-type evm
+ipc-cli wallet new -wallet-type evm
```
raulk, line 63: Should be double dash, right? Elsewhere too...

aakoshh:
I have no problem removing short names from the `ipc-cli`. The ones I saw removed in this PR from the docs, I couldn't have guessed to save my life.

The `fendermint` CLI has short names too, but I tried not to use them in the docs that I wrote, and on some occasions it's not enabled when there is no obvious unique short variant, or the parameter doesn't occur enough time across many commands to warrant it. 

I wouldn't say "never use it because we might have a more important parameter later" because then, well, you never know, so you never use it.

PR #647, opened 2024-01-31
Setup object store actor

sanderpick:
This adds a new custom actor and wires it up into genesis. 

Note that you will have to run `cd contracts && npm install && make compile-abi` after cloning this repo to build without errors.


sanderpick:
Oops, classic... wrong target branch for an experiment :laughing:

PR #656, opened 2024-02-01
Make function calls atomic with ERC20 transfers in token supply source subnets

cryptoAtwill:
Make the `performCall` atomic with respect to ERC20 call with value.

Close ENG-635


contracts/src/lib/SupplySourceHelper.sol
```diff
@@ -101,13 +101,31 @@ library SupplySourceHelper {
...
+            // Perform the call only if the ERC20 was successful.
+            (success, ret) = functionCallWithValue(target, data, 0);
+        }
+        
+        if (!success) {
+            // following the implementation of `openzeppelin-contracts/utils/Address.sol`
+            assembly {
+                let returndata_size := mload(ret)
+                revert(add(32, ret), returndata_size)
             }
```
raulk, line 129: Could you comment why the `add` opcode? There may be comments in the OZ library, but let's avoid reader indirection.

cryptoAtwill: Ok, I will add comments on the code.

raulk: This is the full function from OpenZeppelin. We need to handle the case where there is no return data. The `add` is needed to unpack the variable-size return value (first Ethereum word is the size).

```solidity
function _revert(bytes memory returndata) private pure {
    // Look for revert reason and bubble it up if present
    if (returndata.length > 0) {
        // The easiest way to bubble the revert reason is using memory via assembly
        /// @solidity memory-safe-assembly
        assembly {
            let returndata_size := mload(returndata)
            revert(add(32, returndata), returndata_size)
        }
    } else {
        revert FailedInnerCall();
    }
}
```

contracts/test/unit/SupplySourceHelper.t.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import {SupplySourceHelperMock} from "../mocks/SupplySourceHelperMock.sol";
+
+import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
+
+import {ERC20PresetFixedSupply} from "../helpers/ERC20PresetFixedSupply.sol";
+
+contract SupplySourceHelperTest is Test {
+    /// Call fails but send value works, both should fail
+    function test_revert_atomicity() public {
```
raulk, line 19: Can we test `performCall` instead, please?

raulk (changes requested):

contracts/src/lib/SupplySourceHelper.sol
cryptoAtwill: From my testing, if the `ret` is empty, calling `revert(add(32, returndata), returndata_size)` is the same as revert with no data. I think openzeppelin might be emphasising `FailedInnerCall`? But for us, if the contract returns empty, we can just return empty. Even if we dont, after some parsing of the error bytes, users will still get an empty ret.

contracts/test/unit/SupplySourceHelper.t.sol
cryptoAtwill: added more tests

contracts/src/lib/SupplySourceHelper.sol
cryptoAtwill: From my testing, if the `ret` is empty, calling `revert(add(32, returndata), returndata_size)` is the same as revert with no data. I think openzeppelin might be emphasising `FailedInnerCall`? But for us, if the contract returns empty, we can just return empty. Even if we dont, after some parsing of the error bytes, users will still get an empty ret.

raulk: @cryptoAtwill I suspect you're seeing this because mloading the return data pointer will return the size, which will be 0. You then revert providing an offset equal to the pointer+32, **but with size 0**. This is probably OK, but I'd rather play it safe and do the check anyway in high level Solidity.

contracts/src/lib/SupplySourceHelper.sol
```diff
@@ -101,13 +102,36 @@ library SupplySourceHelper {
...
+            if (ret.length > 0) {
+                assembly {
+                    let returndata_size := mload(ret)
+                    // see https://ethereum.stackexchange.com/questions/133748/trying-to-understand-solidity-assemblys-revert-function
+                    revert(add(32, ret), returndata_size)
+                }
             } else {
-                return (success, ret);
+                revert FailedInnerCall();
             }
```
raulk, line 135: Let's remove the `else` and simply use `revert()`. If the target didn't provide an error, we should not add one into the mix; we're simply interested in passing through here, right?

raulk (approved):

PR #657, opened 2024-02-01
Fork the Ethereum Address Manager + add deployment permissioning

cryptoAtwill:
This one addresses "Add permissioning to the Ethereum Address Manager". 

The approach is pretty simple, it creates an HAMT that tracks the whitelisted deployers. Only when the `caller` is in the HAMT, will the `caller` be allowed to create contracts.  This PR does not copy paste the original `EAM` implementation but imports it, so that it's easier to maintain.

Follow up PR:
- [ ] Update genesis to take whitelisted deployers as argument.


snissn:
Hey! I was curious about this PR because I had done HAMT related work recently for a filecoin migration. The use of a HAMT makes sense to me in that these data structures should be relatively small and it's relatively common to use an associative data array with a null value as a set. I think the HAMT performance of using a null value over using a AMT which should be sufficient is a non issue here and any performance difference or storage overhead is nearly neglible. Still just wanted to share my reflection and feedback since in other scenarios it may be an important distinction to use an [AMT](https://docs.rs/fvm_ipld_amt/latest/fvm_ipld_amt/) as a set instead of an HAMT with has a value that is being ignored

snissn:
Also looks like there's a panic in a test 😄

cryptoAtwill:
@snissn Cool, thx for the tip. I used HAMT because there might be chances of dynamically changing the deployers, which HAMT might be better suited and more extendable.
I think the CI fails because I have yet to update the genesis deployment.

raulk (changes requested):
I'm glad this approach worked! Integrating into genesis should be pretty simple. You'd just place this actor instead of the upstream one at actor ID 10.

I wonder if you want to turn the state field into an enum of `Unpermissioned` and `Permissioned(Optional<Cid>)`.

The default would be `Unpermissioned` (behaving just like the upstream).

Then you have `Permissioned(None)`, where no one can deploy contracts.

And `Permissioned(Cid)` would contain the Hamt.

raulk:
Apologies, my finger slipped, I only wanted to submit comments, not reject it outright.

cryptoAtwill:
@raulk Yep, I tried that in the first place, but seems like `Serialize_tuple` does not support enum. I could implement it manually, but then I figured I can actually deploy the default EAM actor if the genesis does not require permission. Unless we want to stick with more complex permission mode, I think it should be good for now to handle this logic in genesis?

raulk:
Interesting. I recall seeing enum fields in state or in input/response messages in builtin-actors. Could you scan the codebase there? Maybe start with the miner actor as it's the most complex one.

fridrik01 (commented):
lgtm, but we should make sure that tests are passing before merging (its complaining that the new eam actor is not in the blockstore as it has not been loaded at genesis). Fine with addressin that in followup PR as you mention

contracts/binding/src/lib.rs
```diff
@@ -2,25 +2,31 @@
 #[macro_use]
```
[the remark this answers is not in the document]
fridrik01: I pushed a proposal [PR](https://github.com/consensus-shipyard/ipc/pull/659) which keeps this file formatted when creating this file in build.rs.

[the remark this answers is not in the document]
cryptoAtwill: somehow it's auto-formatted and I didnt aware it's pushed, I see if I can revert it.

fendermint/actors/eam/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn name() -> &'static str {
+        IPC_EAM_ACTOR_NAME
+    }
+
+    actor_dispatch_unrestricted! {
+        Constructor => constructor,
+        Create => create,
+        Create2 => create2,
+        CreateExternal => create_external,
+    }
```
aakoshh, line 98: If any and all methods need to be dispatched, isn't there a way not to have to enumerate them at all?

fendermint/actors/eam/src/state.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug, Clone)]
+pub struct State {
+    /// The allowed deployers of contract
+    pub deployers: Cid, // HAMT[Address]()
+}
+
+impl State {
+    pub fn new<BS: Blockstore>(store: &BS) -> Result<State, ActorError> {
+        let empty_deployers = DeployerMap::empty(store, DEFAULT_HAMT_CONFIG, "empty").flush()?;
```
aakoshh, line 20: What is this `"empty"` thing?

aakoshh (approved):
I have no problem removing short names from the `ipc-cli`. The ones I saw removed in this PR from the docs, I couldn't have guessed to save my life.

The `fendermint` CLI has short names too, but I tried not to use them in the docs that I wrote, and on some occasions it's not enabled when there is no obvious unique short variant, or the parameter doesn't occur enough time across many commands to warrant it. 

I wouldn't say "never use it because we might have a more important parameter later" because then, well, you never know, so you never use it.

fendermint/actors/eam/src/lib.rs
cryptoAtwill: We can create macro rules on top of this. I think this is probably the convention used and the actor is quite simple. I guess it's ok?

fendermint/actors/eam/src/state.rs
cryptoAtwill: It's an identifier for the `map`, it's not stored in the storage, but just in memory. I took this from the actor development convention. I think this is a identifier for the map or sth along that line.

fendermint/actors/eam/src/lib.rs
aakoshh: I mean, there must be some general method that would simply forward, like `handle_message(method_number, params)` ?

cryptoAtwill: yeah, there is, but `actor_dispatch_unrestricted` actually contains some checking logic as well.

aakoshh: The `ActorCode` trait has this `invoke_method` method that you can trivially implement for this actor, and then you don't need to implement all the individual methods, because they all check the same thing. Something like this:

```
        fn invoke_method<RT>(
            rt: &RT,
            method: fvm_shared::MethodNum,
            args: Option<fvm_ipld_encoding::ipld_block::IpldBlock>,
        ) -> Result<Option<fvm_ipld_encoding::ipld_block::IpldBlock>, ActorError>
        where
            RT: Runtime,
            RT::Blockstore: Clone,
        {
            Self::ensure_deployer_allowed(rt)?;
            EamActor::invoke_method(rt, method, args)
        }
```

aakoshh: I don't see any documentation on `actor_dispatch_unrestricted` so can't say what's the difference between that and `actor_dispatch`, but the only check I see it does is the method number. That will be performed by `EamActor::invoke_method` as well.

raulk: I agree with @aakoshh. Performing the permissioning at the invoke_method entrypoint should be enough (we should leave the Constructor) out of this extra layer of permissioning, since it can only be invoked by the system actor and it already performs that check.

https://github.com/filecoin-project/builtin-actors/blob/master/actors/eam/src/lib.rs#L237

raulk: This way we can get rid of every method here.

fendermint/actors/eam/src/state.rs
```diff
@@ -0,0 +1,116 @@
...
+            Deserialize::deserialize(serde_tuple::Deserializer(deserializer))?;
+        Ok(match mode {
+            NO_RESTRICTION_MODE => PermissionMode::NoRestriction,
+            WHITELIST_MODE => PermissionMode::Whitelist(
+                Cid::from_bytes(&bytes).map_err(|_| Error::custom("invalid cid"))?,
+            ),
+            _ => return Err(Error::custom("invalid permission mode")),
+        })
+    }
+}
```
raulk, line 95: You should not be needing the custom serde here! According to @vmx [here](https://filecoinproject.slack.com/archives/C015KQQLQQ1/p1707317303298409), libipld_serde_cbor supports externally tagged enums (not clear to me how it lays them out though), as well as `Option` types. We might need to add the #[repr] tag so serde knows which type to use for the discriminator. Perhaps that's why it didn't work for you originally?

fendermint/vm/genesis/src/lib.rs
```diff
@@ -40,6 +40,9 @@ pub struct Genesis {
     /// highest possible fidelity when we are deriving a genesis file in IPC,
     /// where the parent subnet tracks collateral.
     pub validators: Vec<Validator<Collateral>>,
+    /// List of contract addresses that are allowed to deploy contracts in the subnet. If
+    /// empty, indicates there is no restriction on who can deploy contracts
+    pub contract_deployers: Vec<SignerAddr>,
```
raulk, line 45: Let's roll back the genesis changes here, and instead have users who want to use this feature **fork Fendermint** and customize genesis within init. Rationale: there will be plenty of customizations possible down the line, and it won't scale to expose them all through genesis. In the future, these will be Wasm modules.

raulk (changes requested):

fendermint/actors/eam/src/state.rs
```diff
@@ -0,0 +1,116 @@
...
+
+pub type DeployerMap<BS> = Map2<BS, Address, ()>;
+
+/// The permission mode for controlling who can deploy contracts
+#[derive(Debug, Clone, PartialEq)]
+enum PermissionMode {
+    /// No restriction, everyone can deploy
+    NoRestriction,
+    /// Only whitelisted addresses can deploy
+    Whitelist(Cid), // HAMT[Address]()
```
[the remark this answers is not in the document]
cryptoAtwill: why do we use `Option` here? We expect it to be no one can deploy contracts? If efficiency/gas cost is not a concern here, I think using `empty Cid` will make code implementation easier without checking the Option.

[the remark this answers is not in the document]
cryptoAtwill: why do we use `Option` here? We expect it to be no one can deploy contracts? If efficiency/gas cost is not a concern here, I think using `empty Cid` will make code implementation easier without checking the Option.

fendermint/vm/genesis/src/lib.rs
cryptoAtwill: But for now, how are they going to config this?

cryptoAtwill: that means we put the addresses in the source code?

fendermint/actors/eam/src/lib.rs
cryptoAtwill: Updated accordingly

Cargo.toml
```diff
@@ -117,6 +118,7 @@ quickcheck_macros = "1"
 rand = "0.8"
 rand_chacha = "0.3"
 regex = "1"
+rlp = { version = "0.5.1", default-features = false }
```
raulk, line 121: Unused?

raulk (approved):

Cargo.toml
cryptoAtwill: removed

cryptoAtwill: removed

fendermint/actors/eam/src/state.rs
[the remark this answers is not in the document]
raulk: Makes sense.

raulk (approved):

PR #659, opened 2024-02-01
Keep binding/src/lib.rs codegen file formatted

fridrik01:
Proposal: Lets updated the `build.rs` script to format the `binding/src/lib.rs` file to make sure we don't accidentally format it which causes PR noise/merge conflicts.


aakoshh (approved):
I still find it odd that this formatting happens, when this has an explicit exception set for it in `.rustmft` but okay, thanks!

PR #662, opened 2024-02-02
Plug fevm-contract-tests ERC20 and SimpleCoin tests into CI

snissn:
closes ENG-644

We have the fevm-contract-tests repo and github action https://github.com/consensus-shipyard/fevm-contract-tests/blob/main/.github/workflows/ci.yaml that runs the following:

- check out fevm-contracts repo
- fetches and runs latest fendermin docker container from  ghcr.io/consensus-shipyard/fendermint
- Run available solidity test suites on fendermint 
      - Run tests: web3.js SimpleCoin
      - Run tests: web3.js ERC20
      - Run tests: ethers.js SimpleCoin
      - Run tests: ethers.js ERC20


maciejwitowski:
@snissn [The latest job](https://github.com/consensus-shipyard/fevm-contract-tests/actions/runs/7753559950) in fevm-contracts-tests took 30min. Adding such a long test to every PR seems like too much, no?

.github/workflows/ci.yaml
```diff
@@ -107,6 +107,18 @@ jobs:
       needs.changes.outputs.fendermint == 'true' ||
       github.ref == 'refs/heads/main'
 
+  fevm-contract-tests:
```
aakoshh, line 110: Do you think this should be a precondition for publishing?

.github/workflows/fevm-contract-tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - name: 'Create and fund accounts'
+        id: accounts
+        env:
+          NUM_ACCOUNTS: 5
+          ROOT_PRIVATE_KEY: ${{ steps.testnode.outputs.ROOT_PRIVATE_KEY }}
+        working-directory: fevm-contract-tests
+        run: |
+          npx hardhat --network local create-fund-accounts
+      - name: 'Run tests: web3.js SimpleCoin'
+        if: always()
```
aakoshh, line 96: Why are these `if: always()` conditions needed?

.github/workflows/fevm-contract-tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+        timeout-minutes: 2
+        working-directory: fevm-contract-tests
+        env:
+          ROOT_PRIVATE_KEY: ${{ steps.testnode.outputs.ROOT_PRIVATE_KEY }}
+          USER_1_PRIVATE_KEY: ${{ steps.accounts.outputs.ACCOUNT1_PRIVATE_KEY }}
+        run: |
+          npx hardhat --network local test ./test/ethers.js/ERC20.js
+      - name: 'Run tests: Uniswap -- SKIPPING'
+        if: always()
```
aakoshh, line 131: Why not something like `if: never()` and then CI would say `Skipped` on its own?

.github/workflows/fevm-contract-tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+        working-directory: fevm-contract-tests
+        env:
+          USER_1_PRIVATE_KEY: ${{ steps.accounts.outputs.ACCOUNT1_PRIVATE_KEY }}
+          USER_2_PRIVATE_KEY: ${{ steps.accounts.outputs.ACCOUNT2_PRIVATE_KEY }}
+          USER_3_PRIVATE_KEY: ${{ steps.accounts.outputs.ACCOUNT3_PRIVATE_KEY }}
+          USER_4_PRIVATE_KEY: ${{ steps.accounts.outputs.ACCOUNT4_PRIVATE_KEY }}
+          USER_5_PRIVATE_KEY: ${{ steps.accounts.outputs.ACCOUNT5_PRIVATE_KEY }}
+        run: |
+          echo skipping ; #cd ./extern/openzeppelin-contracts/ && yarn && mkdir gasreport && npx hardhat --network itest test
```
aakoshh, line 150: It would be great to have a comment to explain why this and the uniswap one are skipped, so we have an idea when and how to enable them. Maybe a ticket number?

.github/workflows/fevm-contract-tests.yaml
```diff
@@ -0,0 +1,150 @@
+
```
raulk, line 1: Empty line

raulk (changes requested):
This took 27min to run and became the blocking job of the full pipeline. And we're only running the ERC20 and SimpleCoin tests (which shouldn't take more than 1 minute).

**Ideas:**

* Are we able reuse the Docker image the CI pipeline produces? Why do we need to build it from scratch?
* Do we need to run this on every PR? Would it be sufficient to run on master nightly, and automate opening a Linear issue when it fails?
* What is the plan for enabling the OpenZeppelin and Uniswap tests? Those are the ones that give us most coverage. Can you link the plan here?

Nit: in the future, let's please use more descriptive names for PR titles. "Plug fevm-contract-tests into CI" would be shorter and would convey a lot more info than the original title.

.github/workflows/fevm-contract-tests.yaml
snissn: I looked into that and if:never() is not a feature offered by github - I think it makes sense to remove the steps from this PR that are skipped and opened https://linear.app/interplanetary-consensus/issue/ENG-688/remove-skipped-tests-uniswap-and-openzeppelin-from-pr

.github/workflows/fevm-contract-tests.yaml
snissn: In the ongoing future now that we are storing all information on Linear we should be able to find their key information in the linked linear issues ENG-641 and ENG-642. For now however most discussion about openzeppelin and uniswap is on slack - https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1706266000372379 is the most recent

snissn: I have removed those sections from the yaml file 09f0d9bdbaa06fbe5a5f89d2a5662ca789876d82 with the comment `Removing uniswap and openzeppelin tests. Intention is to bring them in future prs. Documented in ENG-641 and ENG-642.`

snissn:
I think that there's two main feedback items. 
1) the test takes 30 minutes to run and that is not acceptable. 

I see that there was confusion that we should be able to resolve. This was assigned during the monorepo switch over time period and the fendermint docker images were not up to date. As a work around, @aakoshh gave me instructions to run docker build in my script. Now however in review it is clear to me that the fendermint-publish action builds a new docker image that this test can use. @aakoshh can you provide instructions for how to use the docker image that the fendermint-publish action publishes? I want to ensure I use the most current source location for our docker images.

2) Ensure that a plan for enabling the OpenZeppelin and Uniswap tests exists and is documented.
[ENG-196](https://linear.app/interplanetary-consensus/issue/ENG-196/integrate-fevm-contract-tests-in-ci) documents everything as sub issues

aakoshh:
It is true that there are some inefficiencies in testing the docker image at the moment:
1. Much like in your case, they are built automatically [during e2e testing](https://github.com/consensus-shipyard/ipc/blob/1ea3558303e22a0af84536504d45433a7fa154f7/.github/workflows/fendermint-test.yaml#L52), however that is not persisted anywhere
2. The [publish phase](https://github.com/consensus-shipyard/ipc/blob/main/.github/workflows/fendermint-publish.yaml) only runs on `main` and builds it again, after the test phase has finished. 

So on `main` the image is built twice, and now with your tests thrice. 

__Corollary__: I tried to speed up the [docker build on CI](https://github.com/consensus-shipyard/ipc/blob/main/fendermint/docker/builder.ci.Dockerfile) by splitting it into two phases: 1) building just the dependencies and 2) building the whole project, hoping that the 1st step would create a docker layer that doesn't change between builds. However, if it's not present on the runner machine then it will be built from scratch, and I think we could ensure that it is available by pulling the [latest image](https://github.com/consensus-shipyard/ipc/pkgs/container/fendermint). 

If we wanted to be more parsimonious with our builds then we have to change our workflow:
  * In PRs, we have to separate the docker build to be distinct from the tests and publish it tagged by the PR number. We have to have something that removes this image after the PR is merged. 
  * We could use a _label_ to enable long running tests on PRs.
  * On `main`, we have to publish with a special tag that is still under testing, test it, then re-tag to `latest`.

Alternatively, ss Raul suggested we could run this particular test on `main` after the merge, even on the published image. The problem is obviously that you wouldn't be able to test specific feature branches, which is why I brought up the option of labelling above.

maciejwitowski:
@snissn As soon as you're able to context switch from USDC (which is the highest priority now), let's jump back to this PR and merge it. We'll need 2 things:
- Reuse the Docker image to speed up this job
- Run this job only nightly on the `main` branch

We can then think about running more often, enable testing feature branches with it etc. For now let's set the goal of running it _at all_.

snissn:
> @snissn As soon as you're able to context switch from USDC (which is the highest priority now), let's jump back to this PR and merge it. We'll need 2 things:
> 
> * Reuse the Docker image to speed up this job
> * Run this job only nightly on the `main` branch
> 
> We can then think about running more often, enable testing feature branches with it etc. For now let's set the goal of running it _at all_.

Hi @maciejwitowski I have updated the PR with 
> * Reuse the Docker image to speed up this job

for 
> * Run this job only nightly on the `main` branch

Given that [now the test runs in around 2 minutes](https://github.com/consensus-shipyard/ipc/actions/runs/7852699316/job/21431457393?pr=662), it makes sense to me to merge this as it currently is so that it runs on each commit. If that's ok with everyone I think we can do a final review and merge.

.github/workflows/ci.yaml
snissn: Hey! @aakoshh that is a great idea! I think that it depends on how long the tests take to run. If they're down to 2 minutes I think we can run them more often than just before publishing.

.github/workflows/fevm-contract-tests.yaml
snissn: This is a preference that @raulk expressed to me but I think his idea is that we want to ensure that all the tests run to ensure we have as much informat as possible in the event we need to debug a problem

maciejwitowski:
@snissn 2 minutes, that's nice! LGTM, I agree with merging this PR as is.

maciejwitowski:
@raulk Are you OK with merging this?

raulk (approved):
This is a great start! We really need to get the OpenZeppelin and Uniswap tests working, since those deliver the most diverse coverage.

Issue #663, opened 2024-02-02
We should go over all our bash scripts and make sure that these are the default


snissn:
We should go over all our bash scripts and make sure that these are the default

_Originally posted by @fridrik01 in https://github.com/consensus-shipyard/ipc/pull/652#discussion_r1472528624_
            
Follow up of ENG-626

PR #664, opened 2024-02-02
 add set -eu and pipefail in bash scripts

snissn:
Fixes https://github.com/consensus-shipyard/ipc/issues/663 ENG-651

ensure bash scripts exit whenever a sub process exits or an unused variable is accessed

### DIFF
--- a/contracts/ops/chain-id.sh
+++ b/contracts/ops/chain-id.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -eu
+set -o pipefail
 
 # Check if RPC_URL is set
 if [[ -z "$RPC_URL" ]]; then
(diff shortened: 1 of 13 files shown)


contracts/ops/chain-id.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -eu
+set -o pipefail
```
fridrik01, line 3: nit: we can do this in one line `set -euo pipefail`

fridrik01 (approved):
lgtm, thanks for adding this!

contracts/ops/chain-id.sh
snissn: that is cool! I did not know that! next time 😉

PR #665, opened 2024-02-02
ENG-578: IT (Part 1) - Manifest

aakoshh:
Creates a `fendermint_testing_materializer` create, currently with just a `Manifest` type with some example JSON/YAML/TOML files and random generation, to circulate as an idea for how we could define a testnet.

The properties I added are just a baseline, we can add stuff as we go, like block intervals, checkpoint intervals, etc.

The idea is that there will be a `Testnet` type that parses and validates a `Manifest`, and then subsequently is able to generate a set of steps that can be executed by a `Materializer` to provision the thing. The `Testnet` will have types like nodes, relayers, whatever we need to talk to. The `Materializer` will have a log that it can use to figure out where it deployed things and how it can execute commands on the resources in the testnet. In the `Testnet` everything will have a unique resource name. 

I added TOML and JSON examples but I think YAML is the most readable and it also supports comments, which might be handy in a manifest. The generated TOML is of course not the only way to do it, you could craft one manually and indent it, but why bother when we have YAML (as long as there is no hard-to-spot indent problems with yaml :confounded:)

TODO: 
- [x] Add nodes
- [x] Add relayers
- [x] TOML support


fridrik01 (approved):
This looks great! 

I like the manifest abstraction and the random generation implementation of it. We can later extend that with builder pattern if needed when we want to test a particular configuration in code.

fendermint/testing/materializer/golden/manifest/toml/manifest.toml
```diff
@@ -0,0 +1,133 @@
...
+
+[subnets.IAYNV.nodes.1ah.mode]
+Validator = "OK5OS6"
+
+[subnets.IAYNV.nodes.yB-S9cL]
+ethapi = true
+seed_nodes = []
+
+[subnets.IAYNV.nodes.yB-S9cL.mode]
+Validator = "My2Bn2XY"
```
raulk, line 37: Why the capital letter here?

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+}
+
+fn gen_manifest(
+    g: &mut Gen,
+    max_children: usize,
+    max_level: usize,
+    default_balance: Balance,
+) -> Manifest {
+    let account_ids = (0..3 + usize::arbitrary(g) % 3)
```
raulk, line 115: Should we supply the number of accounts externally?

fendermint/testing/materializer/src/manifest.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/// The rootnet, ie. the L1 chain, can already exist and be outside our control
+/// if we are deploying to Calibration, or it might be a chain we provision
+/// with CometBFT and Fendermint.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum Rootnet {
+    /// Existing L1 running outside our control.
+    ///
+    /// This implies using some sort of Faucet to get balances for the accounts.
```
raulk, line 104: Can we record these requirements somewhere so we don't surprise ourselves at a later stage, when coming back to this?

fendermint/testing/materializer/src/manifest.rs
```diff
@@ -0,0 +1,223 @@
...
+
+/// An IPC subnet.
+///
+/// The balance of the account on the parent subnet, as declared in this manifest,
+/// _does not_ have to account for the collateral/balance we have to take from it to join/fund the subnet.
+/// When we create the testnet configuration from the manifest we will account for this with a rollup,
+/// so we don't have to do that much mental arithmetic and run into frustrating errors during setup.
+/// If we want to test trying to join with more than what we have, we can do so in the integration test.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Subnet {
```
raulk, line 134: I think it's fine to go without bootstrappers for now, assuming we're bootstrapping in daisy chain, but we may want to add a TODO for a bootstrap strategy? Or, we could add it here and only allow daisy chaining as an option for now.

fendermint/testing/materializer/golden/manifest/toml/manifest.toml
```diff
@@ -0,0 +1,133 @@
...
+
+[accounts.OK5OS6]
+
+[accounts.Z66]
+
+[accounts.wk8j]
+
+[accounts.z_MP]
+
+[rootnet.External.deployment.Existing]
```
raulk, line 11: Is there a way to get rid of this uppercase and lowercase concoction?

fendermint/testing/materializer/src/manifest.rs
```diff
@@ -0,0 +1,223 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Node {
+    /// Indicate whether this is a validator node or a full node.
+    pub mode: NodeMode,
+    /// Indicate whether to run the Ethereum API.
+    pub ethapi: bool,
+    /// The nodes from which CometBFT should bootstrap itself.
+    pub seed_nodes: Vec<NodeId>,
+    /// The parent node that the top-down syncer follows;
+    /// or leave it empty if the parent is CalibrationNet.
```
raulk, line 164: Why do we sneakily assume Calibnet here when empty? How can we model this explicitly?

fendermint/testing/materializer/src/manifest.rs
```diff
@@ -0,0 +1,223 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Relayer {
+    /// The account which will pay for the submission on the parent subnet.
+    pub submitter: AccountId,
+    /// The node which the relayer is following on the subnet.
+    pub follow_node: NodeId,
+    /// The node where the relayer submits the checkpoints;
+    /// or leave it empty if the parent is CalibrationNet.
+    pub submit_node: Option<NodeId>,
```
raulk, line 185: How would we handle the case of an external parent? e.g. Calibnet?

raulk (approved):
Solid start! The Manifest is really shaping up as a TopologyDescriptor. I wonder if the Materializer should take a TopologyDescriptor and _return_ a Manifest?

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+                (a, b)
+            })
+            .collect();
+
+        // Run at least a quroum of validators.
+        let tw: TokenAmount = vs.values().map(|c| c.0.clone()).sum();
+        let qw = tw.mul(2).div_floor(3);
+        let mut ss = Vec::new();
+        let mut ns = NodeMap::default();
+        let mut sw = TokenAmount::zero();
```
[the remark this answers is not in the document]
aakoshh: Yes I agree. I added comments later because I couldn't remember either. I'll rename them the followup PR, thanks

fendermint/testing/materializer/golden/manifest/yaml/manifest.yaml
```diff
@@ -0,0 +1,123 @@
+accounts:
```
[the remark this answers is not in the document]
aakoshh: Thanks. I'm not sure if we have to pick one or we can support multiple ones based on detecting the extension. I expect the authoring experience with YAML to be nicer, but I had episodes of screaming at the computer when a Github Action had a mismatch in the indent and just said "invalid syntax".

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+    //
+    // So we roll the dice and with 30% probability we include the 1st item in the list.
+    //
+    // Then we have two cases to consider:
+    // 1. We included the 1st item, so now we have 2 slots and remaining 9 items to choose from.
+    //    P_incl(2nd | incl(1st)) = 1 - 8/9 * 7/8 = 1 - 7/9 = 2/9
+    // 2. We excluded the 1st item, so we still have 3 slots to fill and remaining 9 items to choose from.
+    //    P_incl(2nd | excl(1st)) = 1 - 8/9 * 7/8 * 6/7 = 1 - 6/9 = 3/9
+    //
+    // Thus, the probability of including each item is `remaining slots / remaining items`
```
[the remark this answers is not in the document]
aakoshh: Haha, it was also me working through what I wanted to do. I remember this as a job interview question in a book, but never had a reason to figure it out. I _think_ it's right, but I don't know why https://en.wikipedia.org/wiki/Reservoir_sampling is comparatively so complicated.

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+        if remaining_slots == 0 {
+            break;
+        }
+        if usize::arbitrary(g) % remaining_items < remaining_slots {
+            chosen.push(x.clone());
+            remaining_slots -= 1;
+        }
+        remaining_items -= 1;
+    }
+    chosen
```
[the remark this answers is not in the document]
aakoshh: It might not be 0, if we ended up quickly filling our slots.

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+const RESOURCE_ID_CHARSET: &[u8] =
+    b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+
+lazy_static! {
+    /// Assume we have as much tFIL on the root as the faucet would give.
+    static ref DEFAULT_BALANCE: Balance = Balance(TokenAmount::from_whole(100));
+}
+
+/// Select some items from a slice.
+fn choose_at_least<T: Clone>(g: &mut Gen, min_size: usize, xs: &[T]) -> Vec<T> {
```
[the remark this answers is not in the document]
aakoshh: The size is given by `let mut remaining_slots = min_size + usize::arbitrary(g) % (xs.len() - min_size);` so we can expect it to be a uniform distribution between 1 and 10 items, with an expected value of (1+10)/2 = 5.5 items.

fendermint/testing/materializer/src/arb.rs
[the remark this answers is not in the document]
aakoshh: I'm really hoping that with 3-8 random characters chosen from the 28 available characters we don't see duplicate IDs in the arbitrary generation. 

Later a lot of the data structures (even this one) are going into `BTreeMap<ID, _>` so if there was a duplicate they would overwrite each other.

aakoshh: I don't think it makes too much a difference at this point. If we end up calling this with the intent that we want exactly N account, sure, but right now I just wanted some... I probably chose 3 because that is kind of the minimums of a non-standalone subnets with no tolerance for byzantine members, and then max 5 so we don't end up with no reuse, but like I say I don't use this for anything at the moment, it doesn't affect the general shape of the testnet like the other settings.

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+            },
+        }
+    } else {
+        let initial_balances = balances.clone();
+        // Reuse the subnet generation logic for picking validators and nodes.
+        let subnet = gen_subnets(g, 0, 2, 2, &account_ids, &[], &mut balances);
+        let subnet = subnet
+            .into_iter()
+            .next()
+            .expect("should have exactly 1 subnet")
```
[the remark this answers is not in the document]
aakoshh: Ah yes, it doesn't really matter, I just wanted to pick some validators and nodes and was like: argh, it's all there in that method, let's just call it. I'll revisit this to make it easier to understand.

fendermint/testing/materializer/src/manifest.rs
aakoshh: These are part of the testnet setup in the followup PR, it won't be a surprise.

fendermint/testing/materializer/src/manifest.rs
aakoshh: Good question, it would be an external IP address of an Ethereum JSON-RPC endpoint, so everyone can specify their favourite parent node instance. AFAIK the glif.io or address that we tend to use in tests has load balancing, so it's the same for every node (which is wrong). 

We can add an enum to capture this.

fendermint/testing/materializer/src/manifest.rs
aakoshh: Currently I thought the tester can just know the load balanced glif.io endpoint, but we can make it an explicit part of the manifest using a JSON-RPC endpoint, configured individually on any node that talks to an external address.

fendermint/testing/materializer/golden/manifest/toml/manifest.toml
aakoshh: We can move them inside the struct, in which case it would still be uppercase but wouldn't be a key: https://serde.rs/enum-representations.html

So instead of this
```json
"rootnet": { 
  "External": {
    "deployment": {
      "Existing": {
        "gateway": "...", 
        "registry": "..."
      }
    }
  }
}
```
it would be like:
```json
"rootnet": { 
    "type": "External",
    "deployment": {
        "type": "Existing",
        "gateway": "...", 
        "registry": "..."
    }
}
```

I'm 90% sure there is also a way to apply some [attribute](https://serde.rs/container-attrs.html) to `lowercase`, but I'm not sure. If it would be internally tagged, I think that wouldn't be necessary. 

Would you prefer internal, or external lowercase?

aakoshh:
> The Manifest is really shaping up as a TopologyDescriptor. I wonder if the Materializer should take a TopologyDescriptor and return a Manifest?

Thanks for these suggestions. The naming was influenced by CometBFT, but there are significant differences at the moment:
* The CometBFT [Manifest](https://github.com/cometbft/cometbft/blob/main/test/e2e/pkg/manifest.go) is the static description of a testnet that you'd either author by hand, or [generate randomly](https://github.com/cometbft/cometbft/blob/main/test/e2e/generator/generate.go)
* The CometBFT [Testnet](https://github.com/cometbft/cometbft/blob/main/test/e2e/pkg/testnet.go) parses the manifest and chooses further random (but deterministic) values for the fields which aren't part of the manifest or are left empty. This is for repeatability in tests. It generates values for IP addresses and ports as well, which then the infrastructure has to allocate accordingly.

Our Manifest is the same as theirs, but the testnet is completely different, it delegates everything about resource allocation to the Materializer and doesn't have any randomness in it, it just drives the process of the order of calling the operations. 

My intention with the Materializer is that one should be able to implement it concentrating on smaller actions, no procedure, just one thing at a time, such as allocating a node, creating a key, joining a subnet - the only thing they should ensure is repeated executions being idempotent, so some kind of persistent run log. 

This could be implemented in the Materializer trait; at the moment it's in the Testnet but it rhymes with "take a topology and return a ~~Manifest~~ Testnet".

fendermint/testing/materializer/src/arb.rs
```diff
@@ -0,0 +1,370 @@
...
+        .iter()
+        .map(|id| (id.clone(), default_balance.clone()))
+        .collect();
+
+    let rootnet = if bool::arbitrary(g) {
+        Rootnet::External {
+            deployment: if bool::arbitrary(g) {
+                IpcDeployment::Existing {
+                    gateway: ArbAddress::arbitrary(g).0,
+                    registry: ArbAddress::arbitrary(g).0,
```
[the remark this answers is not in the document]
aakoshh: Updated to use the `ethers` type.

fendermint/testing/materializer/golden/manifest/toml/manifest.toml
aakoshh: Switched to internal tagging.

fendermint/testing/materializer/golden/manifest/toml/manifest.toml
aakoshh: Switched to internal.

fendermint/testing/materializer/src/manifest.rs
aakoshh: Added a `ParentNode` type for this.

fendermint/testing/materializer/src/manifest.rs
aakoshh: Added the `ParentNode` enum type.

PR #667, opened 2024-02-03
fix: infra script that deploy a testnet

fridrik01:
# Context
Was trying to figure learn exactly how our our infra scripts work together, and noticed that deployign a local testnet was borked after we migrated to monorepo.

### Bring up the testnet
```bash
cd fendermint/infra
cargo make testnet
```

### Got a set of 4 nodes
```
docker ps
```
image

### Take it down:
```
cargo make testnet-down
```

PR #668, opened 2024-02-05
Revert "fix: infra script that deploy a testnet"

fridrik01:
Reverts consensus-shipyard/ipc#667

There was a failing test, we should update Github to block merge to main if there are CI errors


fridrik01:
Should be fixed with https://github.com/consensus-shipyard/ipc/pull/669

PR #669, opened 2024-02-05
Remove Cargo project from infra/

fridrik01:
This PR should fix the regression caused by merging https://github.com/consensus-shipyard/ipc/pull/667 which caused our fendermint/docker/builder.ci.Dockerfile to fail as it ignored the infra path. As we don't need to have a crate here (now at least), lets just remove it.


aakoshh (approved):
Cheers :+1:

raulk (approved):

PR #670, opened 2024-02-05
Fix minor findings

dnkolegov:
This PR fixes minor findings:

1. Renaming variables and using `revert` for consistency.
2. Adding `nonReentrant` for IPC calls.
3. Checking return value from Set calls.

### DIFF
--- a/contracts/src/gateway/GatewayManagerFacet.sol
+++ b/contracts/src/gateway/GatewayManagerFacet.sol
@@ -8,7 +8,7 @@ import {IpcEnvelope} from "../structs/CrossNet.sol";
 import {FvmAddress} from "../structs/FvmAddress.sol";
 import {SubnetID, Subnet, SupplySource} from "../structs/Subnet.sol";
 import {Membership, SupplyKind} from "../structs/Subnet.sol";
-import {AlreadyRegisteredSubnet, AlreadyInSet, NotInSet, CannotReleaseZero, MethodNotAllowed, NotEnoughFunds, NotEnoughFundsToRelease, NotEnoughCollateral, NotEmptySubnetCircSupply, NotRegisteredSubnet, InvalidXnetMessage, InvalidXnetMessageReason} from "../errors/IPCErrors.sol";
+import {AlreadyRegisteredSubnet, CannotReleaseZero, MethodNotAllowed, NotEnoughFunds, NotEnoughFundsToRelease, NotEnoughCollateral, NotEmptySubnetCircSupply, NotRegisteredSubnet, InvalidXnetMessage, InvalidXnetMessageReason} from "../errors/IPCErrors.sol";
 import {LibGateway} from "../lib/LibGateway.sol";
 import {SubnetIDHelper} from "../lib/SubnetIDHelper.sol";
 import {CrossMsgHelper} from "../lib/CrossMsgHelper.sol";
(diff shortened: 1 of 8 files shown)


contracts/src/lib/LibGateway.sol
```diff
@@ -505,8 +505,10 @@ library LibGateway {
             // Check the next subnet (which can may be the destination subnet).
             reject = to.down(s.networkName).getActor().hasSupplyOfKind(SupplyKind.ERC20);
         }
-        if (reject && crossMessage.kind == IpcMsgKind.Transfer) {
-            revert MethodNotAllowed("propagation of `Transfer` messages not suppported for subnets with ERC20 supply");
+        if (reject) {
```
maciejwitowski, line 508: What is the reason behind splitting those `if` statements?

contracts/src/gateway/GatewayManagerFacet.sol
```diff
@@ -114,7 +118,11 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
 
         s.totalSubnets -= 1;
         delete s.subnets[id];
-        s.subnetKeys.remove(id);
+
+        bool removed = s.subnetKeys.remove(id);
+        if (!removed) {
+            revert NotInSet();
```
maciejwitowski, line 124: What is the scenario when this happens?

dnkolegov: A catastrophic scenario is when the integrity of the actors is violated. So it is the same as panicking in Go in case we do not know what to do.

maciejwitowski: I see your point. However, IMO we should be handling cases for which we see _any_ practical possibility.
For example: are you able, using the contract's API to produce a test case which leads to this scenario?

(hm... btw in web2 we could at least add some kind of "WTF" log in this case, but in smart contracts we're just reverting a user transaction and have no way to learn from this)

dnkolegov: We return the specific errors (I introduced) that are returned only in this case.

maciejwitowski: But we have no way to monitor whether they were thrown without running a node on a subnet and having some error filters set up.

dnkolegov: Correct as for all other errors we return using revert.

maciejwitowski: @dnkolegov can you take a look at this question above?
> For example: are you able, using the contract's API to produce a test case which leads to this scenario?

maciejwitowski:
@dnkolegov I stated my concerns [here](https://github.com/consensus-shipyard/ipc/pull/670#discussion_r1483389228) about adding checks for a scenario we see no way of happening. Let's answer the test case question there (ideally by adding a test case) and close it PR so we land other fixes it includes as well.

maciejwitowski (approved):

contracts/src/gateway/GatewayManagerFacet.sol
dnkolegov: I can't do that. If it were possible, we would have a bug in the code. I am fine with removing those checks.

PR #671, opened 2024-02-06
ENG-578: IT (Part 2) - Testnet and abstract Materializer

aakoshh:
Builds on https://github.com/consensus-shipyard/ipc/pull/665 

Creates a `Materializer` trait and a `Testnet` struct to parse the manifest and create resources. 

The `Materializer` has abstract types for the things it creates, so that we can plug in different strategies. The first will be a simple validation that should check that e.g. validator balances can be satisfied and generally all dependencies are in order. 

In future PRs we can implement the `Materializer` in various ways:
* Create docker containers and return structs which stop them when they are dropped
* Parse a materializer log and instantiate nodes upon existing containers, managed externally to the test, and skip destroying them at the end
* Just perform a dry-run that prints CLI commands so we can have a feel for the flow of what would happen.

The `Testnet` itself assumes nothing about the `Node` and other types, nothing that it can actually _do_ with them besides looking them up and passing them back to the `Materializer` by reference. For example a `Node` type in practice could hold API endpoints for a particular remote machine, or port numbers for a local docker container, and it could implement platform specific methods to stop/start the container, to give an API client; but in a dry run it could be just a name that we can turn into cute messages like "Alice is starting a relayer following node-1 on subnet-2" or whatever.

It's not entirely clear yet how we are going to interact with these in a test, but imagine something like this:

```rust

fn my_test<M>(materializer: &mut M) -> anyhow::Result<()>
where 
  M: Materializer, 
  M::Node: HasEthapi, 
  M::Account: HasPrivateKey
{
  let manifest = load_manifest("mynet.yaml");
  Testnet::validate(manifest)?;
  let testnet = Testnet::setup(&mut materializer, "mynet", manifest)?;
  let node: M::Node = testnet.node(testnet.root().subnet("subnet-1").node("node-1"));
  let alice: M::Account = testnet.account("alice");
  let tx: Transaction = todo!();
  node.ethapi().send_raw_transation(tx.sign(alice.private_key()));
}

#[test]
fn my_test_docker() {
  let mut materializer = DockerMaterializer::new();
  my_test(&mut materializer).unwrap();
}
```


fendermint/testing/materializer/src/materializer.rs
```diff
@@ -0,0 +1,246 @@
...
+/// the help of the materializer, which should keep the [Testnet] itself testable.
+///
+/// The materializer might not actually instantiate the resources. By returning
+/// abstract types instead of concrete values, it is possible to just collect the
+/// operations and use them to validate the behaviour of whatever is driving
+/// the materializer. We can use this for dry-runs as well.
+///
+/// A live materializer should persist its logs, so that it can be resumed.
+/// For example we can create an run a testnet externally, then parse the manifest
+/// and the materializer logs inside a test to talk to one of the nodew, and the
```
fridrik01, line 29:
```suggestion
/// and the materializer logs inside a test to talk to one of the nodes, and the
```

fendermint/testing/materializer/src/materializer.rs
```diff
@@ -0,0 +1,246 @@
...
+use tendermint_rpc::Url;
+
+use fendermint_vm_genesis::Collateral;
+
+use crate::{
+    manifest::Balance, AccountName, NodeName, RelayerName, ResourceHash, SubnetName, TestnetName,
+};
+
+/// The materializer is a component to provision resources of a testnet, and
+/// to carry out subsequent commands on them, e.g. to restart nodes.
```
fridrik01, line 16: So basically the materializer implementation provides the concrete implementation of the backing resource (it being docker, local system, raw cli commands, etc) and the Testnet orchestrates these command to get to a valid testnet deployment.

fendermint/testing/materializer/src/materializer.rs
```diff
@@ -0,0 +1,246 @@
...
+/// and the materializer logs inside a test to talk to one of the nodew, and the
+/// materializer should be able to return to the test correct JSON-RPC endpoints.
+///
+/// Some of the operations of the materializer should be idempotent, e.g. the
+/// creation of a wallet or a node should only happen once.
+///
+/// The types returned might have their own logic to execute when dropped, to free
+/// resources. This might happen only if the resource is not an externally managed
+/// one, e.g. a testnet set up before tests are run, which the materializer should
+/// know.
```
fridrik01, line 38: So the resource returned might (if applicable) implement the Drop trait for RAII so we properly free the underlying resource (docker container, process, etc)

fendermint/testing/materializer/src/lib.rs
```diff
@@ -1,6 +1,257 @@
...
+}
+
+impl NodeName {
+    pub fn is_in_subnet(&self, subnet_name: &SubnetName) -> bool {
+        subnet_name.0.is_prefix_of(&self.0)
+    }
+}
+
+impl SubnetName {
+    pub fn subnet<T: Into<SubnetId>>(&self, id: T) -> Self {
```
fridrik01, line 138: nice, this implementation is pretty clever!

fridrik01 (approved):
this looks very good to me, appreciate the detailed comments and readable code! I had just some minor questions

fendermint/testing/materializer/src/lib.rs
```diff
@@ -1,6 +1,257 @@
...
+pub struct ResourceId(String);
+
+impl<'de> Deserialize<'de> for ResourceId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        String::deserialize(deserializer).map(Self::from)
+    }
+}
```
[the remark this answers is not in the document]
aakoshh: That's a great question. I'll should have added a comment because I forgot why I implemented it in the first place. 

I remember now: it's because the way it produces itself from a `String` is by replacing any `/` with `_` so I don't have to reject any name with errors, just silently turn it into one which is compatible with URLs.

fendermint/testing/materializer/src/materializer.rs
aakoshh: Yes, and example of that is in https://github.com/consensus-shipyard/ipc/pull/700/files#diff-e53ae3b1e30b37699016a7e43c4462b00ece2c8803212e253d6f3743f09abb96

fendermint/testing/materializer/src/materializer.rs
aakoshh: Yes, that's pretty much all it does at the moment.

fendermint/testing/materializer/src/lib.rs
aakoshh: I wanted to make it fluent as much as I could, without a lot of `.into()`, so something like `.subnet("foo")` should work in tests.

PR #676, opened 2024-02-08
Wire in the customized EAM actor into genesis

cryptoAtwill:
This one builds upon `eng-608` but performs the deployment of customized EAM actor. The approach is:
- Load the manifest as it is
- Created a `replace_built_in_actor` method that reads the state of system actor, replaces the target built in actor and updates the system actor manifest.


fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -122,6 +129,48 @@ where
...
             stage: Stage::Tree(state_tree),
+
+            builtin_actor_bundle: vec![],
+            custom_actor_bundle: vec![],
+        };
+
+        Ok(state)
+    }
+
+    pub async fn new_state(
```
fridrik01, line 140: Since we no longer parse the bundles here, consider renaming to something like `new_empty_state`

fendermint/vm/interpreter/src/fvm/genesis/builtin.rs
```diff
@@ -0,0 +1,246 @@
...
+        let custom_actor_manifest =
+            CustomActorManifest::load(&store, &custom_manifest_data_cid, custom_manifest_version)?;
+
+        let code = custom_actor_manifest
+            .code_by_name(fendermint_actor_eam::IPC_EAM_ACTOR_NAME)
+            .ok_or_else(|| anyhow!("ipc eam actor not found"))?;
+
+        for (_, code_cid) in builtin.iter_mut().filter(|(name, _)| name == "eam") {
+            *code_cid = *code
+        }
```
fridrik01, line 183: do we not have a constant for the name?, also we should assert that this name was actually found in the builtin vector

fendermint/vm/interpreter/src/fvm/genesis/builtin.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        Ok(addr_to_id)
+    }
+
+    async fn replace_eam_actor<DB>(
+        &self,
+        state: &mut FvmGenesisState<DB>,
+        store: &MemoryBlockstore,
+        builtin: &mut Vec<CodeByName>,
+    ) -> anyhow::Result<(Cid, Cid)>
```
fridrik01, line 168: We should rename this function or split it in two functions as its not clear what its doing by its name. We could then make it more general whereby the caller can send the actor name it wants to replace (and even a list in the future)

raulk (commented):
Let's consider a different approach here.

## Bundles

* Have each custom actor under `fendermint/actors` generate its own bundle. So instead of generating a monolithic custom actors bundle, have the `chainmetadata` build generate a one-actor bundle. Same for the EAM -- generates its own bundle.
* Rename `settings.builtin_actors_bundle` to `settings.system_actor_bundles`. Make it variadic so it can accept many paths. This should take the builtin actors and the chainmetadata actor.
* Add a new field `settings.extra_actor_bundles`, also a variadic setting, it can take many paths. This is the one where users would add the EAM actor if they want to (we don't ship it by default!).

## Blockstore load

1. We load all bundles into the blockstore, starting with system bundles, then moving to extra bundles. We retain each Manifest.
2. We merge all Manifests by key, where earlier manifests are overridden by entries in later manifests. You can probably use [Iterator#fold](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) here. This is how the built-in EAM would be overridden by the custom EAM.

## Actor instantiation

I would keep this as it was (in the original init method), and would instruct users to override whatever they wanted in-place. In the future, we can make this nicer by calling out to a `customize_genesis_state_tree` method that devs can implement to customize the base state tree, but we can follow up on that later.

cryptoAtwill:
I feel it's clearer if we are very specific about which built-in actor we want to replace. For example, if we want to replace EAM actor, we just have a parameter in genesis/config like:
```
built_in_actors:
  default: ...
  customize:
    eam: path to eam
```
I feel this way is better than `extra_actor_bundles`, because after parsing the `extra_actor_bundles`, we need to guess by name if the custom actor should actually replace the default built-in actor, unless we kind of assumed if the name overlaps, we replace the default one. But this is ambiguous as the code id could be different and they just share similar names.

fendermint/app/src/app.rs
```diff
@@ -446,11 +446,11 @@ where
             anyhow!("failed to load custom actor bundle CAR from {custom_actors_bundle:?}: {e}")
         })?;
 
-        let state = FvmGenesisState::new(
+        let state = FvmGenesisState::new_state(
```
aakoshh, line 449: Is this rename necessary? "state::new" says the same thing as "state::new_state" already.

cryptoAtwill: because `new` is taken and I dont want to update this method that will break all the tests. If this approach is ok, I will replace `new`.

aakoshh (commented):
Sorry, I would still prefer the flow I suggested on Slack. 

Instead of moving a large swathe of builtin-actor creation and then modifying parts of it to handle the EAM actor, but nothing else, I would do the following:
* Keep the actor creation where it is, go through and load the entire builtin bundle as it is. 
* Implement the `replace_builtin_actor` method that look in the custom bundle already parsed into the `GenesisState` and do the necessary change in the state tree
  * replace the specified actor ID with the new definition and 
  * load the manifest held by the system actor 
  * replace the CID in the manifest and persist the new version to the store 
  * overwrite the state of the system actor to point at the new manifest root
* Call the method once from the genesis interpreter to replace the built-in EAM with ours

This is much easier to extend to allow users to do their own replacements.

fendermint/vm/interpreter/src/fvm/genesis/builtin.rs
aakoshh: We added one in https://github.com/consensus-shipyard/ipc/pull/610/files#diff-a5d767cb02a5b7648a64451054b2e648413395275d693e7ea683dd9676d01115R129 , although not for this and it's not accessible, but we could move that type to a common place.

The FVM has them [here](https://github.com/filecoin-project/ref-fvm/blob/d3913e483b9234cadf661e5fd9e40632287782ea/fvm/src/machine/manifest.rs#L10-L15) but they aren't accessible.

cryptoAtwill:
It makes me miss Java Spring Boot sometime, we can just create annotation and annotate the class with `@ReplaceBuildInActor(...)`. It's quite difficult to do the same in rust.

cryptoAtwill:
@aakoshh It's actually a bit trickier to replace built-in actors. It needs to be done before `SystemActor` is created. I think the flow of this PR is correct but code arrangement can be changed. It has to be:
1. Load all bundles in memory
2. Replace the built-in actors with state initialization
3. Generate new cid 
4. Init system actor with cid above
5. Init everything else

raulk:
I defer to what @aakoshh suggests. However:
- The chainmetadata actor is compulsory, whilst the modified EAM is not, so I would suggest putting these in two different categories. The former is not a custom actor, it's a system actor.
- I tried to avoid having three tiers of actors: builtin (inherited), system (currently chainmetadata, but in the future could include native versions of IPC actors, and more), custom (whatever the user wants). That's why I suggested consolidating builtin and system into system. But I see the logistical difficulties that Akosh mentions about the "code id".
- I do not agree with wanting to reserve a whole category/tier for builtin-actors simply because they're externally provided (i.e. based on provenance). It's just circumstancial that we use that bundle in bulk today. In the near future I'd expect us to repack that bundle leaving out actors we don't use/wire (miner, storage power, etc.)

cryptoAtwill:
Why not create a ActorBuilder struct that takes the built-in bundle and system/custom actor bundle path. then we add a few methods:
```
fn use_built_in_actor(name);
fn replace_built_in_actor(build_in_actor_name, custom_actor_name);
fn use_custom_actor(name);

fn build() -> (build_in_manifest, custom_manifest)
```

raulk:
@cryptoAtwill I think it's best to not overengineer it at this stage. I agree that a new abstraction (e.g. `ActorBuilder`) may be in order down the line, but let's keep it simple for now? Let's implement what @aakoshh suggested above.

1. Keep the actor state constrution where it was.
2. Keep producing just one custom actors bundle, like we already are (ignoring my idea of producing 1 bundle per actor).
3. Implement the `replace_builtin_actor` as Akosh suggests.
4. In the genesis routine, add an example of how to use `replace_builtin_actor` to use our EAM actor, passing in allowlisted deployers, but keep it commented out.

cryptoAtwill:
@raulk The implementation is updated, but I cannot achieve @aakoshh 's method fully because the Manifest is completely loaded in genesis state. I have to do what’s done in the PR to get around it. If we are going to refactor it further, maybe we can store the built-in actor's name and code cid vector in genesis state instead of the Manifest.

aakoshh:
I don't understand why the system actor state has to be fully created up front. The actor state is just a CID in the `StateTree` under the actor's ID. We should be able to load the content from the blockstore, save a modified copy and update the CID in the `StateTree`.

I wouldn't go for this `ActorBuilder` simply because the `FvmGenesisState` already has these methods, and the interpreter is calling it. Building actors is basically what they do.

fendermint/vm/interpreter/src/fvm/genesis.rs
```diff
@@ -247,6 +235,24 @@ where
             )
             .context("failed to create chainmetadata actor")?;
 
+        let eam_state = fendermint_actor_eam::State::new(
+            state.store(),
+            // or we can allow certain address only with:
+            // fendermint_actor_eam::PermissionModeParams::AllowList(
+            //  vec![...]
+            // )
+            fendermint_actor_eam::PermissionModeParams::Unrestricted,
```
aakoshh, line 244: Is there a ticket number to track that this has to go in the genesis file?

fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -99,25 +100,35 @@ where
...
+        let mut intermediate_builtin = load_builtin(builtin_actor_bundle).await?;
+
+        replace_built_in_actor(
+            fendermint_actor_eam::IPC_EAM_ACTOR_NAME,
+            &custom_actor_manifest,
+            &mut intermediate_builtin,
+        )?;
+
+        // copy builtin actors from memory to store
+        intermediate_builtin.mem_store.copy_to(&store)?;
```
aakoshh, line 121: If you can do this here, I'm confident that you can also do this in the interpreter so that it can be offered to the users as well, and as I mentioned the actor name/CID can be kept together with the definition of its state, rather than be split in two places: that the name is replaced here, _and_ you have to follow up with the corresponding state in a completely different place, and if you don't then things go horribly wrong.

cryptoAtwill: I thought about doing what you mentioned, but the issue I found quite strange was the FvmGenesisState takes the manifest_cid and manifest as part of the struct. When we deploy the contracts, there is no access to the original builtin actor bundle . I understand where you are coming from and I think that’s the better way. But I feel we can refactor the FvmGenesisState to accept bundle_bytes in place of manifest_cid and manifest that way we can have everything in one place and it would feel natural.

aakoshh: I might be wrong as I haven't tried it myself, but I don't think taking manifest bytes or a loaded and parsed manifest makes much of a difference to what I'm saying. The manifests are loaded in `new` and loaded immediately, so that methods like `create_builtin_actor` and `create_custom_actor` can refer to the content by code IDs and names. 

I think you are pointing out this difference because we are still talking about a different thing. I suspect you want to keep it as bytes so you can manipulate those bytes before loading them in the store, as you do now IIRC. I am suggesting not to do that. 

Instead, I suggest that we load all bundles into manifests at the start, as we do now. The builtin actor bundle goes in the ref-fvm Manifest, the custom actors go in our custom manifest that indexes by name. Then we load the regular builtin actors as we do now, no change in that. After that, we implement the code that can `replace_builtin_actor` method: when it's called, the `System` actor will already exist, and its `builtin_actors` CID will match the one in the `FvmGenesisState::manifest_data_cid`. At that point we can ignore the `manifest_data_cid`; we load the `Vec<(String, Cid)>` pointed at the by the `System` actors state, find the name we want to replace, store the new vector, assign its CID to the actor state, store the actor state, modify the system actor's ID in the `StateTree` to point at the new state, and we can also store the values inside the `FvmGenesisState` as well if we load a new `Manifest`. 

This leaves some intermediate values in the blockstore but that's okay because they won't be persisted as nothing will point at them.

cryptoAtwill: @aakoshh I see where you are coming from, so now `manifest_data_cid` is just for initial loading and subsequent updates are directly read from the system actor state. This works for me as well!

fendermint/vm/interpreter/src/fvm/genesis.rs
cryptoAtwill: yep, #706

fendermint/vm/interpreter/src/fvm/genesis.rs
```diff
@@ -189,17 +189,6 @@ where
...
-            .create_builtin_actor(
-                eam::EAM_ACTOR_CODE_ID,
-                eam::EAM_ACTOR_ID,
-                &EMPTY_ARR,
-                TokenAmount::zero(),
-                None,
-            )
-            .context("failed to create EAM actor")?;
-
         // Burnt funds actor (it's just an account).
```
raulk, line 192: Is there any harm in keeping this here? The `replace` method should be able to override this.

fendermint/vm/interpreter/src/fvm/genesis.rs
raulk: @aakoshh My view here is that we should be extremely selective with what goes in genesis, as it could easily lead to configuration bloat as we keep adding more custom actors, or more knobs. Since our early users will be forking Fendermint, I'd rather have them use whichever mechanism they want to set the policy here.

fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -176,6 +176,74 @@ where
...
+        {
+            *code_cid = code
+        }
+
+        let builtin_actors = self.put_state(built_in_actors)?;
+        let new_cid = self.put_state(system::State { builtin_actors })?;
+        let mutate = |actor_state: &mut ActorState| {
+            actor_state.state = new_cid;
+            Ok(())
+        };
```
raulk, line 237: Where do we apply this mutation?

fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -464,4 +532,19 @@ where
             Stage::Exec(ref mut exec_state) => g(exec_state.state_tree_mut()),
         }
     }
+
+    /// Query the actor state from the state tree under the two different stages.
+    fn get_actor_state<T: de::DeserializeOwned>(&self, actor: ActorID) -> anyhow::Result<T> {
+        let actor_state_cid = match &self.stage {
+            Stage::Tree(s) => s.get_actor(actor)?,
+            Stage::Exec(s) => s.state_tree().get_actor(actor)?,
```
raulk, line 540: I'm not familiar with this stuff, so defer to @aakoshh's feedback for correctness here.

fendermint/vm/interpreter/src/fvm/genesis.rs
aakoshh: Yes, please keep it. If we open up replacements to users, we don't want to modify this part of the code, so we should walk that path here as well IMO.

cryptoAtwill: correct, but it should work without this as well.

fendermint/vm/interpreter/src/fvm/genesis.rs
aakoshh: My opinion is what can change should go into the config, rather than be hardcoded and force people to fork just to change a value. We obviously cannot leave it like this, it has to go into some config. Alternatively we should not do the replacement here at all, just farm out the ability to do so into the application root, to minimise the surface area of where users have to look if they fork or import the code. 

In general, things that affect the consensus should be in the `genesis.json` file, and other settings that can vary between nodes should live outside. This once clearly affects the state, so it's a thing for `genesis.json`. 

In the Comsos SDK for example, every module can have their own section in the genesis, it's just a JSON file after all, and which keys goes in there depends on the node configuration. Here's the [genesis for staking](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/staking/v1beta1/genesis.proto) for example. 

I agree that we don't want bloat. For example I didn't put in the ability to disable the chain metadata, even though it affects consensus, because I think it won't be used outside tests. However, this particular actor is useless without configuration, and everyone has to agree on who is allowed to deploy. In this case sharing the genesis is easiest if it's nicely wrapped up in a single file, rather than that file, plus some other config that people have to remember to also send to their peers.

That said, I see that the fact we grab configuration for the genesis from on-chain complicates things to some extent, ie. not sure if we want this to be a configuration of the subnet creation CLI and the contracts. 

Anyway, I just wanted to point out that it has to go _somewhere_.

fendermint/vm/interpreter/src/fvm/state/genesis.rs
aakoshh: Here: https://github.com/consensus-shipyard/ipc/pull/676/files#diff-8f0901d357789002b95cef8fd68d18055c0ef0b06a3873ee9cea6396f177ea3cR240-R241

raulk: Oops, my bad 😆

fendermint/vm/interpreter/src/fvm/state/genesis.rs
aakoshh: It looks correct, however I think it can be implemented on top of `with_state_tree` already:

```rust
let actor_state_cid = self.with_state_tree(
  |s| s.get_actor(actor),
  |s| s.get_actor(actor)
).ok_or_else(|| anyhow!("actor state {actor} not found, is it deployed?"))?
```

Doesn't make a huge difference in this case, but it's why that method exists.

aakoshh: In fact it's already used like that [here](https://github.com/consensus-shipyard/ipc/blob/90dfca6ac2bf9915777a015d15e0bbb1ff6fbdd5/fendermint/vm/interpreter/src/fvm/state/genesis.rs#L365)

aakoshh: @raulk this pattern is used because the `FvmGenesisState` starts with a `StateTree` but later evolves into having a full `FvmExecState` to actually run EVM contract deployments.

cryptoAtwill: that works for me too, when I looked at the function definition, I had the impression `with_state_tree` is mutating the state and it requires `&mut self`. THis way there is no need for `&mut`

fendermint/vm/interpreter/src/fvm/genesis.rs
raulk: @aakoshh the original intention was to offer this actor, but not to wire it in and ship it as a default actor in genesis (which is what this PR is doing). Keeping the code inchantations commented with some instructions is what I originally proposed here.

The rationale being that bloating genesis does not scale as we introduce additional IPC actors down the line: new FVM runtimes, new system services (tunnels, bridges, etc.), etc. We'll eventually move to a content-addressed manifest model that declares the contents of the subnet both at genesis, and at upgrade checkpoints.

But if you feel strongly about shipping with this modified actor as a default and exposing the config through genesis, I can live with that. More so because @cryptoAtwill had already done that and he dropped the change on my request.

raulk: @cryptoAtwill let's just call the shots here -- @cryptoAtwill can you re-add the genesis config for this actor and make sure that it can tackle all three cases (unrestricted, restricted with nobody allowed, restricted with specific addresses allowed).

aakoshh (approved):
Thank you, it looks correct :+1: 

My comments aren't blockers, just minor things I'd consider changing for the sake of consistency.

fendermint/vm/interpreter/src/fvm/state/genesis.rs
aakoshh: Yeah, I didn't want two versions, it was already verbose enough and I had access to `&mut` everywhere I needed this.

fendermint/vm/interpreter/src/fvm/genesis.rs
aakoshh: True, it works without it - the question is, does it work _with_ it?

contracts/binding/src/lib.rs
raulk, line 1: Revert fmt pls.

fendermint/app/src/cmd/genesis.rs
```diff
@@ -45,6 +45,7 @@ cmd! {
       power_scale: self.power_scale,
       validators: Vec::new(),
       accounts: Vec::new(),
+      eam_permission_mode: PermissionMode::Unrestricted,
```
raulk, line 48: How does the user specify the permission mode when constructing genesis?

cryptoAtwill: This is the default setup, so when a new genesis is created by calling fendermint, the default mode is `Unrestricted`. Then users can update this field to restricted.

contracts/binding/src/lib.rs
cryptoAtwill: reverted

raulk, line 1: There is still a diff here @cryptoAtwill. Can you revert all changes on this file?

raulk, line 1: Revert diff please.

raulk: I think your editor is messing with you!

PR #680, opened 2024-02-08
FIX: Ability to skip updating the state on empty blocks

aakoshh:
Adds a setting to disable saving the application state on empty blocks.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -102,6 +102,11 @@ pub struct AppConfig<S: KVStore> {
     pub state_hist_namespace: S::Namespace,
     /// Size of state history to keep; 0 means unlimited.
     pub state_hist_size: u64,
+    /// Whether to update the application state on empty blocks,
+    /// which will cause those blocks to be published even if
+    /// CometBFT would otherwise skip them, due to the change
+    /// in the application hash.
+    pub state_update_on_empty: bool,
     /// Path to the Wasm bundle.
     ///
     /// Only loaded once during genesis; later comes from the [`StateTree`].
@@ -160,6 +165,8 @@ where
     ///
     /// Zero means unlimited.
     state_hist_size: u64,
+    /// Whether to update the application state on empty blocks.
+    state_update_on_empty: bool,
 }
 
 impl<DB, SS, S, I> App<DB, SS, S, I>
@@ -189,6 +196,7 @@ where
             namespace: config.app_namespace,
             state_hist: KVCollection::new(config.state_hist_namespace),
             state_hist_size: config.state_hist_size,
+            state_update_on_empty: config.state_update_on_empty,
             interpreter: Arc::new(interpreter),
             chain_env,
             snapshots,
@@ -675,6 +683,7 @@ where
             tendermint::Hash::Sha256(h) => h,
             tendermint::Hash::None => return Err(anyhow!("empty block hash").into()),
         };
+        let is_empty = request.header.data_hash.is_none();
 
         let db = self.state_store_clone();
         let state = self.committed_state()?;
@@ -692,7 +701,7 @@ where
 
         let state = FvmExecState::new(db, self.multi_engine.as_ref(), block_height, state_params)
             .context("error creating new state")?
-            .with_block_hash(block_hash);
+            .with_block_info(block_hash, is_empty);
 
         tracing::debug!("initialized exec state");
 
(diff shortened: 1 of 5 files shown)


fendermint/app/src/app.rs
```diff
@@ -759,42 +759,60 @@ where
...
+        state.block_height = block_height;
+
+        // If there has been no update to the state then we can skip updating the the timestamp,
+        // because it would be the only thing changing the application state, which would cause
+        // CometBFT to produce an empty block even if it's configured to skip them.
+        // Leaving the timestamp unchanged only affects historical queries targetting these heights.
+        // The state timestamp will catch up with the header timestamp as soon as there is a block
+        // which does something to the ledger. This could be an empty block causing a checkpoint.
+        // See https://docs.cometbft.com/v0.37/core/configuration#empty-blocks-vs-no-empty-blocks
+        let update_params = is_dirty || state_root != state.state_params.state_root;
```
sanderpick, line 789: I played with this and it seems that in every call to `commit`, `state_root != state.state_params.state_root`, even if there are no txns. 

I can confirm that if you force `update_params` to `false`, comet stops calling `commit` when `--consensus.create_empty_blocks=false` is used.

aakoshh: Thanks for trying! 

Ah, I remember now, that's because we have this new step where the block hash is [inserted into the ledger](https://github.com/consensus-shipyard/ipc/blob/c6c5142ddbbcd8b24bbcffc5bbd79faa4aee89c0/fendermint/vm/interpreter/src/fvm/exec.rs#L85-L113) so that we can implement the `BLOCKHASH` function in the `evm` actor.

Hm, I can't see how we could get around that causing a new empty block from being created :thinking: 

The only way would be if we didn't save the hash of empty blocks, but I don't think breaking the 1-to-1 correspondence of the API and the ledger would be a good idea.

aakoshh: I added a setting to disable chain metadata, in case you can live without it in your tests.

sanderpick: Ah! Makes sense. The setting looks great. Thanks for the quick help on this :pray:

raulk:
@aakoshh Basic question: is there no difference in the pattern of ABCI calls emitted by CometBFT for us to detect this edge case, and solve it in a less intrusive manner? There seem to be three scenarios here:

1. Transactions were included. Block is emitted unconditionally. Must commit new blockhash into chainmetadata actor.
2. Transactions were not included. CMT_CONSENSUS_CREATE_EMPTY_BLOCKS=false. No block emitted.
3. Transactions were not included. CMT_CONSENSUS_CREATE_EMPTY_BLOCKS=true. Block emitted regardless.

I'm thinking maybe we're hooking into the ABCI flow at an incorrect point? If we hooked in elsewhere, could we detect if a block is **actually** being produced?

aakoshh:
@raulk thanks for asking, I also misunderstood what was going on first. 

There is another scenario, namely that if we change the app hash (ie. if we write anything to the block store), then an empty block _will_ be produced by CometBFT, because that is the vehicle to broadcast the commit over that hash.

So this:
> Transactions were not included. CMT_CONSENSUS_CREATE_EMPTY_BLOCKS=false. No block emitted. Avoid committing to the chainmetadata actor in all cases.

Is not true, if we did something, and empty block will come even if we asked CometBFT not to produce one. And that block exists, it's not a tentative block in the sense that if we don't change the app hash, the block won't be created. It has already happened, CometBFT is just telling us about it, so we have to save its hash if we want to remember what it was at that height.

aakoshh:
Of course we could do it completely differently by not storing the state of the chain metadata in the actor store, but rather somewhere else, deliberately keeping it out of the application hash, however we'd have to cater for snapshotting and garbage collection in a different way then.

maciejwitowski:
@sanderpick For visibility: we discussed in the team that merging this PR would be a bad precedent of adding a workaround instead of fixing the root of the problem. Realistically speaking, we would never revisit it in the future.
This PR has been marked "On Hold" and we will keep it open. Feel free include these changes in the Fendermint fork you maintain, to help you with getting through all the logs while testing.

aakoshh:
Some comments for reference:
* CometBFT decides whether to create a proof block [here](https://github.com/cometbft/cometbft/blob/cfe8b888a1e7e357da39b17a9201c77a2d5071f1/internal/consensus/state.go#L1094-L1097), there's no way to disable this. 
* It would be a huge refactor to start storing the timestamp and the chain metadata in a way that it doesn't affect the application hash. Removing it from that umbrealla would also mean that when CometBFT offers a snapshot, the step where we [validate the state params](https://github.com/consensus-shipyard/ipc/blob/7bb9c69134489b20c866808ee8b68729926e3ad3/fendermint/app/src/tmconv.rs#L411-L419) in the snapshot against the app hash validated by the light client would not cover some of the state, opening up an attack vector of malicious snapshot content that restores off-band state with invalid values that subsequently affect contract execution, leading to potential consensu failure.

sanderpick:
@maciejwitowski @aakoshh makes sense. apologies for the distraction!

PR #681, opened 2024-02-09
[ipc-cli] Add ipc wallet list command

mb1896:
https://linear.app/interplanetary-consensus/issue/ENG-611/ipc-wallet-list-command

Example usage:
```
$ ../target/release/ipc-cli wallet list --wallet-type evm
Address: 0xb4f69b8abc5b3a7f9b085661500b898a29fabe78     PubKey: 045dad983ebf44763387eb8ad65c1f32ea95fc6ac59425862365e13a68e699d4de2f4c4ffb866da7cd3ed56b5a8f7bbcc580107d316d5dc70eefbb33f7f78fdf06
Address: 0xefbd91dcdfd39a7aa2cb6566538ac57c8fe16093     PubKey: 04bdecde198fffcaf8dfc4e98a31ad07b5e653d961eedaba702505c46d749125fe5818a568af1e54de825b581d62734a9adf806e9c32bbbd8ac075af27bce27775
Address: 0xf161ba5e71ea64b34e95a860f3b093425c7ca9af     PubKey: 040e37bd1810d1519b2cf2d8993fe24872f81a81d068252d512cf9be1d0c9918526602423ef53a1523d3ef00658aeeeb1bcdb93942df0696b87b7becf437e1eee9

$ ../target/release/ipc-cli wallet list --wallet-type fvm
Address: t1giqacc3q4ielfoasz5ypllcfh7f2u6yidgujeba      PubKey: 04a2035d5a5e61dbb5ea313c2b5eda3cc473cda236e7fe3841c4f0d989a18e6384af36a49ec35af0b9178719ebbfb5599aa1984efb926d8f3789758003ef835f8d     KeyType: Secp256k1
```

Closes ENG-611

### DIFF
--- a/ipc/cli/src/commands/wallet/list.rs
+++ b/ipc/cli/src/commands/wallet/list.rs
@@ -1,0 +1,42 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: MIT
+//! Wallet list cli handler
+
+use async_trait::async_trait;
+use clap::Args;
+use ipc_wallet::{EthKeyAddress, EvmKeyStore};
+use std::fmt::Debug;
+
+use crate::{get_ipc_provider, CommandLineHandler, GlobalArguments};
+
+pub(crate) struct WalletList;
+
+#[async_trait]
+impl CommandLineHandler for WalletList {
+    type Arguments = WalletListArgs;
+
+    async fn handle(global: &GlobalArguments, _: &Self::Arguments) -> anyhow::Result<()> {
+        let provider = get_ipc_provider(global)?;
+        let wallet = provider.evm_wallet()?;
+        let addresses = wallet.read().unwrap().list()?;
+        for address in addresses.iter() {
+            if *address == EthKeyAddress::default() {
+                continue;
+            }
+
+            print!("Address: {}", address.to_string());
+
+            let key_info = wallet.read().unwrap().get(address)?.unwrap();
+            let sk = libsecp256k1::SecretKey::parse_slice(key_info.private_key())?;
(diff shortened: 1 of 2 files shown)


raulk (commented):
Excellent! Let's query all wallets and display the key type on every row.

ipc/cli/src/commands/wallet/list.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub(crate) struct WalletList;
+
+#[async_trait]
+impl CommandLineHandler for WalletList {
+    type Arguments = WalletListArgs;
+
+    async fn handle(global: &GlobalArguments, _: &Self::Arguments) -> anyhow::Result<()> {
+        let provider = get_ipc_provider(global)?;
+        let wallet = provider.evm_wallet()?;
```
fridrik01, line 20: Its not clear to me that this command only support evm wallet types, could we let it handle fvm wallet type as well?

aakoshh (approved):
I am not familar with the `ipc-cli`, but if it works it looks fine to me. 

I said a couple of times before that I thought it would be good to establish some conventions about how to print stuff from the CLI. The `fendermint` commands print JSON, which can be parsed with `jq` so that it can be handled programmatically. I understand the CLI is a human tool, so maybe a `--format` argument would be helpful. 

Since there is no such convention, your format is as good as any :+1:

ipc/cli/src/commands/wallet/list.rs
mb1896: Thanks. Added support for FVM now.

PR #683, opened 2024-02-09
ENG-578: IT (Part 3) - Validation

aakoshh:
Builds on https://github.com/consensus-shipyard/ipc/pull/671

Adds a `ValidatingMaterializer` and a `LoggingMaterializer`. This required quite a bit of massaging of the Rust lifetimes, and I also had to separate out the generic types into their own trait, so that I can wrap one materializer into another. 

Validation is tested with the arbitrary manifests, so we verify that the arbitrary manifests make sense and also that the testnet issues sensible operations.


fendermint/testing/materializer/src/validation.rs
```diff
@@ -0,0 +1,382 @@
...
+    // /// ```text
+    // /// RUST_LOG=info cargo test -p fendermint_testing_materializer prop_validation -- --nocapture
+    // /// ```
+    // fn init_log() {
+    //     let _ = env_logger::builder().is_test(true).try_init();
+    // }
+
+    /// Check that the random manifests we generate would pass validation.
+    #[quickcheck_async::tokio]
+    async fn prop_validation(id: TestnetId, manifest: Manifest) -> anyhow::Result<()> {
```
fridrik01, line 379: Curious why we need a tokio executor here?

fendermint/testing/materializer/src/validation.rs
```diff
@@ -0,0 +1,382 @@
...
+
+        let pbs = self.balances.get_mut(&parent).unwrap();
+        let pb = pbs.get_mut(account).unwrap();
+
+        if *pb < amount {
+            bail!("{account:?} has less than {amount} on {parent:?}, cannot fund {subnet:?}");
+        }
+        *pb = pb.clone().sub(amount.clone());
+
+        if credit_child {
```
fridrik01, line 149: where will the funds go if this is false?

fendermint/testing/materializer/golden/manifest/json/manifest.json
```diff
@@ -1,99 +1,97 @@
 {
   "accounts": {
-    "AJxB": {},
-    "Aj0I_t": {},
-    "TeBcA": {},
-    "ccXUc3": {}
+    "4rENv3UF": {},
```
fridrik01, line 3: What is the plan with these golden manifest files, should we not pick one of json/toml? I find the json one much easier to read and understand than the others. Also, if we commit this in the repo, what is the rule for when it changes, I guess its seeded somewher so it should change but didn't find that in the code

fendermint/testing/materializer/src/validation.rs
```diff
@@ -0,0 +1,382 @@
...
+        balance: Balance,
+        reference: Option<ResourceHash>,
+    ) -> anyhow::Result<()>
+    where
+        's: 'a,
+    {
+        // Debit parent balance, but do not make the funds available in the child
+        self.fund_from_parent(subnet, account, collateral.0, false)?;
+        // Debit parent balance; Credit child balance
+        self.fund_from_parent(subnet, account, balance.0, true)?;
```
fridrik01, line 319: Doesn't this need to be atomic? Like if first one succeeds and second fails

fendermint/testing/materializer/src/validation.rs
```diff
@@ -0,0 +1,382 @@
...
+        account: &'a VAccount,
+        subnet: &'a VSubnet,
+        amount: TokenAmount,
+        reference: Option<ResourceHash>,
+    ) -> anyhow::Result<()>
+    where
+        's: 'a,
+    {
+        // Debit parent balance; Credit child balance
+        self.fund_from_parent(subnet, account, amount, true)?;
```
fridrik01, line 299: I'm probably confusing terminology but are we not moving funds from parent to child? So aren't we crediting the parent and debiting the child?

fendermint/testing/materializer/src/materializer.rs
```diff
@@ -12,6 +12,31 @@ use crate::{
     manifest::Balance, AccountName, NodeName, RelayerName, ResourceHash, SubnetName, TestnetName,
 };
 
+/// Type family of all the things a [Materializer] can create.
+///
+/// Kept separate from the [Materializer] so that we can wrap one in another
+/// and pass the same types along.
```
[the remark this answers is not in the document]
aakoshh: Haha, wicked, yeah, I haven't thought about doing local and remote at the same time, and I'm also not sure exactly if the types would check out if you wrap two different kinds of materializers at the same time, but it should be possible!

fendermint/testing/materializer/src/logging.rs
```diff
@@ -0,0 +1,238 @@
...
+        self.inner.create_account(account_name)
+    }
+
+    async fn fund_from_faucet<'s, 'a>(
+        &'s mut self,
+        account: &'a M::Account,
+        reference: Option<ResourceHash>,
+    ) -> anyhow::Result<()>
+    where
+        's: 'a,
```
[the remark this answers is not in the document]
aakoshh: Yeah, courtesy of using `&mut` with `async`. I found these in https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/second-edition/ch19-02-advanced-lifetimes.html but the TL;DR is that this says "self outlives the parameter", so they can survive when suspended.

fendermint/testing/materializer/src/testnet.rs
```diff
@@ -373,45 +380,51 @@ where
...
+        m: &mut R,
         parent_subnet_name: &SubnetName,
         subnet_id: &SubnetId,
-        subnet: Subnet,
+        subnet: &Subnet,
     ) -> anyhow::Result<()> {
         let subnet_name = parent_subnet_name.subnet(subnet_id);
 
-        // Pre-fund the accounts, create the subnet, start the nodes.
+        // Create the subnet
```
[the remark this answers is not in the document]
aakoshh: Actually they are required by the compiler in this case. When I call `let parent_submit_config = self.submit_config(parent_subnet_name)?;` it contains references to `&self` inside the `SubnetConfig` (to `Node` and `Deployment` records). At the end of the block here we have `self.subnets.insert(subnet_name.clone(), created_subnet);` which requires `&mut self` again, and the compiler is not happy that we have a variable that has a shared reference to `&self` via the `SubmitConfig` and now we want to modify it too. But if we put it in a block, it is free to drop the submit config, and then there's only the mutable reference. 

This is why in the next block the same `submit_config` has to be called again, to recreate that variable after it's been dropped. So basically the blocks are building up to something being stored in `self`, and then a new block does the next thing.

fendermint/testing/materializer/src/validation.rs
aakoshh: Because the methods are asynchronous. In some other tests like the `smt_` ones I created a runtime manually, but I thought creating 100 one might be a bit of an overkill and a distraction in this test. Luckily this crate exists to make it available in quickcheck.

fendermint/testing/materializer/src/validation.rs
aakoshh: They don't go anywhere in this validator. In reality they get locked up in the `SubnetActor` on the parent. For example currently when you call `join` you say how many tokens you want to put down as collateral, and the `SubnetActor` holds on to it until you are ready to `leave`, or if (in the future) you get slashed. 

Disclaimer I think locking collateral like this isn't great, and I proposed https://github.com/consensus-shipyard/ipc-actors/issues/27 to do it differently.

fendermint/testing/materializer/golden/manifest/json/manifest.json
aakoshh: I'd pick YAML, defnitely not TOML, potentially JSON. Not sure we have to pick one, or let users pick their favourite and just pick the right deserializer based on the extension. We can remove the golden files for formats that nobody would pick.

Unfortunately it's not seeded in a repeatable way (quickcheck doesn't support seeding :( ), so if you delete it will be a new random one. It just serves as 1) illustration and 2) early warning for breaking changes.

fendermint/testing/materializer/src/validation.rs
aakoshh: This is just validation, so I don't think it matters much. Originally I thought I'll just collect errors in a list and return all the problems at the end, rather than shortcut at the first, then you fix it and it dies on the second. I thought the [Validated](https://typelevel.org/cats/datatypes/validated.html) monad worked like that, or maybe it was [this older one](https://eed3si9n.com/learning-scalaz/Validation.html). 

Then I thought I'm not going to bother, the testnet can fail early on other problems too, and with the logs it's easier to see the steps end right where the error happened. 

So, this is really just a way to say "hold on, something's not right HERE" and let you figure it out, rather than a proper accounting system. 

On a "real" materializer, the `reference` should help to make things resumable, ie. if the first succeeded, the second failed, and you run the whole thing again, it should skip the first and only try the second.

fendermint/testing/materializer/src/validation.rs
aakoshh: I might have been using it wrong. The +/- depends on which side of the balance sheet you are on: https://accountingplay.com/debits-and-credits/

If I'm holding money in the bank, I'm a liability to them, so credit increases my balance, debit decreases. 
If I took at loan from the bank, I'm an asset to them, so credit decreases and debit increases. 

I thought these are more like the "I keep money in the subnet", the subnet owes me money the same way a bank owes it for my deposits, so I thought when I get tokens on a subnet, they are "credited to me".

fendermint/testing/materializer/src/testnet.rs
```diff
@@ -373,45 +380,51 @@ where
...
             // Assume that all subnets are deployed with the default contracts.
             self.deployments
-                .insert(subnet_name.clone(), m.default_deployment(&subnet_name));
+                .insert(subnet_name.clone(), m.default_deployment(&subnet_name)?);
 
             // Where can we reach the gateway and the registry.
             let parent_submit_config = self.submit_config(parent_subnet_name)?;
 
             // Create the subnet on the parent.
-            m.create_subnet(
```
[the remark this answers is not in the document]
fridrik01: like wasnt there an ignored return value?

fendermint/testing/materializer/golden/manifest/json/manifest.json
raulk: Would consider options like [HCL](https://github.com/hashicorp/hcl) (which formally translates to JSON), and [JSON5](https://json5.org/), which are good tradeoffs between human and machine readable.

fendermint/testing/materializer/src/validation.rs
fridrik01: makes sense!

fendermint/testing/materializer/src/testnet.rs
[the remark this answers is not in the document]
aakoshh: Oh right; there is not `#[must_use]` annotation on it, so I don't think it would care if the value is dropped, but the test must have brought it out.

fendermint/testing/materializer/src/validation.rs
fridrik01: got it!

fendermint/testing/materializer/src/validation.rs
fridrik01: ok wasn't sure if that comment was outdated or not but that makes sense from that perspective

fridrik01 (approved):
lgtm!

PR #684, opened 2024-02-10
[ipc-cli] Use string for argument --supply-source-kind and --permission-mode

mb1896:
https://linear.app/interplanetary-consensus/issue/ENG-613/ipc-cli-avoid-using-int-enum-values-in-arguments

Closes ENG-613

### DIFF
--- a/ipc/api/src/subnet.rs
+++ b/ipc/api/src/subnet.rs
@@ -78,6 +78,19 @@ impl TryFrom<u8> for PermissionMode {
     }
 }
 
+impl TryFrom<&str> for PermissionMode {
+    type Error = anyhow::Error;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        Ok(match s {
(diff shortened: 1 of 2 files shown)


maciejwitowski (approved):

maciejwitowski:
@mb1896 LGTM. I also added this "Closes ENG-613" for linear to automatically close the issue when merged. Let's use it by default with every PR when there's 1:1 mapping between Linear issue and PR. Thanks!

raulk (approved):
This LGTM, although it could be more idiomatic by implementing [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html).

We can also remove the TryFrom<u8> trait implementations.

I'll submit a follow up PR to use the [`strum::EnumString`](https://docs.rs/strum/latest/strum/derive.EnumString.html) macros.

But we should look into using clap-rs Enums

Issue #685, opened 2024-02-12
Address slow integration tests on CI


fridrik01:
I think we should address the slow integration/e2e tests on CI, its creeping over 40+ minutes. 

The issue with this is not that the actual tests are so slow, but the whole compile/build steps are. 

Some ideas:
- Try caching all the rust dependencies (we aren't upgrading them so often), the contracts, etc.
- Commit contract artifacts to git repo, then we don't need to compile them (this takes many minutes on CI)
- Consider whether downloading prebuilt docker container (which we built each time PR gets merged to main) would speed things up for us. 
- more ideas?

nahsi:
> Commit contract artifacts to git repo
You can also cache them like [this](https://github.com/fluencelabs/deal/blob/main/.github/workflows/artifacts.yml#L61)

aakoshh:
Caching used to work, but I don't see the big caches updated any more at https://github.com/consensus-shipyard/ipc/actions/caches, they are months old :thinking: 

* For Rust, it's set up [here](https://github.com/consensus- shipyard/ipc/blob/aad693d88b26a5845f3fdab69a3e0406f9c41ddc/.github/workflows/fendermint-test.yaml#L78-L85) and [here](https://github.com/consensus-shipyard/ipc/blob/aad693d88b26a5845f3fdab69a3e0406f9c41ddc/.github/workflows/fendermint-publish.yaml#L37-L46)
* The contract compilation output is cached [here](https://github.com/consensus-shipyard/ipc/blob/aad693d88b26a5845f3fdab69a3e0406f9c41ddc/.github/workflows/fendermint-test.yaml#L87-L91) 

There is a mechanism to cache compiled dependencies in docker build [here](https://github.com/consensus-shipyard/ipc/blob/aad693d88b26a5845f3fdab69a3e0406f9c41ddc/fendermint/docker/builder.ci.Dockerfile#L15-L73), however I'm afraid because the latest docker image isn't pulled in the beginning we don't get the benefit of the cached layers. This should be an easy fix.

aakoshh:
Perhaps I misunderstood how [updating](https://github.com/consensus-shipyard/ipc/blob/aad693d88b26a5845f3fdab69a3e0406f9c41ddc/.github/actions/setup-cache/action.yaml#L23) works. 

We have this cache:
```
[...]

PR #687, opened 2024-02-12
ENG-719: Pull the latest image before building docker

aakoshh:
Addresses https://github.com/consensus-shipyard/ipc/issues/685

Pull the latest docker image on CI, which might help with the layer based caching. When I added that caching I forgot that new runners on github won't have the earlier docker registry artifacts at hand.


fridrik01 (approved):
LGTM!

PR #691, opened 2024-02-12
ipc-cli: use strum for PermissionMode and SupplySourceKind enums.

raulk:
Follow-up to https://github.com/consensus-shipyard/ipc/pull/684.

This is a bit less pedantic. Ideally we would use clap's [`ValueEnum`](https://docs.rs/clap/latest/clap/trait.ValueEnum.html) going forward, but it's tricky because these types are shared between the cli and the api packages.

### DIFF
--- a/ipc/cli/src/commands/subnet/create.rs
+++ b/ipc/cli/src/commands/subnet/create.rs
@@ -2,15 +2,16 @@
 // SPDX-License-Identifier: MIT
 //! Create subnet cli command handler.
 
-use anyhow::Context;
+use std::fmt::Debug;
+use std::str::FromStr;
+
 use async_trait::async_trait;
 use clap::Args;
 use fvm_shared::address::Address;
 use fvm_shared::clock::ChainEpoch;
+
 use ipc_api::subnet::{PermissionMode, SupplyKind, SupplySource};
 use ipc_api::subnet_id::SubnetID;
-use std::fmt::Debug;
-use std::str::FromStr;
 
 use crate::commands::get_ipc_provider;
 use crate::{f64_to_token_amount, require_fil_addr_from_str, CommandLineHandler, GlobalArguments};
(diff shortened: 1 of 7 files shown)


ipc/cli/src/commands/subnet/create.rs
```diff
@@ -112,11 +118,15 @@ pub struct CreateSubnetArgs {
         long,
         help = "The permission mode for the subnet: collateral, federated and static"
     )]
+    // TODO figure out a way to use a newtype + ValueEnum, or reference PermissionMode::VARIANTS to
+    //  enumerate all variants
     pub permission_mode: String,
```
aakoshh, line 123: It would be nice to use a custom parser function like [here](https://github.com/consensus-shipyard/ipc/blob/f53dfbeaadc52aec6284214083c6936f33f95d2a/fendermint/app/options/src/genesis.rs#L58-L62) instead of allowing a `String` and failing later.

aakoshh: Maybe a default value as well?

ipc/cli/src/commands/subnet/create.rs
```diff
@@ -112,11 +118,15 @@ pub struct CreateSubnetArgs {
         long,
         help = "The permission mode for the subnet: collateral, federated and static"
     )]
+    // TODO figure out a way to use a newtype + ValueEnum, or reference PermissionMode::VARIANTS to
```
aakoshh, line 121: I'm not familiar with `strum`, but there's only a few methods that [ValueEnum](https://docs.rs/clap/4.5.0/clap/trait.ValueEnum.html) needs to implement. Maybe the [derivations](https://github.com/clap-rs/clap/blob/master/clap_derive/src/derives/value_enum.rs) give some clues.

cryptoAtwill (approved):

ipc/cli/src/commands/subnet/create.rs
raulk: I'd advise against setting a default value because these are foundational, immutable configuration properties for the subnet, and it'd be wise for the user to set them explicitly.

raulk: > It would be nice to use a custom parser function like [here](https://github.com/consensus-shipyard/ipc/blob/f53dfbeaadc52aec6284214083c6936f33f95d2a/fendermint/app/options/src/genesis.rs#L58-L62) instead of allowing a `String` and failing later.

clap's `ValueEnum` would be the right choice, but these types sit in the `api` crate which should not know about CLI tooling. Hence my comment about a newtype above.

raulk: [`e817bb3` (#691)](https://github.com/consensus-shipyard/ipc/pull/691/commits/e817bb3dfc8166dd925814b5dc14de4e2bf7ce9a)

We now use `value_parser` for these fields, but we'll need to look at newtype + ValueEnum another time.

mb1896 (approved):

PR #695, opened 2024-02-12
Fix npm

dnkolegov:
This PR fixes NPM conflict.

### DIFF
--- a/contracts/binding/src/lib.rs
+++ b/contracts/binding/src/lib.rs
@@ -2,31 +2,25 @@
 #[macro_use]
 mod convert;
 #[allow(clippy::all)]
-pub mod checkpointing_facet;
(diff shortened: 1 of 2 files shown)


contracts/binding/src/lib.rs
```diff
@@ -2,31 +2,25 @@
 #[macro_use]
```
fridrik01, line 2: is this change needed? I thought we forced this to be formatted in https://github.com/consensus-shipyard/ipc/pull/659

dnkolegov:
> is this change needed? I thought we forced this to be formatted in https://github.com/consensus-shipyard/ipc/pull/659

@fridrik01 
This is generated during CI; git on my machine shows no changes in that file.

PR #699, opened 2024-02-13
ENG-719: Tag and push the fendermint builder image

aakoshh:
Related to [#719](https://github.com/consensus-shipyard/ipc/issues/685)


aakoshh:
Based on my local testing using `PROFILE=ci make docker-build` I don't think this works :( 

The dependencies seem to be built every time, even when there is no change in anything.

aakoshh:
Or I can see it working locally but on CI it failed to even export the builder image, it's 10GB locally.

fridrik01:
Thanks! There are some CI errors though

aakoshh:
Yeah I don't understand the errors. The `fendermint-publish.yaml` also has a step to clean up the disk, but now it looks like since I added that to make room for the docker image it can't work. 

I stopped looking at this for now because I need to make some headway with my other task.

maciejwitowski:
This is still a problem but needs more investigation.
We'll keep the Linear issue open to get back to this.

PR #700, opened 2024-02-13
ENG-578: IT (Part 4) - Docker 

aakoshh:
Builds on https://github.com/consensus-shipyard/ipc/pull/683

TODO: 
- [x] Docker cleanup runtime
- [x] Implement `create_network` - `docker create network`
- [x] Implement `create_account` 
- [ ] Implement `fund_from_faucet` - `curl`
- [ ] Implement `new_deployment` - `deploy.ts`
- [x] Implement `existing_deployment`
- [x] Implement `default_deployment`
- [x] Implement `create_root_genesis`
- [x] Implement `create_node` - `docker create container`
- [ ] Implement `start_node` - `docker start`
- [ ] Implement `create_subnet` - `ipc-cli`
- [ ] Implement `fund_subnet` - `ipc-cli`
- [ ] Implement `join_subnet` - `ipc-cli`
- [ ] Implement `create_subnet_genesis` - `fendermint` CLI
- [ ] Implement `create_relayer` - TBD
- [ ] Start testnet with Fendermint as rootnet
- [ ] Test restarting a node - `docker restart`
- [ ] Test querying the Ethereum API of a node


fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+    materials::export_file,
+    NodeName, ResourceHash,
+};
+
+// TODO: Add these to the materializer.
+const COMETBFT_IMAGE: &str = "cometbft/cometbft:v0.37.x";
+const FENDERMINT_IMAGE: &str = "fendermint:latest";
+
+const STATIC_ENV: &str = "static.env";
+const DYNAMIC_ENV: &str = "dynamic.env";
```
fridrik01, line 35: I am slightly confused by the difference between `STATIC_ENV` and  `DYNAMIC_ENV`. Why do we need a dynamic.env file instead of just passing them at runtime, is it a docker thing? Also, didn't find any cases of env DYNAMIC_ENV variables, is that expected?

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+    };
+}
+
+/// A Node consists of multiple docker containers.
+pub struct DockerNode {
+    /// Logical name of the node in the subnet hierarchy.
+    node_name: NodeName,
+    fendermint: DockerContainer,
+    cometbft: DockerContainer,
+    ethapi: Option<DockerContainer>,
```
fridrik01, line 64: Love it!

fendermint/testing/materializer/src/docker/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+pub use node::DockerNode;
+pub use relayer::DockerRelayer;
+
+const STATE_JSON_FILE_NAME: &str = "materializer-state.json";
+
+const DOCKER_ENTRY_SCRIPT: &str = include_str!("../../scripts/docker-entry.sh");
+const DOCKER_ENTRY_FILE_NAME: &str = "docker-entry.sh";
+
+const PORT_RANGE_START: u32 = 30000;
+const PORT_RANGE_SIZE: u32 = 100;
```
fridrik01, line 51: Do we bind these ports to the host?

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+    port_range: DockerPortRange,
+    path: PathBuf,
```
fridrik01, line 66: is this the path on the host machine which is shared as volume with the docker contains?

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+                    (cometbft_dir.clone(), "/cometbft"),
+                    (node_config.genesis.path.clone(), "/fendermint/genesis.json"),
+                ],
+            );
+
+            // Init cometbft to establish the network key.
+            cometbft_runner
+                .run_cmd("init")
+                .await
+                .context("cannot init cometbft")?;
```
fridrik01, line 134: in case any of these fail, should we delete the `cometbft_dir` so it will not skip this step on resume?

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+
+        // Wrap an entry point with the docker entry script.
+        let entrypoint = |ep: &str| {
+            format!(
+                "{} '{ep}' {} {}",
+                *DOCKER_ENTRY_PATH, *STATIC_ENV_PATH, *DYNAMIC_ENV_PATH
+            )
+        };
+
+        // Create a fendermint container mounting:
```
fridrik01, line 293:
```suggestion
        // Create a fendermint container
```

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+
+    fn docker(&self) -> &Docker {
+        &self.dh.docker
+    }
+
+    // Tag containers with resource names.
+    fn labels(&self) -> HashMap<String, String> {
+        [
+            ("testnet", self.node_name.testnet().path()),
+            ("node", self.node_name.path()),
```
fridrik01, line 433: would it make sense to also have a `container_name` label?

fridrik01 (commented):
This is looking really great, almost like reading our quickstart guide :)

fendermint/testing/materializer/src/docker/node.rs
aakoshh: Thanks for pointing this out, I'll leave a comment about it here. I mentioned it somewhere else but on their own they are confusing. The dynamic env only comes into play in https://github.com/consensus-shipyard/ipc/pull/744

So the idea was that we have `Materializer::create_node` and `Materializer::start_node` separately, and the only reason for this is to support the use case where we have node A bootstrapping from node B, and at the same time node B bootstrapping from A. This looks like a contradiction, but often the P2P layer allows things to kick off and wait for the other node to appear. This makes sense to me as I can operate 2 seeds: start the 1st, then the 2nd connecting to it, then if I need to restart the 1st, it connects to the 2nd. I think CometBFT persists previous connections, so maybe allowing this is an overkill. 

But there is another use case: in our current scripts, we create a fully connected network (every node to every other), and this requires knowing not just the address of other peers, but their peer IDs (the hash of their network keys). 

In my model the `create_node` is responsible for setting up everything, which includes the network key. If it started the node as well, then the nodes would have to form a DAG, only connecting to already created and started nodes, because that's when you can already look up the network ID. Instead, we first create everything, then start everything, and in that second pass, everyone's IDs are already known, so anyone can connect to anybody else as they please. 

And so, `static.env` is what can be put into file during creation, while `dynamic.env` is what is only apparent when the node starts. I could have just appended more values to a file, but thought it's easier if we restart nodes to just write it to a different file, which can thus be overwritten.

fendermint/testing/materializer/src/docker/node.rs
aakoshh: I'll add a comment; yes, it's the path on the host machine to all the files and directories for the containers. Not all are volumes, but most are. I added an example to the description of https://github.com/consensus-shipyard/ipc/pull/744

fendermint/testing/materializer/src/docker/node.rs
aakoshh: Yeah, that's a good point! I'll do it in the next PR, had to fix many things with this.

fendermint/testing/materializer/src/docker/node.rs
aakoshh: Depends on what exactly that is. You mean something to tell which of the three it is, like "I want to find all ethapi containers"? You don't mean the unique name produced by the `container_name()` function, right? Because that will be the actual name, so there is no reason to add it as metadata.

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -0,0 +1,566 @@
...
+            .docker()
+            .create_container::<&str, _>(None, config)
+            .await
+            .context("failed to create container")?
+            .id;
+
+        self.docker()
+            .start_container::<&str>(&id, None)
+            .await
+            .context("failed to start container")?;
```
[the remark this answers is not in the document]
aakoshh: In my experience in the next PR, it can fail if we misconfigure docker somehow, but having the failed container stick around means we can still inspect it to see what was wrong with it. It's easy to remove them with `docker system prune`, but I added a step to the tests to remove all previous containers labelled with the testnet.

fendermint/testing/materializer/src/docker/network.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+            1 => {
+                let id = networks[0]
+                    .id
+                    .clone()
+                    .ok_or_else(|| anyhow!("docker network {network_name} has no id"))?;
+
+                (id, true)
+            }
+            n => bail!("there are multiple docker networks with the same name: {network_name}"),
```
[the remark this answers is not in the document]
aakoshh: I don't think so. Highly defensive; I guess I can simplify it to just take the `.next()` of the iterator and that's it.

fendermint/testing/materializer/src/docker/mod.rs
aakoshh: Yes, that's how the test will be able to interact with the container. Accidentally it means all tests will share the same materializer state, so they nodes across different tests don't get the same ports.

fendermint/testing/materializer/src/docker/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+/// Allocated (inclusive) range we can use to expose containers' ports on the host.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DockerPortRange {
+    pub from: u32,
+    pub to: u32,
+}
+
+/// Mapping ports assuming a 100 size ranges.
+impl DockerPortRange {
+    /// Mapping 26655 to the host.
```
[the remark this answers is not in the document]
aakoshh: Yes that's right. I meant the internal port, in case someone recognises them, to make it easier to understand, but I'll add better comments.

fendermint/testing/materializer/src/docker/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+    /// this materializer in a test and allocate more if needed without clashes.
+    fn port_range(&mut self, node_name: &NodeName) -> anyhow::Result<DockerPortRange> {
+        if let Some(range) = self.state.port_ranges.get(node_name) {
+            return Ok(range.clone());
+        }
+        // Currently the range allocations are not dropped from the materializer,
+        // so the length can be used to derive the next available port. Otherwise
+        // we could loop through to find an unused slot.
+        let node_count = self.state.port_ranges.len();
+        let from = PORT_RANGE_START * node_count as u32;
```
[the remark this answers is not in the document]
aakoshh: Eagle eyes! Thanks, this was fixed in https://github.com/consensus-shipyard/ipc/pull/744

fendermint/testing/materializer/src/docker/container.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl Drop for DockerContainer {
+    fn drop(&mut self) {
+        if !self.container.external {
+            let container_name = self.container.name.clone();
+            let docker = self.dh.docker.clone();
+            self.dh.drop_handle.spawn(async move {
+                if let Err(e) = docker
+                    .stop_container(&container_name, Some(StopContainerOptions { t: 5 }))
```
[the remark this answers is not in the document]
aakoshh: I think so, the docs say:
```
/// Number of seconds to wait before killing the container
```

I'll add a variable name as `t` is not very telling.

fendermint/testing/materializer/src/docker/node.rs
aakoshh: Fixed in https://github.com/consensus-shipyard/ipc/pull/744/commits/e3858ff082e0017ca43be6c8322e32b2b4fd26c4

fridrik01 (approved):

PR #703, opened 2024-02-13
Change argument --subnet into --network for some of the ipc-cli subnet commands

mb1896:
Closes ENG-639

Please the discussion in linear ticket. Not sure what other subnet commands should adopt this change.

### DIFF
--- a/ipc/cli/src/commands/subnet/list_subnets.rs
+++ b/ipc/cli/src/commands/subnet/list_subnets.rs
@@ -21,7 +21,7 @@ impl CommandLineHandler for ListSubnets {
         log::debug!("list subnets with args: {:?}", arguments);
 
         let provider = get_ipc_provider(global)?;
-        let subnet = SubnetID::from_str(&arguments.subnet)?;
+        let subnet = SubnetID::from_str(&arguments.network)?;
 
         let gateway_addr = match &arguments.gateway_address {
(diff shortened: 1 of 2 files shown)


aakoshh (commented):
I have always been confused by the `network`. I thought it can be `testnet` or `mainnet`, corresponding to the FVM `Network` type which is part of every `Address`. That's how it's used in the [fendermint CLI](https://github.com/consensus-shipyard/ipc/blob/316c97f69326631640f4ef7ba1945845c9036d27/fendermint/app/options/src/lib.rs#L44-L46)

I was hoping that all remaining occurrence of `network` in the contracts can be replaced by `subnet` actually. When it's a `SubnetID`, why not call a spade a spade?

raulk:
@aakoshh there are 3 CLI commands being affected here:
- list subnets: more appropriate to call this flag `--parent`.
- RPC and chain ID query: these work for any network including the root, so `--subnet` is confusing and `--network` is more accurate IMO.

> I thought it can be testnet or mainnet, corresponding to the FVM Network type which is part of every Address. 

That should've really been called "network type" :-/

mb1896:
@raulk Updated argument names as you mentioned. Does it look good to you now?

maciejwitowski (approved):

maciejwitowski:
LGTM, merging not to block on @raulk  confirmation.

PR #704, opened 2024-02-14
Workflow to auto deploy contracts when changed

mb1896:
Sample output file: https://github.com/consensus-shipyard/ipc/commit/7d65801134fbf9369e20c342b835acb265f0f275
Saved in `cd/contracts` branch https://github.com/consensus-shipyard/ipc/commits/cd/contracts/

Closes ENG-627


.github/workflows/auto-deploy-contracts.yaml
```diff
@@ -0,0 +1,42 @@
...
+concurrency:
+  # Only allow one run at a time for this workflow
+  group: auto-deploy-contracts
+  cancel-in-progress: true
+
+jobs:
+  deploy-contracts:
+    runs-on: ubuntu-latest
+    env:
+      RPC_URL: https://calibration.filfox.info/rpc/v1
```
raulk, line 22: Not necessary, but can we move the URL to a secret, and make sure that the subnet deployment automation uses the same URL? Trying to reduce entropy and consolidate to a single configuration spot, in case endpoints become unstable and we need to change them.

raulk (changes requested):
We'll need to publish the addresses somewhere accessible. If we stop here, I understand we can do the following:

1. Open the GitHub Actions UI.
2. Find the last job where this workflow ran.
3. Scroll down the list to the last step.
4. Expand the logs.
5. Find the two addresses we care about in the firehose of logs
6. Copy each address.

However, that is wasteful for us devs and very poor UX for adopters.

Instead, let's parse the two addresses we care about, construct a .json file, and commit that .json file elsewhere public, associated with the commit. We can then use [shields.io's JSON support](https://shields.io/badges/dynamic-json-badge) to display those addresses in the repo's README.

Here's the cleanest and most scalable way of doing this, in a way that does not muddy the commit history of `main`:

1. Create a long lived `cd/contracts` branch (standing for: continuous deployment > contracts).
2. When contracts change, have the workflow merge the `main` branch into `cd/contracts` with a merge commit.
3. Execute the deployment, and extract the gateway and registry addresses.
4. Generate a `deployments/r314159.json` file with the two addresses + master commit deployed.

```json
{
    "commit": "<sha1>",
    "gateway_addr": "<eth addr>",
    "registry_addr": "<eth addr>",
}
```

5. [Push a commit](https://github.com/stefanzweifel/git-auto-commit-action) from within the action, containing the merge and the updated deployments metadata.

This solution allows us to hit all checkboxes:

- Tracks exactly which contracts commit was deployed (thanks to the merge preciding the metadata update commit).
- Keeps an easiliy accessible history of contract deployments (in case someone wants to test with an early version of contracts, e.g. when bisecting a bug).
- Addresses are always viewable by users, in a predictable URL.
- Latest addresses are machine-processable, always in a predictable URL.
- Repo stays lean in size.

.github/workflows/auto-deploy-contracts.yaml
fridrik01: Note that github secrets do not propagate between github workflows unless you explicitly forward them with (`secrets: inherit`) which is what I had to do in this [PR](https://github.com/consensus-shipyard/ipc/pull/674/files)

maciejwitowski:
Just a note that we discussed it in sync, and aligned that we need the addresses to be available _somewhere_.
We haven't 100% decided where exactly as @mb1896 had reservations about the branch suggested by @raulk .
@mb1896 feel free to propose alternatives. Let's aim to reach the decision today since it'd be valuable to have it next week when the hackathon starts.

raulk:
For the record, the alternative that @mb1896 suggested was placing the addresses in a file in `main`. This was not seen as a good solution because bots pushing to `main` causes all sorts of problems. The parallel branch with merges from main is cleaner. And it's easy to change if we feel it doesn't work for us. So let's just go do it.

raulk:
@mb1896 this is the GitHub Action you should use: https://github.com/stefanzweifel/git-auto-commit-action It does not require a Personal Access Token.

mb1896:
> Just a note that we discussed it in sync, and aligned that we need the addresses to be available _somewhere_. We haven't 100% decided where exactly as @mb1896 had reservations about the branch suggested by @raulk . @mb1896 feel free to propose alternatives. Let's aim to reach the decision today since it'd be valuable to have it next week when the hackathon starts.

Thanks for the comments. I'm following @raulk 's direction now. Will sure to make this ready soon for external developers to check out.

maciejwitowski:
@mb1896 just cross-posting from Slack:
I think it would help reviewers if:
- the commits had descriptive names, not "test" ("test" is fine for local development but when you want someone to review, it's better to squash the commits and made them clearer)
- there was a TODO included in the description to indicate what is left to code so reviewers know what to expect

.github/workflows/auto-deploy-contracts.yaml
mb1896: I don't think the URL is a secret per se. Besides, it's often that a RPC URL fails or take too long to response. In that case I would just make a PR to switch it to a different provider. But if we keep track them in secrets, we probably already forgotten what URL we were using. That just make things more implicit for maintainers.

.github/workflows/auto-deploy-contracts.yaml
```diff
@@ -0,0 +1,85 @@
...
+          deployed_registry_address=$(echo "$deploy_output" | grep '"SubnetRegistry"' | awk -F'"' '{print $4}')
+          echo "gateway_address=$deployed_gateway_address" >> $GITHUB_ENV
+          echo "registry_address=$deployed_registry_address" >> $GITHUB_ENV
+          echo "commit_hash=$(git rev-parse HEAD)" >> $GITHUB_ENV
+      - name: Review deployed addresses
+        run: |
+          echo "commit_hash: $commit_hash"
+          echo "gateway_address: $gateway_address"
+          echo "registry_address: $registry_address"
+      - name: Switch code repo to cd/contract branch
```
raulk, line 64: @mb1896 we're missing the merge from `main` into the cd branch, prior to committing the file. Can you please add it? This will (a) place the source and the addresses in one convenient branch, (b) create a nice merge graph tracking at the git history level what got deployed when.

raulk (changes requested):

.github/workflows/auto-deploy-contracts.yaml
mb1896: Fixed. Please take another look.

maciejwitowski (approved):
LGTM, thanks!

mb1896:
@raulk Can you take another look?

.github/workflows/auto-deploy-contracts.yaml
```diff
@@ -0,0 +1,114 @@
...
+      RPC_URL: https://calibration.filfox.info/rpc/v1
+      PRIVATE_KEY: ${{ secrets.CONTRACTS_DEPLOYER_PRIVATE_KEY }}
+    steps:
+      - name: Checkout cd/contract branch
+        uses: actions/checkout@v4
+        with:
+          ref: cd/contracts
+          submodules: recursive
+
+      - name: (Dry run) Try merge from main branch and update cd/contracts branch
```
mb1896, line 31: Dry run pull merge is to verify that the condition of cd/contracts branch is able to merge with latest main branch. If failed, we will resolve whatever error manually.

fridrik01: Calling this a dry run is at bit confusing as its actually pulling and merging (so has side effects on the runner)... its just not pushing changes. 

Also "... and update cd/contracts branch" does not seem to be right

fridrik01 (approved):
had some minor comment, but otherwise this looks good to me

maciejwitowski:
@mb1896 When you're online, let's address Fridrik's comments, merge `main` and merge this PR. Raul's comments have been addressed so no need to block this PR on them. Thanks!

maciejwitowski:
Btw this is great, I was already able to unblock a dev who encountered error because of contracts incompatibility by sending addresses from [this commit](https://github.com/consensus-shipyard/ipc/commit/07778692998ce1b627bd910b52dd3bc2ca375261) to them..

.github/workflows/auto-deploy-contracts.yaml
mb1896: Yes, the "dry" part mostly means that we will not modify remote code repos. This is useful because we can fail early when merging generate conflicts (rather than complain after deploying the contract).

I updated the step's name to better describe what it does.

raulk (approved):

PR #711, opened 2024-02-15
Example contracts for Axelar-bridged tokens, straight into IPC subnets.

raulk:


dnkolegov (approved):

examples/axelar-token/src/IpcTokenSender.sol
```diff
@@ -0,0 +1,53 @@
...
+
+        // Perform some sanity checks.
+        require(token.balanceOf(msg.sender) >= amount, "insufficient token balance");
+        require(token.allowance(msg.sender, address(this)) >= amount, "insufficient token allowance");
+
+        // Lock the value under custody, and authorize the Axelar ITS to spend it on our behalf.
+        token.transferFrom(msg.sender, address(this), amount);
+        token.approve(address(_axelarIts), amount);
+
+        // Tell the IpcTokenHandler on the IPC L1 rootnet to credit these funds to the specified beneficiariy
```
maciejwitowski, line 41:
```suggestion
        // Tell the IpcTokenHandler on the IPC L1 rootnet to credit these funds to the specified beneficiary
```

maciejwitowski (approved):
Clever idea 👏

extras/axelar-token/src/IpcTokenHandler.sol
```diff
@@ -0,0 +1,115 @@
...
+        if (available == 0) {
+            revert NothingToWithdraw();
+        }
+
+        delete _claims[msg.sender][token];
+        IERC20(token).safeTransfer(msg.sender, available);
+    }
+
+    // @notice Queries the claim of a beneficiary over a particular token.
+    function claimFor(address beneficiary, address token) external view returns (uint256) {
```
snissn, line 112: claimFor as a name implies a change is there a name like getClaimFor that we can use to make it more clearly a view?

extras/axelar-token/src/IpcTokenSender.sol
```diff
@@ -0,0 +1,53 @@
...
+        require(tokenAddress != address(0), "could not resolve token address");
+
+        IERC20 token = IERC20(tokenAddress);
+
+        // Perform some sanity checks.
+        require(token.balanceOf(msg.sender) >= amount, "insufficient token balance");
+        require(token.allowance(msg.sender, address(this)) >= amount, "insufficient token allowance");
+
+        // Lock the value under custody, and authorize the Axelar ITS to spend it on our behalf.
+        token.transferFrom(msg.sender, address(this), amount);
```
snissn, line 38: Do we have error handling here if the Axelar call fails - do we need to track deposits and manage receipts or is that not needed with Axelar?

extras/axelar-token/test/TestHandler.sol
```diff
@@ -0,0 +1,43 @@
...
+        vm.startPrank(axelarIts);
+
+        vm.mockCall(
+            address(ipcGateway),
+            abi.encodeWithSelector(TokenFundedGateway.fundWithToken.selector, subnet, recipient.from(), 1),
+            abi.encode("")
+        );
+        handler.executeWithInterchainToken(bytes32(""), "", "", params, bytes32(""), address(token), 1);
+    }
+
```
snissn, line 40: It would be good to include more test coverage, could we simply expand on the todos for testing such that we can have clear next steps for us organized?

snissn: Having TODOS that detail future coverage of _claims feels right and that book keeping that we are creating on top of the third party calls that may be difficult to test

snissn (approved):
Approving with some minor requests for clarification in the documentation and requesting a few more details on future tests in a TODO somewhere

extras/axelar-token/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+> Cross L1 chain atomic deposit of ERC20 tokens into token-supply subnets, via the Axelar network.
+
+**⚠️ Disclaimer: These contracts and implementation herein have not been security-reviewed or audited, and are considered bleeding-edge in terms of development.
+IPC developers do not make any guarantees.
+USE AT YOUR OWN RISK.**
+
+## Motivation
+
+Imagine an ERC20 token residing on the Ethereum L1.
```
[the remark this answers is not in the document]
raulk: Sure, good feedback!

extras/axelar-token/Makefile
```diff
@@ -0,0 +1,49 @@
+include .env
+export
+
+.PHONY: deps
+deps:
```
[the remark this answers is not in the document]
raulk: Interesting. The intention is you'd pull and initialize recursive submodules, then you wouldn't need to run install. In fact, I think this concrete `forge install` may be the wrong thing to run here since we probably want to lock dependency versions.

extras/axelar-token/src/IpcTokenSender.sol
raulk: This is running on the foreign L1. But indeed, the Axelar call can fail in which case it would return the funds back to this contract. I'll check with the Axelar team what's their suggested solution. Good flag!

extras/axelar-token/remappings.txt
```diff
@@ -0,0 +1,7 @@
+@axelar-network/interchain-token-service/=lib/interchain-token-service/contracts/
+@axelar-network/axelar-gmp-sdk-solidity/=lib/axelar-gmp-sdk-solidity/
+ds-test/=lib/forge-std/lib/ds-test/src/
+forge-std/=lib/forge-std/src/
+openzeppelin-contracts/=lib/openzeppelin-contracts/contracts/
+fevmate/=../../contracts/lib/fevmate/contracts/
+@ipc/=../../contracts/
```
snissn, line 7: This is great :) :) I will be able to use this @ipc remapping in order to have the cross-token usdc code in the separate PR also in the extras folder

extras/axelar-token/src/IpcTokenHandler.sol
raulk: Good point! The convention for getters is to not prefix them with `get`, but agree that it becomes ambiguous in this case as `claim` can be confused for a verb/action. Will change.

extras/axelar-token/test/TestHandler.sol
raulk: Added more TODOs and opened a Linear issue to track: https://linear.app/interplanetary-consensus/issue/ENG-741/enhance-test-coverage-for-axelar-token.

extras/axelar-token/Makefile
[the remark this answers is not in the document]
raulk: I think I've fixed this in [`1723cdb` (#711)](https://github.com/consensus-shipyard/ipc/pull/711/commits/1723cdbbb7af7298fe64e0f5950594d45661500d).

PR #717, opened 2024-02-16
Add an Upgrade Scheduler

fridrik01:
This PR adds an `UpgradeScheduler` which enables users to code predefined upgrades at certain height. The UpgradeScheduler manages a list of `Upgrades` where each upgrade defines the `chain_name` it should be applied to, the `height` it should be executed on and a `MigrationFunc` which users need to implement to carry out the migration.

`Upgrade` is defined as follows:

```rust

pub type MigrationFunc<DB> = fn(state: &mut FvmExecState<DB>) -> anyhow::Result<()>;

pub struct Upgrade<DB>
where
    DB: Blockstore + 'static + Clone,
{
    /// the chain name on which the upgrade should be executed
    chain_name: String,
    /// the chain id is calculated from the chain_name
    chain_id: ChainID,
    /// the block height at which the upgrade should be executed
    block_height: BlockHeight,
    /// the application version after the upgrade (or None if not affected)
    new_app_version: Option<u64>,
    /// the migration function to be executed
    migration: MigrationFunc<DB>,
}
```

At runtime, the `FvmMessageInterpreter` checks after processing each block, whether the `UpgradeScheduler` contains an upgrade at the current height. If so, it will execute the `migration` function. 

When the migration function finishes, it will either:
- Succeed, and returns a new application version (or 0 if not affected) which will be saved into the state
- Fails, which results in the interpreter returning an error to the abci's `end_block` callback.

The PR comes with a small testing framework `Tester` which simulates block processing and allows us to run Upgrades at different heights.


fendermint/app/src/app.rs
```diff
@@ -180,6 +184,9 @@ where
         chain_env: ChainEnv,
         snapshots: Option<SnapshotClient>,
     ) -> Result<Self> {
+        let upgrade_scheduler = load_upgrade_scheduler();
+        upgrade_scheduler.validate()?;
```
aakoshh, line 188: Shouldn't it already return a validated scheduler, or otherwise make sure we can't forget to call it?

fendermint/app/src/app.rs
```diff
@@ -752,6 +760,18 @@ where
             .await
             .context("end failed")?;
 
+        // check for upgrades in the upgrade_scheduler
+        if let Some(upgrade) = self.upgrade_scheduler.get_upgrade(request.height) {
+            // there is an upgrade scheduled for this height, lets run the migration
+            let res = (upgrade.migration)(request.height);
```
aakoshh, line 766: I am surprised that the migration doesn't get access to the `exec_state`. How else is it going to change the actor stuff. There are pending changes in there as well, so if it did something out of bounds, it would be overwritten in the commit, which sets the new state root.

fendermint/vm/upgrade_scheduler/src/upgrade_scheduler.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// the migration function to be executed
+    pub migration: MigrationFunc,
+}
+
+/// UpgradeScheduler represents a list of upgrades to be executed at given heights
+/// During each block height we check if there is an upgrade scheduled at that
+/// height, and if so the migration for that upgrade is performed.
+#[derive(Clone)]
+pub struct UpgradeScheduler {
+    pub upgrades: BTreeMap<i64, Upgrade>,
```
aakoshh, line 25: It seems redundant to index the upgrades by heights and also have a `height` property.

fendermint/vm/upgrade_scheduler/src/upgrade_scheduler.rs
```diff
@@ -0,0 +1,91 @@
...
+/// a function type for migration
+pub type MigrationFunc = fn(i64) -> i64;
+
+/// Upgrade represents a single upgrade to be executed at a given height
+#[derive(Clone)]
+pub struct Upgrade {
+    /// the block height at which the upgrade should be executed
+    pub height: i64,
+    /// the migration function to be executed
+    pub migration: MigrationFunc,
```
aakoshh, line 17: Slightly curious that the upgrade knows when it's supposed to be applied, but the function takes the height as an argument too. It sounds like it should already know, as it's part of the upgrade itself?

fendermint/vm/upgrade_scheduler/src/upgrade_scheduler.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn new() -> Self {
+        Self {
+            upgrades: BTreeMap::new(),
+        }
+    }
+
+    // add a new upgrade to the schedule
+    pub fn add_upgrade(&mut self, upgrade: Upgrade) -> Result<()> {
+        if upgrade.height <= 0 {
+            return Err(anyhow::anyhow!("Upgrade height must be greater than 0"));
```
aakoshh, line 39: Then maybe we can use the `BlockHeight` alias which is `u64`.

fendermint/app/src/upgrades.rs
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use fendermint_vm_upgrade_scheduler::upgrade_scheduler::UpgradeScheduler;
+
+// initialize default upgrade scheduler
+pub fn load_upgrade_scheduler() -> UpgradeScheduler {
+    let upgrade_scheduler = UpgradeScheduler::new();
+
+    // Add your upgrades to the upgrade_scheduler here
+    // ...
```
aakoshh, line 11: Just a note that `add_upgrade` returns `Result`

aakoshh (commented):
Nice, defining upgrades in code is probably the only way to manipulate state. It can potentially execute wasm code too. 

You need to pass it [FvmExecState](https://github.com/consensus-shipyard/ipc/blob/main/fendermint/vm/interpreter/src/fvm/state/exec.rs).

fendermint/app/src/app.rs
fridrik01: @aakoshh I know that the migration function signature is now what it should be, its just taking and receiving ints for the purpose of get things to compile and see fendermint calling a migration function at specified epoch

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+        self.state_params.circ_supply = circ_supply;
+
+        Ok(())
+    }
+}
+
+#[test]
+fn testest() {
+    let rt: tokio::runtime::Runtime =
+        tokio::runtime::Runtime::new().expect("create tokio runtime for init");
```
aakoshh, line 220:
```suggestion
#[tokio::test]
asyn fn testest() {
```
I only created runtimes manually because I was running with the `arbtest` framework; or `quickcheck`, but then discovered that crate which supports `tokio` with it.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+        migration: |state| {
+            println!(
+                "Migration at height {} just prints out chainmetadata actor",
+                state.block_height()
+            );
+
+            let actor = state
+                .state_tree_mut()
+                .get_actor(system::SYSTEM_ACTOR_ID)
+                .unwrap();
```
aakoshh, line 238:
```suggestion
                .get_actor(system::SYSTEM_ACTOR_ID)?;
```

It looks like this closure should be able to handle errors. I think this test might be one that any prospective upgrade author would look at, so we should try to be as close as possible to how we'd like migrations to be written.

It will result in a panic but at least we'd know where it happens: in the ABCI service wrapper.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+fn testest() {
+    let rt: tokio::runtime::Runtime =
+        tokio::runtime::Runtime::new().expect("create tokio runtime for init");
+
+    let mut us = UpgradeScheduler::default();
+
+    let chain_id = 1942764459484029;
+
+    us.add(Upgrade {
+        chain_id,
```
aakoshh, line 227: Why do we need the chain ID? It will be part of the state when it executes. Is this some kind of a failsafe device, that if you aren't running the chain you expect, it will not upgrade in an unintended height?

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+            println!(
+                "Migration at height {} confirms the chainmetadata actor is deleted",
+                state.block_height()
+            );
+
+            let actor = state
+                .state_tree_mut()
+                .get_actor(system::SYSTEM_ACTOR_ID)
+                .unwrap();
+            print!("chainmetadata actor after delete: {:?}", actor);
```
aakoshh, line 274: Wrong ID. Maybe it could print `actor.is_none()`, it would be a bit more suggestive of what it expects.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+            Ok(())
+        },
+    })
+    .unwrap();
+
+    let interpreter: FvmMessageInterpreter<MemoryBlockstore, HttpClient> =
+        FvmMessageInterpreter::new(
+            tendermint_rpc::HttpClient::new("http://127.0.0.1:26657").unwrap(),
```
aakoshh, line 282: Can you use the `MockClient` like I do in other tests? It would be more obvious that we don't expect CometBFT to run. Which we don't, right?

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+            None,
+            contracts_path(),
+            1.05,
+            1.05,
+            false,
+            Some(us),
+        );
+
+    let mut tester = Tester::new(interpreter, MemoryBlockstore::new());
+    rt.block_on(tester.init_genesis()).unwrap();
```
aakoshh, line 292:
```suggestion
   tester.init_genesis().await.unwrap();
```

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+#[derive(Clone)]
+struct Tester<I> {
+    interpreter: Arc<I>,
+    state_store: Arc<MemoryBlockstore>,
+    multi_engine: Arc<MultiEngine>,
+    exec_state: Arc<tokio::sync::Mutex<Option<FvmExecState<MemoryBlockstore>>>>,
+    state_params: FvmStateParams,
+}
+
+impl<I> Tester<I>
```
aakoshh, line 43: Great job extracting this from the `App`. For a moment I was thinking it could be a nice refactor, to disentangle the state management from all the other stuff that goes on there. It's not clear how much can be done, as the method parameters are slightly different there, even though the overall shape is the same. In any case it's a great reference, and I think you could move this tester to live under `testing/contract-test/src/lib.rs`, as an interpreter driver for FVM messages. It allows block boundaries, which hasn't been done so far, but would be useful in other tests I'm sure.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+            self.state_store_clone(),
+            self.multi_engine.clone(),
+            &bundle,
+            &custom_actors_bundle,
+        )
+        .await
+        .context("failed to create genesis state")?;
+
+        let genesis = Genesis {
+            chain_name: "test".to_string(),
```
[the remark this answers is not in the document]
aakoshh: Also, maybe this genesis should be an input to `init_chain`, so tests can customise it?

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,300 @@
...
+            timestamp: Timestamp(0),
+            network_version: NetworkVersion::V21,
+            base_fee: TokenAmount::zero(),
+            power_scale: 0,
+            validators: Vec::new(),
+            accounts: Vec::new(),
+            eam_permission_mode: PermissionMode::Unrestricted,
+            ipc: None,
```
aakoshh, line 112: Note that if this is `None` then IPC actors won't be deployed. That is fine, and another reason to pass in genesis from the test, so this `Tester` can be as general as possible, with no hidden behaviour.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,6 +212,33 @@ where
...
+        match &self.upgrade_scheduler {
+            Some(upgrade_scheduler) => {
+                let chain_id: u64 = state.chain_id().into();
+                let block_height: u64 = state.block_height().try_into().unwrap();
+
+                if let Some(upgrade) = upgrade_scheduler.get(chain_id, block_height) {
+                    tracing::info!(
+                        chain_id = chain_id,
+                        height = block_height,
+                        "Running migration",
```
aakoshh, line 225: Maybe a migration title would be informative in the future. Or a more formal application version. 

That is actually part of what is returned by the `info` method in the `app.rs`. There are `version` and `app_version` there, both hardcoded at the moment:
```rust
    /// The application software semantic version.
    pub version: String,
    /// The application protocol version.
    pub app_version: u64,
```

I thought `version` is only about the `fendermint` binary, and `app_version` is what we are migrating here. It could be interesting to give some structure to `app_version` with shifts, e.g. an IPC version and a Fluence version could be packed in it with bit shifting, although I'm not sure if CometBFT expects some ordering on it. Or we can just surrender control over it to these upgrades, after all it would be the operator who applies them, they have to opt into IPC upgrades as much as their own. 

In any case, it would be something to think about: to return not just `PowerUpdates` but a potential new app version as well from the interpreter. Then it would also be clear in which method the upgrades are applied, if it was part of the `EndOutput`, and then `App` would know to save it somewhere.

Looking at the `FvmStateParams::network_version`, I reckon `app_version` belongs there too, and it's not clear what the difference between the two is exactly. `App::info` does have access to it at least, and I think if you ask what the app version was in a previous height, it should report the one before the migration, so it's part of the state.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,6 +212,33 @@ where
...
+                    );
+
+                    // there is an upgrade scheduled for this height, lets run the migration
+                    let res = (upgrade.migration)(&mut state);
```
aakoshh, line 229: I'm not sure what exactly `res` is here, but if it can fail, it should stop the execution.

fendermint/vm/interpreter/src/fvm/upgrade_scheduler.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use fvm_ipld_blockstore::Blockstore;
+
+use super::state::{snapshot::BlockHeight, FvmExecState};
+
+/// a function type for migration
+// TODO: Add missing parameters
+pub type MigrationFunc<DB> = fn(state: &mut FvmExecState<DB>) -> anyhow::Result<()>;
+
+pub type ChainID = u64;
```
aakoshh, line 14: There is a [type for it](https://github.com/consensus-shipyard/ipc/blob/1a72b58a94b8d7efb78a7560e3e5d5a6101269ae/fendermint/vm/core/src/chainid.rs#L7)

fendermint/vm/interpreter/src/fvm/exec.rs
aakoshh: If you add a field for it, this might just be the time to start worrying about backwards compatibility. I should have added a golden file for `FvmStateParams` CBOR, but I doubt that I did :(

fendermint/vm/interpreter/src/fvm/upgrade_scheduler.rs
```diff
@@ -0,0 +1,110 @@
...
+where
+    DB: Blockstore + 'static + Clone,
+{
+    // add a new upgrade to the schedule
+    pub fn add(&mut self, upgrade: Upgrade<DB>) -> anyhow::Result<()> {
+        if self
+            .upgrades
+            .contains_key(&(upgrade.chain_id, upgrade.block_height))
+        {
+            return Err(anyhow::anyhow!(
```
aakoshh, line 62:
```suggestion
            bail!(
```

fendermint/vm/interpreter/src/fvm/upgrade_scheduler.rs
```diff
@@ -0,0 +1,110 @@
...
+    let upgrade = Upgrade {
+        chain_id: 1,
+        block_height: 20,
+        migration: |_state| Ok(()),
+    };
+    let res = us.add(upgrade);
+    assert!(res.is_err());
+
+    assert!(us.get(1, 9).is_none());
+    assert!(us.get(1, 10).is_some());
```
aakoshh, line 109:
```suggestion
    assert!(us.get(1, 10).is_some());
    assert!(us.get(2, 10).is_none());
```

fendermint/testing/contract-test/tests/upgrades.rs
aakoshh: I'd actually enable these as assertions, so you can really test your expectations, and also some kind of assertion in the test that the upgrades _did_ run. Maybe you actually matched the chain ID to the "test"; without a comment it looked just like a random number. If someone changed it, it would be important to test that they didn't just skip all the upgrades. The app version could help with this.

fendermint/testing/contract-test/tests/upgrades.rs
fridrik01: Correct, but the issue is that the MockClient is not cloneable which is AFAICT required by the FvmExecState<MemoryBlockstore> contraint being clonable. Probably there is some way around that, need to check better

fendermint/testing/contract-test/tests/upgrades.rs
fridrik01: Oh, awesome!

fendermint/vm/interpreter/src/fvm/exec.rs
fridrik01: done

fendermint/testing/contract-test/tests/upgrades.rs
fridrik01: exactly, right now I am initializing part of the genesis from outside the tester (accounts), will try to make it as generic as possible

fendermint/testing/contract-test/tests/upgrades.rs
aakoshh: But it's not part of the execution state, it's only a dependency of the interpreter, which stays the same throughout the execution.

fendermint/testing/contract-test/tests/upgrades.rs
fridrik01: Yes, indeed, using chain name would be more convenient!

fendermint/vm/genesis/src/lib.rs
```diff
@@ -47,6 +47,8 @@ pub struct Genesis {
     /// IPC related configuration, if enabled.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub ipc: Option<ipc::IpcParams>,
+    /// The application protocol version.
+    pub app_version: u64,
```
aakoshh, line 51: Don't you think it would be okay to say that at genesis the app version is always 0, and subsequent upgrades can set it to whatever they like?

fendermint/app/src/app.rs
```diff
@@ -749,7 +752,16 @@ where
...
         // TODO: Return events from epoch transitions.
         let ret = self
-            .modify_exec_state(|s| self.interpreter.end(s))
+            .modify_exec_state(|state| async {
+                let ((env, mut state), res) = self.interpreter.end(state).await?;
+
+                // update the app_version in case there was an upgrade defined at this height which upgraded it
+                state.update_app_version(|app_version| {
+                    *app_version = res.app_version;
+                });
```
aakoshh, line 761: This could be done in the interpreter where the upgrades are applied.

fendermint/vm/genesis/src/lib.rs
aakoshh: The only reason I could think of for _not_ to start with 0 is if there was some kind of a configuration that operators can opt-into, which results in a descriptor that other chains can interpret, e.g. if the thing supported multiple protocols, it could describe that "this chain supports IPC v6.0" and then other chains that only do v5.x would know that they arent' compatible.  

But we have a single number, which I think like a private version number for this ledger, which just started. We think upgrades will be coming from either endorsed IPC upgrades mixed with subnet specific ones, so the sequence of upgrades will mean completely different things in different subnets. It's just a means to compare nodes on the same subnet.

fendermint/app/src/app.rs
```diff
@@ -407,7 +409,7 @@ where
         Message = Vec<u8>,
         BeginOutput = FvmApplyRet,
         DeliverOutput = BytesMessageApplyRes,
-        EndOutput = PowerUpdates,
+        EndOutput = FvmEndOutput,
```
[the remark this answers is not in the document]
aakoshh: Because we aren't doing anything with it, it's actually a needless constraint from the `App` to demand that it's the `end_block` where upgrades are applied, and not the `begin_block` for example. I can't think of a reason to prefer one over the other, so we can leave it at the discretion of the interpreter I think. 

Sorry, I know it was my idea.

fendermint/app/src/app.rs
```diff
@@ -501,6 +503,7 @@ where
                 circ_supply: out.circ_supply,
                 chain_id: out.chain_id.into(),
                 power_scale: out.power_scale,
+                app_version: out.app_version,
```
aakoshh, line 506: Don't forget to serve this in `App::info`.

fendermint/app/src/app.rs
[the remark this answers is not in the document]
fridrik01: No problem, I'll change it back and call `update_app_version` inside the interpreter on upgrades

fendermint/vm/genesis/src/lib.rs
fridrik01: @aakoshh Are you suggesting removing `app_version` from the `Genesis` and just assume for now that is should always start with `0`? That makes sense, we _could_ add that in later if needed I guess.

fridrik01: Done, removed from Genesis!

fendermint/testing/contract-test/src/lib.rs
```diff
@@ -56,3 +69,182 @@ pub async fn init_exec_state(
...
+    }
+
+    /// Take the execution state during block execution. Has to be non-empty.
+    async fn take_exec_state(&self) -> FvmExecState<MemoryBlockstore> {
+        let mut guard = self.exec_state.lock().await;
+        guard.take().expect("exec state empty")
+    }
+
+    pub async fn begin_block(&self, block_height: ChainEpoch) -> Result<()> {
+        // TODO: generate block hash based on input
```
aakoshh, line 194: Maybe like the timestamp this could be based on the `block_height` as well, so it's not a constant, just in case something relies on its uniqueness.

fendermint/testing/contract-test/src/lib.rs
```diff
@@ -56,3 +69,182 @@ pub async fn init_exec_state(
...
+            },
+            _,
+        ) = exec_state.commit().context("failed to commit FVM")?;
+
+        self.state_params.state_root = state_root;
+        self.state_params.power_scale = power_scale;
+        self.state_params.circ_supply = circ_supply;
+        self.state_params.app_version = app_version;
+
+        println!("self.state_params: {:?}", self.state_params);
```
aakoshh, line 242:
```suggestion
        eprintln!("self.state_params: {:?}", self.state_params);
```
I tend to use `eprintln!` for debug logs in tests, so `println!` can be reserved for stuff you really want as output, in general.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+use fendermint_vm_interpreter::fvm::store::memory::MemoryBlockstore;
+use fendermint_vm_interpreter::fvm::upgrade_scheduler::UpgradeScheduler;
+use fendermint_vm_interpreter::fvm::{
+    bundle::contracts_path, upgrade_scheduler::Upgrade, FvmMessageInterpreter,
+};
+use fendermint_vm_message::chain::ChainMessage;
+
+fn eth_addr_to_h160(eth_addr: &EthAddress) -> H160 {
+    ethers::core::types::Address::from_slice(&eth_addr.0)
+}
```
aakoshh, line 36: This conversion exists as a `From` [here](https://github.com/consensus-shipyard/ipc/blob/6282ccc3f214b2de795dae4838c8f9ae7ffeea61/fendermint/vm/actor_interface/src/eam.rs#L93-L97).

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+                                .context("error parsing contract")
+                                .unwrap(),
+                        ),
+                        Bytes::default(),
+                        TokenAmount::default(),
+                        GAS_PARAMS.clone(),
+                    )
+                    .unwrap()
+                {
+                    ChainMessage::Signed(signed) => signed.into_message(),
```
aakoshh, line 92: I see, I forgot that the message factory actually signs messages as well, which is a bit inconvenient in our case, to have to produce a throwaway secret key just for this. Maybe there should be an `FvmMessageFactory` underneath that just has the `from` set to some arbitrary FVM `Address` and doesn't sign anything; not sure if it should own and increment the nonce or not. The `MessageFactory` can wrap to sign messages, but we could just not use it here.

aakoshh: And then we could deploy as the system actor, although again I forgot how the address would be assigned in that case. Maybe that salted version of actor creation would be useful, with a deterministic salt based on the block height.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+                    &res.msg_receipt.return_data,
+                )
+                .unwrap();
+                assert_eq!(
+                    res.delegated_address(),
+                    Address::from_str(CONTRACT_ADDRESS).unwrap()
+                );
+
+                // lets upgrade the app version as well
+                let app_version = 1;
```
aakoshh, line 115: Should the app version be an actual _property_ of the upgrade, like the block height itself? That way we can validate that successive upgrades are incrementing the app version, without running them. What's the argument for making it a computed value?

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+                let simple_coin = SimpleCoin::new(EthAddress::from_id(101), client.into());
+                let call = simple_coin.send_coin(
+                    // the address we are sending the balance to (which is us in this case)
+                    eth_addr_to_h160(&EthAddress::from(my_secret_key().public_key())),
+                    // the amount we are sending
+                    U256::from(SEND_BALANCE_AMOUNT),
+                );
+
+                // create a message for sending the balance
+                let mut mf = MessageFactory::new_secp256k1(my_secret_key(), 1, state.chain_id());
```
aakoshh, line 142: Is it account sending the tokens to itself? Does `from` equal `to`?

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+                // execute the message
+                let (res, _) = state.execute_implicit(message).unwrap();
+                assert!(
+                    res.msg_receipt.exit_code.is_success(),
+                    "{:?}",
+                    res.failure_info
+                );
+
+                // this upgrade didn't affect our app version, so return 0
+                Ok(0)
```
aakoshh, line 165: Right. Maybe return `None` - to make it more difficult for people to misunderstand. In my experience people tend to forget about 0 being a special case. I used this for the Solidity contracts when we have no change in the validator set, because Solidity doesn't have `Option`, but had to explain it many times later.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,6 +212,36 @@ where
...
+                Ok(new_app_version) => {
+                    tracing::info!(result =? res, "upgrade done");
+
+                    // update the app_version in case there was an upgrade defined at this height which upgraded it
+                    if new_app_version != 0 {
+                        state.update_app_version(|app_version| {
+                            *app_version = new_app_version;
+                        });
+
+                        tracing::info!(app_version = state.app_version(), "upgraded app version");
```
aakoshh, line 235: @raulk do you have ready ideas about the best way to use `app_version`? CometBFT recognises a single number for the app version, so we thought upgrades are one of the ways we can evolve them, but there are some question:
* Should it be mandatory for each upgrade to increment the app version?
* Can the app version go backwards?
* Can it increase by multiple numbers at a time? 
* Is it a computed value or something predetermined by the upgrade?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,6 +212,36 @@ where
...
+        if let Some(upgrade) = self.upgrade_scheduler.get(chain_id, block_height) {
+            // TODO: consider using an explicit tracing enum for upgrades
+            tracing::info!(?chain_id, height = block_height, "Executing an upgrade");
+
+            // there is an upgrade scheduled for this height, lets run the migration
+            let res = upgrade.execute(&mut state);
+
+            match res {
+                Ok(new_app_version) => {
+                    tracing::info!(result =? res, "upgrade done");
```
aakoshh, line 227: We already know the result is `Ok`, so I'm not sure why we wouldn't just log the contents.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,6 +212,36 @@ where
...
+                            *app_version = new_app_version;
+                        });
+
+                        tracing::info!(app_version = state.app_version(), "upgraded app version");
+                    }
+                }
+                Err(err) => {
+                    tracing::error!(?err, "Upgrade failed");
+                    return Err(err);
+                }
```
aakoshh, line 241: I reckon you could use `let app_version = upgrade.execute(&mut state).context("upgrade failed")?`; and let the top handler do the logging.

fendermint/vm/interpreter/src/fvm/mod.rs
```diff
@@ -27,7 +28,7 @@ pub use query::FvmQueryRet;
 use tendermint_rpc::Client;
 
 pub use self::broadcast::Broadcaster;
-use self::state::ipc::GatewayCaller;
+use self::{state::ipc::GatewayCaller, upgrade_scheduler::UpgradeScheduler};
```
aakoshh, line 31:
```suggestion
use self::{state::ipc::GatewayCaller, upgrades::UpgradeScheduler};
```

Maybe we can simplify the module name?

fendermint/vm/interpreter/src/fvm/mod.rs
```diff
@@ -56,7 +57,10 @@ impl<C> ValidatorContext<C> {
 
 /// Interpreter working on already verified unsigned messages.
 #[derive(Clone)]
-pub struct FvmMessageInterpreter<DB, C> {
+pub struct FvmMessageInterpreter<DB, C>
+where
+    DB: fvm_ipld_blockstore::Blockstore + 'static + Clone,
```
aakoshh, line 62:
```suggestion
    DB: Blockstore + 'static + Clone,
```
Just a bit of a mouthful and usually it appears without the prefix.

fendermint/vm/interpreter/src/fvm/upgrade_scheduler.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// UpgradeScheduler represents a list of upgrades to be executed at given heights
+/// During each block height we check if there is an upgrade scheduled at that
+/// height, and if so the migration for that upgrade is performed.
+#[derive(Clone)]
+pub struct UpgradeScheduler<DB>
+where
+    DB: Blockstore + 'static + Clone,
+{
+    // TODO: Consider using Set
```
aakoshh, line 67: Why not use a `BTreeMap` which is ordered by chain ID and block height?

fendermint/vm/interpreter/src/fvm/upgrade_scheduler.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.upgrades.push(upgrade);
+
+        Ok(())
+    }
+
+    // check if there is an upgrade scheduled for the given chain_id at a given height
+    pub fn get(&self, chain_id: ChainID, height: BlockHeight) -> Option<&Upgrade<DB>> {
+        self.upgrades
+            .iter()
+            .find(|u| u.chain_id == chain_id && u.block_height == height)
```
aakoshh, line 117: It is highly unlikely that this would cause a performance issue, but a `BTreeMap` would really be more efficient.

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+    abigen!(SimpleCoin, "../contracts/SimpleCoin.abi");
+    // once we deploy this contract, this is the address we expect the contract to be deployed to
+    const CONTRACT_ADDRESS: &str = "f410fnz5jdky3zzcj6pejqkomkggw72pcuvkpihz2rwa";
+    // the amount we want to send to the contract
+    const SEND_BALANCE_AMOUNT: u64 = 1000;
+
+    let mut upgrade_scheduler = UpgradeScheduler::new();
+    upgrade_scheduler
+        .add(
+            Upgrade::new("mychain".to_string(), 1, |state| {
```
aakoshh, line 71: Maybe take `impl Into<String>` or `impl ToString` for the chain name so that it accepts just `"mychain"` for this manually configured part?

fendermint/testing/contract-test/tests/upgrades.rs
```diff
@@ -0,0 +1,281 @@
...
+            Upgrade::new("mychain".to_string(), 2, |state| {
+                println!(
+                    "[Upgrade at height {}] Sends a balance",
+                    state.block_height()
+                );
+
+                // build the calldata for the send_coin function
+                let (client, _mock) = ethers::providers::Provider::mocked();
+                let simple_coin = SimpleCoin::new(EthAddress::from_id(101), client.into());
+                let call = simple_coin.send_coin(
```
aakoshh, line 134: I think that after it's been deployed you can even use the `ContractCaller` to make implicit calls to the thing. Here's an [example](https://github.com/consensus-shipyard/ipc/blob/6282ccc3f214b2de795dae4838c8f9ae7ffeea61/fendermint/testing/contract-test/src/ipc/subnet.rs#L63-L65) of using it to join a subnet by a non-system actor.

aakoshh (commented):
Good stuff! 

I don't see any blockers, but would like to hear @raulk's thoughts on the app version. 

I think using the `ContractCaller` would make the tests easier to read. Maybe it would be worth having some more direct support for deploying contracts as well in the upgrades, e.g so they can be deployed with a system ID, or replace a system actor. 

The tests seem to use a deterministic secret key with nonce=1 to deploy the contract, which makes its address predictable, which is then hardcoded in the test, but I'm not sure this is realistic. I mean in tests, sure, but for a real upgrade, it would create an actual usable secret key that anyone can use to sign transactions, and if the contract would use it as its owner, that would be a vulnerability. It would be better to use a public key that we don't have a secret key for or the system address. 

Then the address of the deployed contract - it depends on the nonce which the tests hardcode to 1 every time, so if someone were to blindly copy this and try to deploy a new contract, it would get the same address, because with implicit execution it would not check the nonce. We should look into a version that uses that _salt_ I saw in the EAM actor to isolate upgrades from each other.

fendermint/vm/interpreter/src/fvm/upgrade_scheduler.rs
fridrik01: @aakoshh yeah, had it BTreeMap originally, but noticed that `ChainID` did not implement the `std::cmp::Ord` trait so left a todo instead.

But will add it in now, I think it should work to add a custom key type which implements the Ord trait that allows me to call `Into` on the ChainID

fendermint/testing/contract-test/tests/upgrades.rs
fridrik01: yes, its sending coin to itself

fendermint/testing/contract-test/tests/upgrades.rs
fridrik01: I think that makes sense as it pulls the version up to the top making it also visible that its indeed upgrading the app version (without requiring to parse the `MigrationFunc` code)

fendermint/testing/contract-test/tests/run_upgrades.rs
```diff
@@ -0,0 +1,272 @@
...
+        .add(
+            Upgrade::new("mychain", 1, Some(1), |state| {
+                println!(
+                    "[Upgrade at height {}] Deploy simple contract",
+                    state.block_height()
+                );
+
+                // create a message for deploying the contract
+                let mut mf =
+                    SignedMessageFactory::new_secp256k1(my_secret_key(), 1, state.chain_id());
```
aakoshh, line 73: Do we still have to use this or can we just use the `MessageFactory` that doesn't sign now?

fendermint/testing/contract-test/tests/run_upgrades.rs
```diff
@@ -0,0 +1,272 @@
...
+                                .context("error parsing contract")
+                                .unwrap(),
+                        ),
+                        Bytes::default(),
+                        TokenAmount::default(),
+                        GAS_PARAMS.clone(),
+                    )
+                    .unwrap()
+                {
+                    ChainMessage::Signed(signed) => signed.into_message(),
```
aakoshh, line 87: So that we don't have to do this unwrapping?

fendermint/testing/contract-test/tests/run_upgrades.rs
```diff
@@ -0,0 +1,272 @@
...
+
+                // parse the message receipt data and make sure the contract was deployed to the expected address
+                let res = fvm_ipld_encoding::from_slice::<eam::CreateReturn>(
+                    &res.msg_receipt.return_data,
+                )
+                .unwrap();
+                assert_eq!(
+                    res.delegated_address(),
+                    Address::from_str(CONTRACT_ADDRESS).unwrap()
+                );
```
aakoshh, line 107: Regarding this, I thought that we could investigate how to use this [salted version](https://github.com/filecoin-project/builtin-actors/blob/a66372756f7f42cecd661d5bc5ac64d5a3847bb3/actors/eam/src/lib.rs#L57) to achieve a predictable address. 

The `fevm_create` method uses `CreateExternal` which ends up depending on the nonce, which is why I said that if this is how we recommend an upgrade deploy a contract, then we'd need to be careful because `execute_implicit` doesn't check nonces, so you could potentially try to overwrite an existing contract. 

But I am confused about all these create methods and not sure what are the rules.

fendermint/testing/contract-test/tests/run_upgrades.rs
```diff
@@ -0,0 +1,272 @@
...
+                let call = simple_coin.send_coin(
+                    // the address we are sending the balance to (which is us in this case)
+                    EthAddress::from(my_secret_key().public_key()).into(),
+                    // the amount we are sending
+                    U256::from(SEND_BALANCE_AMOUNT),
+                );
+
+                // create a message for sending the balance
+                let mut mf =
+                    SignedMessageFactory::new_secp256k1(my_secret_key(), 1, state.chain_id());
```
aakoshh, line 135: See you are using the same nonce and same secret key here, which means if this migration also tried to deploy a contract, it would compute the same address.

aakoshh (approved):
Thanks for the updates! The only thing I don't understand is why the tests still use the `SignedMessageFactory`. 

Not sure what to do about the deployment address - it could be assigned to some global value just to remove the hardcoding, but like I say ideally I'd try to see if we can use `CREATE2` and deploy with the System Actor address, rather than any particular secret key. 

That way the address would be something like:

```
new_addr = hash(sender, salt, source-code)
```
where we could set the `salt` to depend on the upgrade itself, e.g. the height of it, and `sender` would be the system address which doesn't have a private key. This way it would be a deterministic and unique function of the block height and the contract code.

fridrik01:
> Not sure what to do about the deployment address - it could be assigned to some global value just to remove the hardcoding,

Hmm, unless I am misunderstanding its not possible to access global variables in `MigrationFunc` as they are not known at compile time

> but like I say ideally I'd try to see if we can use `CREATE2` and deploy with the System Actor address, rather than any particular secret key.
> 
> That way the address would be something like:
> 
> ```
> new_addr = hash(sender, salt, source-code)
> ```
> 
> where we could set the `salt` to depend on the upgrade itself, e.g. the height of it, and `sender` would be the system address which doesn't have a private key. This way it would be a deterministic and unique function of the block height and the contract code.

That makes sense, but not sure how to set that up as `create2` requires that [only EVM actors can call it](https://github.com/filecoin-project/builtin-actors/blob/master/actors/eam/src/lib.rs#L260).

PR #720, opened 2024-02-19
instrument tracing

cryptoAtwill:
Closing ENG-657

Adding instrument attribute to emit IPC events in fendermint. Currently the span include the full span lifecycle such as `enter` and all the way to `close`, example:
```
{"timestamp":"2024-02-19T10:45:08.270754Z","level":"INFO","fields":{"message":"enter"},"filename":"fendermint/app/src/app.rs","line_number":747,"span":{"request":"EndBlock { height: 984 }","name":"new-block"},"spans":[{"request":"EndBlock { height: 984 }","name":"new-block"}]}
{"timestamp":"2024-02-19T10:45:08.273989Z","level":"INFO","fields":{"message":"exit"},"filename":"fendermint/app/src/app.rs","line_number":747,"span":{"request":"EndBlock { height: 984 }","name":"new-block"},"spans":[]}
```
We can tune this in the future once we have grafana up to save log spaces, but so far it should work now.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -40,6 +40,7 @@ use num_traits::Zero;
 use serde::{Deserialize, Serialize};
 use tendermint::abci::request::CheckTxKind;
 use tendermint::abci::{request, response};
+use tracing::instrument;
 
 use crate::{tmconv::*, VERSION};
 use crate::{BlockHeight, APP_VERSION};
@@ -743,6 +744,7 @@ where
     }
 
     /// Signals the end of a block.
+    #[instrument(skip(self))]
     async fn end_block(&self, request: request::EndBlock) -> AbciResult<response::EndBlock> {
         tracing::debug!(height = request.height, "end block");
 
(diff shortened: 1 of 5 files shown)


raulk:
@cryptoAtwill thanks! Could you elaborate on why you chose to use #[instrument] (which is geared towards emitting start/end spans) versus manually recording point-in-time events?

maciejwitowski:
@cryptoAtwill The goal of [Add instrumentation to critical events](https://linear.app/interplanetary-consensus/issue/ENG-657/add-instrumentation-to-critical-events) was to add logs for the critical events, with very specific metadata (listed in the issue, but may need more thinking) in order to investigate issues with running subnets. 
This way we'll see in Fendermint logs things like `log_name: latest_parent_height_changes; height: 2000`

IIUC this PR adds tracing which is a lower priority at the moment, and not part of this task.

cryptoAtwill:
@maciejwitowski Yeah, I checked, it's all logged. I will change to `info!`. But using `#[instrument]`, we can easily log input parameters and return easily, by using `FmtSpan::Close`, it's effectively `info!`. We can also trace the execution time.

PR #723, opened 2024-02-19
Adding initial events from fendermint

cryptoAtwill:
Adding initial set of events for logging from fendermint.

The approach is rather simple at this stage to avoid complex structs/macro. The event type is centralised in `core` crate then other crate can call `emit!` to emit the event in the format that is most suitable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,7 +192,6 @@ fil_actor_eam = { git = "https://github.com/filecoin-project/builtin-actors", ta
 fil_actors_runtime = { git = "https://github.com/filecoin-project/builtin-actors", tag = "v12.0.0" }
 
 fendermint_actor_eam = { path = "./fendermint/actors/eam" }
-fendermint_vm_core = { path = "./fendermint/vm/core" }
 
 cid = { version = "0.10.1", default-features = false, features = [
   "serde-codec",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,6 +192,7 @@ fil_actor_eam = { git = "https://github.com/filecoin-project/builtin-actors", ta
 fil_actors_runtime = { git = "https://github.com/filecoin-project/builtin-actors", tag = "v12.0.0" }
 
 fendermint_actor_eam = { path = "./fendermint/actors/eam" }
+fendermint_vm_core = { path = "./fendermint/vm/core" }
 
 cid = { version = "0.10.1", default-features = false, features = [
   "serde-codec",
@@ -199,6 +200,8 @@ cid = { version = "0.10.1", default-features = false, features = [
 ] }
 
 frc42_dispatch = "6.0.0"
+
+strum_macros = "0.25.2"
 
 # Using the same tendermint-rs dependency as tower-abci. From both we are interested in v037 modules.
 tower-abci = { version = "0.7" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -201,8 +201,6 @@ cid = { version = "0.10.1", default-features = false, features = [
 
 frc42_dispatch = "6.0.0"
 
-strum_macros = "0.25.2"
-
 # Using the same tendermint-rs dependency as tower-abci. From both we are interested in v037 modules.
 tower-abci = { version = "0.7" }
 tendermint = { version = "0.31", features = ["secp256k1"] }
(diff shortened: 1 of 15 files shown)


fendermint/vm/core/src/event.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+#[derive(strum_macros::Display)]
```
raulk, line 4: I would move this to a top-level `fendermint/event` package, and rename it to `EventType`.

fendermint/vm/core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fendermint_vm_core"
-description = "Core data structures shared between genesis and messages"
+description = "Core data structures shared between the vm"
```
raulk, line 3: ?

Cargo.toml
```diff
@@ -200,6 +201,8 @@ cid = { version = "0.10.1", default-features = false, features = [
 
 frc42_dispatch = "6.0.0"
 
+strum_macros = "0.25.2"
```
raulk, line 204: You already have `strum` in the workspace, which exports the macros; no need to import this crate specifically.

fendermint/vm/core/src/event.rs
```diff
@@ -0,0 +1,22 @@
...
+pub enum VMEvent {
+    NewParentNullRound,
+    NewParentNonNullRound,
```
raulk, line 7: These should be the same event with a `null_round` bool field.

fendermint/vm/genesis/Cargo.toml
```diff
@@ -26,7 +26,7 @@ fendermint_actor_eam = { workspace = true }
 
 fendermint_crypto = { path = "../../crypto" }
 fendermint_testing = { path = "../../testing", optional = true }
-fendermint_vm_core = { path = "../core" }
+fendermint_vm_core = { workspace = true }
```
raulk, line 29: Why this change? What do we gain?

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -116,6 +121,14 @@ where
         power_diff(curr_power_table, next_power_table)
     };
 
+    emit!(
+        VMEvent::NewBottomUpCheckpoint,
+        height = height.value(),
+        block_hash = hex::encode(block_hash),
```
raulk, line 127: Perhaps we add a fmt::Display implementation to whatever type `block_hash` is instead? I see us doing this in several places.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -116,6 +121,14 @@ where
         power_diff(curr_power_table, next_power_table)
     };
 
+    emit!(
+        VMEvent::NewBottomUpCheckpoint,
```
raulk, line 125: What level is this emitted at exactly? And what's the name of this field?

raulk (commented):
I think this pattern works, but let's pay attention to:

1. What events end up looking like in practice (can you paste the output?).
2. DRY implementation, e.g. we might need to add some fmt::Display implementations for a few types.
3. Clean event schema, let's try not to explode on event types, and instead favour fields to indicate variants.

fendermint/vm/genesis/Cargo.toml
cryptoAtwill: oh, it's just using the workspace build, so that there is just one build, so the crates all use the same version. Should not have any effect.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
cryptoAtwill: block_hash is `Vec<u8>` or slice, unless we crate a wrapper type, it's kind of difficult to impl that.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
cryptoAtwill: I used `info!` and field name is `event`.

cryptoAtwill:
@fridrik01 The rest of the metrics are from harder to get, most are from combetbft I think. The ABCI ones we can use `#[instrument]`, should be easily doable.

fridrik01 (approved):

PR #726, opened 2024-02-20
FIX: Allow the top range in get_logs to be higher than what is available

aakoshh:
## Error
I've tried to connect our Rust Peer (aka Nox) to the IPC, and got this error
```
json rpc error: Chain JSON RPC error: Tendermint RPC error: response error\\n\\nCaused by:\\n    Internal error: height 10004 must be less than or equal to the current blockchain height 9542 (code: -32603)
```
It comes from `eth_getLogs`. Probably it doesn't like that we ask for ranges where right boundary is in the future.
Here's an example:
```
% curl https://ipc-stage.fluence.dev/ \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_getLogs","params":[{"address": "0x56526B14e265c143E177fBd14d4b6a8bc77445C9", "fromBlock": "0x0", "toBlock": "0xf00d"}],"id":2,"jsonrpc":"2.0"}'

{"jsonrpc":"2.0","error":{"code":0,"message":"Tendermint RPC error: response error\n\nCaused by:\n    Internal error: height 61453 must be less than or equal to the current blockchain height 10245 (code: -32603)\n\nLocation:\n    /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9"},"id":2}%
```
What Nox does, it goes through the chain in ranges: [0, 500], [501, 1001], ... and so on. So we always move right boundary by a constant. In the request above, toBlock is the right boundary, and usually Ethereum RPC doesn't put restrictions on it, and it makes sense because otherwise software would need to call `eth_blockNumber` before it can call `eth_getLogs`.

## Fix
The PR changes `get_logs` to trim the top range to what is available on the chain.


fendermint/eth/api/examples/ethers.rs
```diff
@@ -621,6 +621,14 @@ where
         |logs| *logs == receipt.logs,
     )?;
 
+    // Check that requesting logs with higher-than-highest height does not fail.
+    request(
+        "eth_getLogs /w too high",
```
raulk, line 626:
```suggestion
        "eth_getLogs /w too high 'to' height",
```

raulk (approved):

PR #727, opened 2024-02-20
Remove redundant quickstart-calibration after it was merged into official docs

maciejwitowski:
https://docs.ipc.space/quickstarts/deploy-a-subnet


raulk (commented):
Do we not have any links pointing to this anywhere? It has been a common resource so far and an HTTP 404 with no indication where to go next would be poor experience. Would suggest adding a "Moved to [link]" message instead.

maciejwitowski:
That's a very good point, updated!

PR #728, opened 2024-02-20
FIX: Propagate the return_data in transaction submission errors

aakoshh:
## Error

We have a EVM contract method for checking randomx proofs with delegatecalling precompiled contract for checks. It works normally, but sometimes we have revert with no data:
```
 code: 'UNKNOWN_ERROR',
  error: {
    code: 33,
    message: 'message failed with backtrace:\n' +
      '00: t0124 (method 3844450837) -- contract reverted (33)\n' +
      '01: t0124 (method 6) -- contract reverted (33)\n',
    data: ''
  },
```
  we usually have these errors at the end of a block (I mean in time – just before a new block is mined)
Could it be an IPC bug? As I understand, method 3844450837 is for invoking the EVM

## Fix

The PR only attempts to fix the `data` field being empty in the error above. As it turned out the `return_data` hasn't been returned from the `check_tx` method, because it was originally written when we did not execute the transaction in check. (This can be disabled in the config, but by default it is on). 

Hopefully with this fix whatever the reason is for the contract to revert will be revealed to us in the form of a hex encoded error message from the EVM.


raulk (approved):
This is great, thanks @aakoshh! While we're at it, we should also intelligently parse error return data to return human-friendlier messages to users during gas estimation (and perhaps eth_call)? https://github.com/filecoin-project/lotus/blob/8ba491b6b475f34caa141b5c6edebd508be11a8a/node/impl/full/eth_utils.go#L301

raulk:
@aakoshh Yep, agree. The data should be left alone, but some RPCs like `eth_call` and `eth_estimateGas` don't actually return the receipt (and therefore they don't relay the return data). In those cases, clients appreciate an (ideally useful) error if something goes wrong, vs. a generic revert. Will open a Linear issue to track!

PR #735, opened 2024-02-21
BUILD: Enable main-like behaviour on tagged commits

aakoshh:
The PR changes `ci.yaml` that previously on ran on `refs/heads/main` to also execute on tagged commits. The tag will become the tag of the docker release. 

So for example if we do `git tag hackathon` and `git push --tags` then it should build and push a `ghcr.io/consensus-shipyard/fendermint:hackathon` image. 

The code previously wanted to strip the `v` prefix from the tag, e.g. ff the tag was like `v1.2.3` then it would have produced `fendermint:1.2.3` image. I commented this out so it becomes `fendermint:v1.2.3`. CometBFT tags also have `v` in them, so I thought why not leave them as-is.

### Testing

I have done the above actually:
```shell
git checkout docker-push-tagged
git tag hackathon
git push --tagged
```

And it correctly triggered a build on this PR, even though no code changed and the first commit for example didn't run any of the tests. This should result in a docker image being pushed in the end.

### DIFF
--- a/.github/workflows/fendermint-publish.yaml
+++ b/.github/workflows/fendermint-publish.yaml
@@ -74,7 +74,7 @@ jobs:
           VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
 
           # This strips the "v" prefix from the tag name.
-          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+          # [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
 
           # This uses the Docker `latest` tag convention.
           [ "$VERSION" == "main" ] && VERSION=latest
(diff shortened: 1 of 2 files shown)


.github/workflows/fendermint-publish.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
           VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
 
           # This strips the "v" prefix from the tag name.
-          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+          # [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
```
maciejwitowski, line 76: Should this line be removed then?

aakoshh: This whole thing was from an example from Github docs, so I kinda left it as reference so we don't have to remember how to massage tags. Could be removed and we're happy with what it does.

PR #736, opened 2024-02-22
Multisubnet test

cryptoAtwill:
Adding more multisubnet test cases.

### DIFF
--- a/contracts/test/integration/MultiSubnet.t.sol
+++ b/contracts/test/integration/MultiSubnet.t.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.23;
 import "forge-std/Test.sol";
 import "../../src/errors/IPCErrors.sol";
 import {EMPTY_BYTES, METHOD_SEND} from "../../src/constants/Constants.sol";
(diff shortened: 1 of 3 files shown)


contracts/test/integration/MultiSubnet.t.sol
```diff
@@ -127,6 +127,105 @@ contract MultiSubnetTest is Test, IntegrationTestBase {
...
+        IpcEnvelope memory resultMsg = CrossMsgHelper.createResultMsg(crossMsg, OutcomeType.Ok, new bytes(0));
+        require(resultMsg.value == 0, "ok receipt should have 0 value");
+
+        IpcEnvelope[] memory msgs = new IpcEnvelope[](1);
+        msgs[0] = resultMsg;
+
+        executeTopDownMsgs(msgs, nativeSubnetName, address(nativeSubnetGateway));
+
+        // works with no state changes
+        assertEq(recipient.balance, amount);
```
maciejwitowski, line 161: I think this test would be easier to understand this test if it started with 
- recipient/parent having balance=3
- caller/child having balance=0 

This way the reader could imagine that the initial state of the tests is when the child released its balance completely.

maciejwitowski (approved):

PR #737, opened 2024-02-22
FIX: Sync as many parent blocks as we can.

aakoshh:
The PR changes the `LotusParentSyncer::sync` method to by default sync as many blocks as it can, until it either reaches the top of the finalized chain, or fills the cache to the brim. This is so that we don't sync 1 block per 10s when trying to catch up with a long history, e.g. after opening up a developer laptop in the morning.


snissn:
Thank you! The solution looks clean and given the explanation of why my history was so far behind I think this should resolve the issues on my laptop. I haven't touched this part of the code base yet so I am hesitant to give a GH approval, but the code looks clean and the test looks valid! ty @aakoshh 🎉

raulk (approved):

PR #742, opened 2024-02-22
FIX: Adjust query height in the eth api

aakoshh:
Ethereum tools like The Graph expect that they can query the _effects_ of the execution of block H at height H. This can lead to problems: for example we deploy a contract at H and it will send an `eth_call` to it at H, but it doesn't exist, because the application state is saved at H+1 to be in line with https://docs.cometbft.com/v0.34/spec/abci/abci#query-1

I did it this way because I think it's important that the CometBFT description of what height to set in queries is respected, even if we don't support proofs based on the state hash. I didn't know what ethereum tools expected, only catered for them by lying about what the latest block height is, so they don't ask for results.

Let's see if the integration tests break.

### DIFF
--- a/fendermint/eth/api/src/state.rs
+++ b/fendermint/eth/api/src/state.rs
@@ -137,7 +137,10 @@ where
     /// Get the height of the latest commit.
     pub async fn latest_height(&self) -> JsonRpcResult<tendermint::block::Height> {
         let res: commit::Response = self.tm().latest_commit().await?;
-        Ok(res.signed_header.header.height)
+        // Return -1 so we don't risk having no data to serve.
+        let h = res.signed_header.header.height.value();
+        let h = h.saturating_sub(1);
+        Ok(Height::try_from(h).context("decrementing should be fine")?)
     }
 
     /// Get the Tendermint block at a specific height.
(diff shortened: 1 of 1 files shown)


fendermint/eth/api/src/state.rs
```diff
@@ -221,19 +224,23 @@ where
...
     /// Return the height of a block which we should send with a query,
     /// or None if it's the latest, to let the node figure it out.
+    ///
+    /// Adjusts the height of the query to +1 so the effects of the block is visible.
     pub async fn query_height(&self, block_id: et::BlockId) -> JsonRpcResult<FvmQueryHeight> {
         match block_id {
             et::BlockId::Number(bn) => match bn {
-                et::BlockNumber::Number(height) => Ok(FvmQueryHeight::from(height.as_u64())),
+                // The client might be asking by height of a block, expecting to see the results.
+                et::BlockNumber::Number(height) => Ok(FvmQueryHeight::from(height.as_u64() + 1)),
```
raulk, line 233: Can we document how +1 here does not break overall API consistency? We might want to document this at the top as general comments.

aakoshh: Thanks for keeping me honest, I'll add more comments. 

My thinking is that as long as we try to make sure that if someone asks for the latest block number we give them the last committed one -1, then _that_ block should be executed, and therefore it's safe to ask for +1 here. And if someone got a block from the subscription, that should be executed too, so again it's okay to send the query to +1, where the results are stored.

raulk (approved):

PR #743, opened 2024-02-23
Better logging of checkpoint and its associated structs

cryptoAtwill:
Adding `Display` implementation for bottom up related type in `ipc-cli` using JSON formating.

Closes ENG-814


ipc/api/src/checkpoint.rs
```diff
@@ -86,3 +86,43 @@ pub struct BottomUpCheckpoint {
...
+impl Display for BottomUpCheckpoint {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "BottomUpCheckpoint(subnet_id = {}, height = {}, hash = {}, next_config_number = {}, msgs = {})",
+            self.subnet_id,
+            self.block_height,
+            hex::encode(&self.block_hash),
+            self.next_configuration_number,
+            join_to_string(self.msgs.as_slice()),
```
aakoshh, line 99: Do you really want to log all messages included in a checkpoint? This isn't even the `Debug` format. Wouldn't the number of messages be enough, perhaps?

ipc/api/src/checkpoint.rs
```diff
@@ -86,3 +86,43 @@ pub struct BottomUpCheckpoint {
...
+            f,
+            "BottomUpCheckpointBundle (checkpoint = {}, signatures = {}, signatories = {}",
+            self.checkpoint,
+            join_to_string(
+                self.signatures
+                    .iter()
+                    .map(hex::encode)
+                    .collect::<Vec<_>>()
+                    .as_slice()
+            ),
```
aakoshh, line 116: Do we really need to see the signatures in the log?

ipc/api/src/checkpoint.rs
```diff
@@ -86,3 +86,43 @@ pub struct BottomUpCheckpoint {
...
+        )
+    }
+}
+
+fn join_to_string<T: Display>(items: &[T]) -> String {
+    items
+        .iter()
+        .map(|v| v.to_string())
+        .collect::<Vec<_>>()
+        .join(",")
```
aakoshh, line 127: Do you want a space between them, maybe a bracket around them?

ipc/cli/src/commands/checkpoint/bottomup_bundles.rs
```diff
@@ -27,12 +27,12 @@ impl CommandLineHandler for GetBottomUpBundles {
...
-                "checkpoint: {:?}, signatures: {:?}, signatories: {:?}",
-                bundle.checkpoint, bundle.signatures, bundle.signatories,
-            );
-            println!("{bundle:?}");
+            let Some(bundle) = provider.get_bottom_up_bundle(&subnet, h).await? else {
+                continue;
+            };
+
+            println!("bottom up checkpoint bundle at height: {}", h);
+            println!("{}", bundle);
```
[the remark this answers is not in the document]
cryptoAtwill: I used this one for debugging, to check if the messages matches with expected message, for manually eyeball testing. Previously it's printed bytes, which makes it very difficult to check. Perhaps we don't need the signatures.

ipc/api/src/checkpoint.rs
cryptoAtwill: I used them to debug tests, I mean these cli methods should be used for checking errors, they have been handy when system goes wrong. But I think you are right in the sense that JSON might be better format.

cryptoAtwill:
@aakoshh I have used JSON for the logging of checkpoints.

ipc/api/src/checkpoint.rs
```diff
@@ -53,22 +57,27 @@ impl Display for QuorumReachedEvent {
 pub struct BottomUpCheckpointBundle {
     pub checkpoint: BottomUpCheckpoint,
     /// The list of signatures that have signed the checkpoint hash
+    #[serde(deserialize_with = "deserialize_vec_bytes_from_vec_hex")]
+    #[serde(serialize_with = "serialize_vec_bytes_to_hex")]
```
aakoshh, line 61: Should this say `serialize_vec_bytes_to_vec_hex`, so the two have identical names?

ipc/api/src/checkpoint.rs
```diff
@@ -86,3 +96,81 @@ pub struct BottomUpCheckpoint {
...
+        ret.push(s);
+    }
+    Ok(ret)
+}
+
+pub fn serialize_vec_bytes_to_hex<T: AsRef<[u8]>, S>(data: &[T], s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let values = data.iter().map(hex::encode).collect::<Vec<_>>();
```
aakoshh, line 121: It doesn't looks like there is a need to collect into a `Vec`, you can just iterate over the elements and hex them as you go, no?

ipc/api/src/checkpoint.rs
```diff
@@ -86,3 +96,81 @@ pub struct BottomUpCheckpoint {
...
+
+        let t = T {
+            d: vec![vec![1; 30], vec![2; 30]],
+            subnet_id,
+            ipc_address,
+        };
+        let s = serde_json::to_string(&t).unwrap();
+        assert_eq!(
+            s,
+            r#"{"d":["010101010101010101010101010101010101010101010101010101010101","020202020202020202020202020202020202020202020202020202020202"],"subnet_id":"/r31415926/f2xwzbdu7z5sam6hc57xxwkctciuaz7oe5omipwbq","ipc_address":"/r31415926/f2xwzbdu7z5sam6hc57xxwkctciuaz7oe5omipwbq:f0101"}"#
```
aakoshh, line 169: Curious: what is that `f0101` at the end?

ipc/api/src/cross.rs
```diff
@@ -18,12 +21,15 @@ pub struct IpcEnvelope {
...
     /// can be inspected without having to decode the message.
+    #[serde_as(as = "HumanReadable")]
     pub to: IPCAddress,
     /// Value included in the envelope
     pub value: TokenAmount,
     /// address sending the message
+    #[serde_as(as = "HumanReadable")]
     pub from: IPCAddress,
     /// abi.encoded message
+    #[serde_as(as = "serde_with::hex::Hex")]
```
aakoshh, line 32: Does this affects the actual encoding even inside IPC applications? Will it turn every byte sequence into a `String`, doubling it in size, even if you encode the message as CBOR? Don't you want to do the `HumanReadable` trick to only do `Hex` when the target is JSON or other text format?

ipc/api/src/lib.rs
```diff
@@ -31,3 +33,30 @@ pub fn ethers_address_to_fil_address(addr: &ethers::types::Address) -> anyhow::R
     let eth_addr = EthAddress::from_str(&raw_addr)?;
     Ok(Address::from(eth_addr))
 }
+
+/// Marker type for human readable form.
+pub struct HumanReadable;
```
aakoshh, line 38: The existing `HumanReadable` implementations only apply the string format [if the serializer is human readable](https://github.com/consensus-shipyard/ipc/blob/9fe1ee01e85ef430b77aceccc90a934ec947edee/fendermint/vm/encoding/src/lib.rs#L36-L39).

I reckon you should come up with a different name because this is misleading.

ipc/provider/src/checkpoint.rs
```diff
@@ -146,8 +146,13 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
             return Ok(());
         }
 
-        let bundle = self.child_handler.checkpoint_bundle_at(height).await?;
-        log::debug!("bottom up bundle: {bundle:?}");
+        let Some(bundle) = self.child_handler.checkpoint_bundle_at(height).await? else {
+            return Err(anyhow!(
```
aakoshh, line 150:
```suggestion
            bail!(
```

ipc/provider/src/checkpoint.rs
```diff
@@ -146,8 +146,13 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                "inconsistent contract state, last bottom up checkpoint height occurs at {height} but no checkpoint bundle found"
+            ));
+        };
+
+        log::debug!("bottom up bundle: {}", serde_json::to_string(&bundle)?);
```
aakoshh, line 155:
```suggestion
        log::debug!(bundle = serde_json::to_string(&bundle)?, "bottom up checkpoint");
```

ipc/provider/src/checkpoint.rs
```diff
@@ -193,11 +198,18 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
                     .child_handler
                     .checkpoint_bundle_at(event.height)
-                    .await?;
-                log::debug!("bottom up bundle: {bundle:?}");
+                    .await?
+                else {
+                    return Err(anyhow!(
+                        "inconsistent contract state, quorum reached event found at height {} but not bundle",
+                        event.height
+                    ));
```
aakoshh, line 209:
```suggestion
                    bail!(
                        "inconsistent contract state, quorum reached event found at height {} but not bundle",
                        event.height
                    );
```

ipc/provider/src/checkpoint.rs
```diff
@@ -193,11 +198,18 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                };
+
+                log::debug!("bottom up bundle: {}", serde_json::to_string(&bundle)?);
```
aakoshh, line 212:
```suggestion
                log::debug!(bundle = serde_json::to_string(&bundle)?, height = event.height, "bottom up event");
```

ipc/api/src/checkpoint.rs
cryptoAtwill: That's the ActorID serialized to string.

ipc/api/src/lib.rs
cryptoAtwill: Changed name to `AsString`.

ipc/api/src/checkpoint.rs
```diff
@@ -53,22 +57,27 @@ impl Display for QuorumReachedEvent {
 pub struct BottomUpCheckpointBundle {
     pub checkpoint: BottomUpCheckpoint,
     /// The list of signatures that have signed the checkpoint hash
+    #[serde(deserialize_with = "deserialize_vec_bytes_from_vec_hex")]
+    #[serde(serialize_with = "serialize_vec_bytes_to_vec_hex")]
     pub signatures: Vec<Signature>,
```
aakoshh, line 62: `serde_as` can actually be used in combination with containers. For example further down you have this:
```rust
    #[serde_as(as = "HumanReadable")]
    pub block_hash: Vec<u8>,
```
But here we have a `Vec<Vec<u8>>`. In this situations it can be used like [here](https://github.com/consensus-shipyard/ipc/blob/a33cfee763cafaafdd00ad38a1367ee262e75fa4/fendermint/app/settings/src/resolver.rs#L58-L59): 

```rust
    #[serde_as(as = "Vec<HumanReadable>")]
    pub signatures: Vec<Signature>,
```

It also works with `Option`, etc.

ipc/api/src/checkpoint.rs
```diff
@@ -53,22 +57,27 @@ impl Display for QuorumReachedEvent {
...
+    #[serde_as(as = "HumanReadable")]
     pub subnet_id: SubnetID,
     /// The height of the child subnet at which the batch was cut
     pub block_height: ChainEpoch,
     /// Batch of messages to execute
     pub msgs: Vec<IpcEnvelope>,
 }
 
+#[serde_as]
 #[derive(PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
```
aakoshh, line 81: The title says "Adding display to bottom up checkpoint". Is `Display` supposed to be here?

ipc/api/src/checkpoint.rs
aakoshh: It would be nicer because then in normal encoding it would still use the potentially more space efficient byte encoding (although at this point I'm not actually sure if for that we need to involve the `strict_bytes` module. But at the very least it would be a number, rather than 2 characters per byte.

aakoshh: By contrast I think this might be backwards incompatible: if it met an existing record encoded as IPLD bytes it would look for a `String`.

ipc/api/src/checkpoint.rs
cryptoAtwill: It started with adding display, then slowly grows into better presenting/logging checkpoint and its related structs. Naming is hard. I will change the title.

ipc/api/src/checkpoint.rs
cryptoAtwill: Nice tip, I tried and it works. I have updated accordingly to use:
```
#[serde_as(as = "Vec<HumanReadable>")]
pub signatures: Vec<Signature>,
```

ipc/cli/src/commands/checkpoint/mod.rs
```diff
@@ -35,7 +31,9 @@ pub(crate) struct CheckpointCommandsArgs {
 impl CheckpointCommandsArgs {
     pub async fn handle(&self, global: &GlobalArguments) -> anyhow::Result<()> {
         match &self.command {
-            Commands::ListBottomup(args) => ListBottomUpCheckpoints::handle(global, args).await,
+            Commands::ListBottomup(args) => {
+                LastBottomUpCheckpointHeight::handle(global, args).await
+            }
```
raulk, line 36: What's with the duplication between `ListBottomup` and `LastBottomupCheckpointHeight`?

They call the same handler and accept the same arguments. Does it make sense to remove `LastBottomupCheckpointHeight` in favour of `ListBottomup`?

raulk (approved):

ipc/cli/src/commands/checkpoint/mod.rs
cryptoAtwill: oh yeah, probably need to remove.

PR #744, opened 2024-02-23
ENG-578: IT (Part 5) - Docker start

aakoshh:
Builds on https://github.com/consensus-shipyard/ipc/pull/700

Trying to implement `start_node`. Currently stuck on capturing docker output.


fendermint/testing/materializer/src/docker/dropper.rs
```diff
@@ -0,0 +1,84 @@
...
+    DropContainer(String),
+}
+
+pub type DropHandle = tokio::sync::mpsc::UnboundedSender<DropCommand>;
+
+/// Start a background task to remove docker constructs.
+///
+/// The loop will exit when all clones of the sender channel have been dropped.
+pub fn start(docker: Docker) -> DropHandle {
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
```
fridrik01, line 24: Ah, so this is a nice Rust way to implement the producer/consumer pattern between async tasks, nice!

fendermint/testing/materializer/tests/docker.rs
```diff
@@ -0,0 +1,132 @@
...
+    testnet::Testnet,
+    validation::validate_manifest,
+    TestnetName,
+};
+use futures::Future;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref CI_PROFILE: bool = std::env::var("PROFILE").unwrap_or_default() == "ci";
+    static ref STARTUP_WAIT_SECS: u64 = if *CI_PROFILE { 20 } else { 15 };
```
fridrik01, line 24: because of slow CI I assume?

fendermint/testing/materializer/tests/docker.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Parse a manifest file in the `manifests` directory, clean up any corresponding
+/// testnet resources, then materialize a testnet and run some tests.
+pub async fn with_testnet<F>(manifest_file_name: &str, f: F) -> anyhow::Result<()>
+where
+    // https://users.rust-lang.org/t/function-that-takes-a-closure-with-mutable-reference-that-returns-a-future/54324
+    F: for<'a> FnOnce(
+        &Manifest,
+        &mut DockerMaterializer,
+        &'a mut Testnet<DockerMaterials, DockerMaterializer>,
+    ) -> Pin<Box<dyn Future<Output = anyhow::Result<()>> + 'a>>,
```
fridrik01, line 66: Djeesus :face_with_spiral_eyes:, that is some advanced Rust lifetime syntax! 

Haven't seen this `for<'a>` before, looked it up [here](https://doc.rust-lang.org/stable/reference/trait-bounds.html#higher-ranked-trait-bounds, sounds like fun!

fendermint/testing/materializer/tests/docker.rs
```diff
@@ -0,0 +1,132 @@
...
+{
+    let testnet_name = TestnetName::new(
+        PathBuf::from(manifest_file_name)
+            .file_stem()
+            .expect("there is a file step")
```
fridrik01, line 71: Not sure I understand the error message, maybe?

```suggestion
            .expect("filename missing")
```

fendermint/testing/materializer/tests/docker.rs
```diff
@@ -0,0 +1,132 @@
...
+    let manifest = read_manifest(manifest_file_name)?;
+
+    // First make sure it's a sound manifest.
+    validate_manifest(&testnet_name, &manifest)
+        .await
+        .context("failed to validate manifest")?;
+
+    let mut materializer = DockerMaterializer::new(&test_data_dir(), 0)?;
+
+    materializer
```
fridrik01, line 85:
```suggestion
    // make sure we start with clean slate by removing any previous files
    materializer
```

fendermint/testing/materializer/tests/docker.rs
```diff
@@ -0,0 +1,132 @@
...
+        .remove(&testnet_name)
+        .await
+        .context("failed to remove testnet")?;
+
+    let mut testnet = Testnet::setup(&mut materializer, &testnet_name, &manifest)
+        .await
+        .context("failed to set up testnet")?;
+
+    // Allow time for things to consolidate and blocks to be created.
+    tokio::time::sleep(Duration::from_secs(*STARTUP_WAIT_SECS)).await;
```
fridrik01, line 95: nit: Optimally we want to wait until all nodes are up, maybe we can poll here the RPC API on the cometBFT container until a block has been produced?

fendermint/testing/materializer/src/docker/dropper.rs
aakoshh: Yes, I'd say this is the more common practice, and then with the `select!` macro in `tokio` you can do a whole lot of stuff like prioritising certain queues, doing timeouts, like [here](https://github.com/consensus-shipyard/ipc/blob/b2bdb1f7fd192e7d678cb19bacb92d2b17e1d848/ipld/resolver/src/service.rs#L267). The STM stuff is way less common. What I did with the runtime handle compiled but didn't run.

fendermint/testing/materializer/src/docker/container.rs
```diff
@@ -2,31 +2,53 @@
...
+};
 
 /// Time to wait before killing the container if it doesn't want to stop.
 const KILL_TIMEOUT_SECS: i64 = 5;
 
 pub struct DockerContainer {
-    pub dh: DockerWithDropHandle,
-    pub container: DockerConstruct,
+    docker: Docker,
+    dropper: DropHandle,
```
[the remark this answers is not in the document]
aakoshh: Yeah, it felt like both are needed most of the time, but it was very awkward.

fendermint/testing/materializer/tests/docker.rs
aakoshh: I'm not used to them at all either! The only other place I have used an example for this was [here](https://github.com/consensus-shipyard/ipc/blob/b2bdb1f7fd192e7d678cb19bacb92d2b17e1d848/fendermint/vm/resolver/src/pool.rs#L110) where I read `impl<T> ResolvePool<T> where   for<'a> ResolveKey: From<&'a T>,` as there exists a transformation from "any reference to T to ResolveKey". 

What's happening here, I didn't even know if it was possible. This is a very similar situation to the interpreters which consume some `State` as input and then return it as output, partly because I didn't know this could work. The compromise there was that if the interpreter fails and returns an `Err`, the state is lost, but that was okay there as I said any error like that can result in a panic and shut down Fendermint after logging. Here, however, we want to tear down the testnet, so I need it even if there is an error, but I can't expect it to be _returned_ along with an error. 

So this syntax means something like: whatever lifetime the testnet has, the future returned has the exact same lifetime, so it cannot outlive the testnet it borrowed. I tried first with two lifetimes and tried to say `'a: 'b` and that `Future<...> + 'b` but that didn't work.

fendermint/testing/materializer/tests/docker.rs
aakoshh: Yeah I agree, this is very crude. I think the place to do it though is in the test, where we at least know what nodes should exist, and what interface is available to query. We can always assume CometBFT I suppose, but I'm not sure every test will always be able to progress, maybe some setups are deliberately not starting a node that would allow quorum. 

Should we do these utilities as we add tests? I'm sure patterns will emerge.

aakoshh: Maybe I can add a loop to wait until some general API responds from CometBFT, not necessarily block production.

fridrik01 (approved):
:1st_place_medal:

PR #745, opened 2024-02-26
Remove submission on previous bottom up checkpoints

cryptoAtwill:
Some initial modification to relayer:
- Remove submit previous epoch. This was added to encourage more relayer submit previous epoch and claim rewards. currently we dont have reward scheme anymore. This becomes unmotivated.
- Adding more info into the error log.

### DIFF
--- a/ipc/provider/src/checkpoint.rs
+++ b/ipc/provider/src/checkpoint.rs
@@ -106,18 +106,28 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
         log::info!("launching {self} for {submitter}");
 
         loop {
-            if let Err(e) = self.submit_checkpoint(&submitter).await {
-                log::error!("cannot submit checkpoint for submitter: {submitter} due to {e}");
-            }
-
+            self.submit_checkpoint(&submitter).await;
             tokio::time::sleep(submission_interval).await;
         }
     }
 
     /// Submit the checkpoint from the target submitter address
-    pub async fn submit_checkpoint(&self, submitter: &Address) -> Result<()> {
-        self.submit_last_epoch(submitter).await?;
-        self.submit_next_epoch(submitter).await
+    pub async fn submit_checkpoint(&self, submitter: &Address) {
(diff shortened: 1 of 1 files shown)


aakoshh:
Just wanted to point out that the title says it just adds more logging, yet the description talks about changing the semantics of submission when it comes to rewards. I think the second is the more significant change and should be reflected in the title.

ipc/provider/src/checkpoint.rs
```diff
@@ -106,18 +106,28 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
 
     /// Submit the checkpoint from the target submitter address
-    pub async fn submit_checkpoint(&self, submitter: &Address) -> Result<()> {
-        self.submit_last_epoch(submitter).await?;
-        self.submit_next_epoch(submitter).await
+    pub async fn submit_checkpoint(&self, submitter: &Address) {
+        let next_submission_height = if let Ok(h) = self.next_submission_height().await {
+            h
+        } else {
+            log::error!("cannot fetch next submission height for submitter {submitter}");
```
aakoshh, line 119:
```suggestion
            log::error!(submitter, "cannot fetch next submission height");
```

Would it be worth logging the actual error as well?

ipc/provider/src/checkpoint.rs
```diff
@@ -106,18 +106,28 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+            return;
+        };
+
+        if let Err(e) = self.submit_epoch(next_submission_height, submitter).await {
+            log::error!(
+                "cannot submit at bottom up checkpoint for height: {} and submitter {} due to {}",
+                next_submission_height,
+                submitter,
+                e
+            )
```
aakoshh, line 129:
```suggestion
            log::error!(error = e, height = next_submission_height, submitter,  "cannot submit bottom up checkpoint");
```

cryptoAtwill:
Closing this one as we will skip empty bottom up checkpoints, see #766

PR #746, opened 2024-02-26
Block checkpoint resubmission

cryptoAtwill:
Closes ENG-764.

This is an initial proposal to address the issues raised in the above ticket, see code comment for more discussion.

### DIFF
--- a/contracts/src/subnet/SubnetActorCheckpointingFacet.sol
+++ b/contracts/src/subnet/SubnetActorCheckpointingFacet.sol
@@ -40,21 +40,22 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
         // agree on submission, it's accepted. This is also needed when a bottom up msg bundle
         // becomes full, a bottom up checkpoint will be created before the checkpoint period
         // is reached.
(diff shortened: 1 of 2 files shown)


contracts/src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -32,22 +32,29 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
 
         bytes32 checkpointHash = keccak256(abi.encode(checkpoint));
 
-        if (checkpoint.blockHeight == s.lastBottomUpCheckpointHeight + s.bottomUpCheckPeriod) {
```
cryptoAtwill, line 35: This will not work for a full msg batch. Because a message batch is created if the max number of bottom up cross messages are received, and a new checkpoint will be created, so the checkpoint epoch will not be at the checkpoint period.

cryptoAtwill, line 35: This will not work for a full msg batch. Because a message batch is created if the max number of bottom up cross messages are received, and a new checkpoint will be created, so the checkpoint epoch will not be at the checkpoint period.

PR #747, opened 2024-02-26
FIX: Re-publish finality vote after a minute

aakoshh:
Adds a new setting to the vote publishing loop that causes the last vote to be re-published after a timeout if there are no new events from the syncer. 

The motivation is to handle a potential situation where the finality cache is saturated with blocks, and therefore they don't trigger new heights to vote on, yet for some reason we aren't finalizing anything new, which means the cache isn't being emptied either. This could potentially happen if somehow everyone has missed the gossip to propose new a new finality to the others.

### DIFF
--- a/fendermint/vm/topdown/src/voting.rs
+++ b/fendermint/vm/topdown/src/voting.rs
@@ -322,7 +322,10 @@ where
 /// Poll the vote tally for new finalized blocks and publish a vote about them if the validator is part of the power table.
 pub async fn publish_vote_loop<V, F>(
     vote_tally: VoteTally,
+    // Throttle votes to maximum 1/interval
     vote_interval: Duration,
+    // Publish a vote after a timeout even if it's the same as before.
+    vote_timeout: Duration,
     key: libp2p::identity::Keypair,
     subnet_id: ipc_api::subnet_id::SubnetID,
     client: ipc_ipld_resolver::Client<V>,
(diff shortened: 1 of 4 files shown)


fridrik01:
@aakoshh I recall from huddle that this PR did not fixed the issue, or was I misunderstanding? Do you still want this merged?

aakoshh:
@fridrik01 yes I think this should be merged regardless of whether it fixes the immediate problem, it's more like a situation which if it arises, it would stall finality until people restart their nodes. Not sure it _can_ happen, but who knows.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -336,46 +339,68 @@ pub async fn publish_vote_loop<V, F>(
...
         .await;
 
         let (next_height, next_hash, has_power) = match result {
-            Ok(vs) => vs,
-            Err(e) => {
-                tracing::error!(error = e.to_string(), "faled to get next height to vote on");
+            Ok(Ok(vs)) => vs,
+            Err(_) => {
+                if let Some(ref vs) = prev {
+                    vs.clone()
```
fridrik01, line 383: would it be interesting to log this?

fridrik01 (approved):

fendermint/vm/topdown/src/voting.rs
aakoshh: Yeah, why not!

PR #748, opened 2024-02-26
Eng 766. List bottom up checkpoint status.

cryptoAtwill:
Closes ENG-766.

Adding a new CLI method that lists the checkpoint status in bottom up checkpoint submission.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/748)


ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        println!(
+            "last checkpoint height: {}, chain head {}",
+            height, chain_head
+        );
+        println!("last submitted checkpoint: {:?}", checkpoint);
+
+        let max_pending = arguments.max_pending.unwrap_or(DEFAULT_MAX_PENDING);
+        for h in height + 1..max_pending as ChainEpoch {
+            let c = provider.get_bottom_up_bundle(&subnet, h).await?;
```
aakoshh, line 43: I don't really understand what's happening here. What are we trying to do, what is max_pending?

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -32,22 +32,29 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
 
         bytes32 checkpointHash = keccak256(abi.encode(checkpoint));
 
-        if (checkpoint.blockHeight == s.lastBottomUpCheckpointHeight + s.bottomUpCheckPeriod) {
-            // validate signatures and quorum threshold, revert if validation fails
-            validateActiveQuorumSignatures({signatories: signatories, hash: checkpointHash, signatures: signatures});
+        if (s.committedCheckpoints[checkpoint.blockHeight].blockHeight != 0) {
+            revert BottomUpCheckpointAlreadySubmitted();
```
aakoshh, line 36: I think the reason this wasn't an error condition is because we wanted to allow multiple relayers to submit the checkpoints and add them to pool of relayers to be rewarded. The difference was that the first relayer caused the checkpoint to be executed, while subsequent ones only went through validation of the payload, and being registered for rewards.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -1076,6 +1076,37 @@ impl BottomUpCheckpointRelayer for EthSubnetManager {
 
         Ok(events)
     }
+
+    async fn max_quorum_reached_height(
```
aakoshh, line 1080: It would be helpful to explain what this method is trying to do.

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -1076,6 +1076,37 @@ impl BottomUpCheckpointRelayer for EthSubnetManager {
...
+
+            if self.quorum_reached_events(mid).await?.is_empty() {
+                r = mid - 1;
+            } else {
+                maybe_height = Some(mid);
+                l = mid + 1;
+            }
+        }
+
+        Ok(maybe_height)
```
aakoshh, line 1107: It may be worth factoring out the bisection algorithm and make some unit tests, because it's not a standard one and looks tricky. Not sure what are the invariants it's trying to adhere to. 

I may be wrong but it looks to me like it could get into an infinite loop: 
1. Say our check period is 10, and we call this method with `from = 53` and `to = 62`. 
2. `mid = (53+62)/2 = 57`
3. `mid = mid / 10 * 10` -> I'm not sure what this should do; mathematically it should be 57, but it might give 50 due to truncation. Let's say after this `mid = 50`; that's outside `from` and `to`. Is that okay for the method to return?

4. Say `quorum_reached_events(50)` is _not_ empty. `maybe_height = 50`, and `from = 51`. 
5. Goto 2. `mid = (51+62)/2 = 56`, and we'll enter another loop with `mid = 50`. 

4. Say `quorum_reached_events(50)` is empty. in this case `to = 49` and we exit with an empty result. Maybe there would have been a result at 60?

I'm also not sure what why we're looking for quorum at fixed multiples of the period? Signatures are submitted asynchronously, they can gather at any height. I'm not even sure why we'd be looking for checkpoints added to the ledger at fixed multiples of the period, since batching could trigger a checkpoint at any height.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
aakoshh: I don't really have a pro or contra argument against turning this into an error. Adding rewards will require migration, and undoing this change would require a migration too, but if we need migration anyway then there is little reason to keep accepting duplicate submissions.

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: yeah, there is a `max-pending` flag now, I dont think this would be needed anymore. Might be difficult to add tests for this one too.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
cryptoAtwill: The thing is, we dont have reward now, so I'm not sure if we still want to do that.

aakoshh: Yeah, that's what I mean, when we'll have rewards, this will need a migration even if you leave it alone, so I don't have an opinion on error vs noop.

ipc/cli/src/commands/checkpoint/status.rs
cryptoAtwill: just a parameter that perform look ahead on how many bottom up checkpoint pending submission.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
cryptoAtwill: Noop for sure will not kill so many tests

raulk:
@cryptoAtwill please add a description and title here

jsoares:
This looks close enough to the finish line and it's just non-critical cli code. @aakoshh's comments are all on outdated code, though still open. @aakoshh can we get a re-review?

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[derive(Debug, Args)]
+#[command(about = "List bottom up checkpoint status of the subnet")]
+pub(crate) struct StatusArgs {
+    #[arg(long, help = "The target subnet to check status")]
+    pub subnet: String,
+    #[arg(
+        long,
+        help = "The max number of pending checkpoint to print, default 10"
```
raulk, line 69:
```suggestion
        help = "Limit unsubmitted checkpoints to print (looking forward from last submitted), default: 10"
```

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+        );
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "List bottom up checkpoint status of the subnet")]
+pub(crate) struct StatusArgs {
+    #[arg(long, help = "The target subnet to check status")]
```
raulk, line 65:
```suggestion
    #[arg(long, help = "The subnet to inspect")]
```

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+            "there are at least {} number of check ahead",
+            checkpoints_ahead
+        );
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(about = "List bottom up checkpoint status of the subnet")]
```
raulk, line 63:
```suggestion
#[command(about = "Summarise the bottom up checkpointing status of a subnet")]
```

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+        println!(
+            "last checkpoint height: {}, chain head {}",
+            height, chain_head
+        );
+        println!("last submitted checkpoint: {:?}", checkpoint);
+
+        let max_pending = arguments.max_pending.unwrap_or(DEFAULT_MAX_PENDING);
+
+        let start = height + 1;
+        let ending = max_pending as ChainEpoch * period + start;
```
raulk, line 45: Would be clearer.

```suggestion
        let ending = max_unsubmitted as ChainEpoch * period + start;
```

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+        let period = provider.checkpoint_period(&subnet).await?;
+        let chain_head = provider.get_chain_head_height(&subnet).await?;
+
+        println!(
+            "last checkpoint height: {}, chain head {}",
+            height, chain_head
+        );
+        println!("last submitted checkpoint: {:?}", checkpoint);
+
+        let max_pending = arguments.max_pending.unwrap_or(DEFAULT_MAX_PENDING);
```
raulk, line 42: Would be clearer.

```suggestion
        let limit_unsubmitted = arguments.limit_unsubmitted.unwrap_or(DEFAULT_LIMIT_UNSUBMITTED);
```

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+        let ending = max_pending as ChainEpoch * period + start;
+        let mut checkpoints_ahead = 0;
+        for h in start..=ending {
+            let c = provider.get_bottom_up_bundle(&subnet, h).await?;
+            if c.checkpoint.block_height != 0 {
+                checkpoints_ahead += 1;
+            }
+        }
+        println!(
+            "there are at least {} number of check ahead",
```
raulk, line 54:
```suggestion
            "there are at least {} unsubmitted checkpoints (limiting query to: {})", limit_unsubmitted
```

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+        let provider = get_ipc_provider(global)?;
+        let subnet = SubnetID::from_str(&arguments.subnet)?;
+
+        let height = provider.last_bottom_up_checkpoint_height(&subnet).await?;
+        let checkpoint = provider.get_bottom_up_bundle(&subnet, height).await?;
+        let period = provider.checkpoint_period(&subnet).await?;
+        let chain_head = provider.get_chain_head_height(&subnet).await?;
+
+        println!(
+            "last checkpoint height: {}, chain head {}",
```
raulk, line 37:
```suggestion
            "subnet chain head height: {}",
```

raulk (commented):
This looks good, but we can make the output clearer, and our lives easier by using more specific names for variables.

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,72 @@
...
+        );
+        println!("last submitted checkpoint: {:?}", checkpoint);
+
+        let max_pending = arguments.max_pending.unwrap_or(DEFAULT_MAX_PENDING);
+
+        let start = height + 1;
+        let ending = max_pending as ChainEpoch * period + start;
+        let mut checkpoints_ahead = 0;
+        for h in start..=ending {
+            let c = provider.get_bottom_up_bundle(&subnet, h).await?;
```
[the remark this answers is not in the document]
cryptoAtwill: this is handled in follow up PR: https://github.com/consensus-shipyard/ipc/pull/743

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,76 @@
...
+            "last submitted checkpoint (@ subnet height {}): {:?}",
+            last_checkpointed_height, checkpoint
+        );
+
+        let limit_unsubmitted = arguments
+            .limit_unsubmitted
+            .unwrap_or(DEFAULT_LIMIT_UNSUBMITTED);
+
+        let start = last_checkpointed_height + 1;
+        let ending = limit_unsubmitted as ChainEpoch * period + start;
```
aakoshh, line 49: It would be nice to explain what's happening. 

For example I don't know if `last_checkpointed_height` is a) height in the subnet of the last created checkpoint or b) the last submitted checkpoint in the parent.

ipc/cli/src/commands/checkpoint/status.rs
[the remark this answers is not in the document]
aakoshh: So you query the whole bundle at every height to see if there is a potential checkpoint, right? Can't this be queried with the events API?

ipc/cli/src/commands/checkpoint/status.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+#[derive(Debug, Args)]
+#[command(about = "Summarise the bottom up checkpointing status of a subnet")]
+pub(crate) struct StatusArgs {
+    #[arg(long, help = "The subnet to inspect")]
+    pub subnet: String,
+    #[arg(
+        long,
+        help = "Limit unsubmitted checkpoints to print (looking forward from last submitted), default: 10"
```
aakoshh, line 73: But it doesn't necessarily limit the number of checkpoints to print, only the number of checkpoint _periods_ to look ahead.

ipc/cli/src/commands/checkpoint/status.rs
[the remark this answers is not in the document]
cryptoAtwill: There is indeed a PR that I changed the query from `height to height` to a range query, but somehow that PR is not merged. I can recreate that on a follow up.

ipc/cli/src/commands/checkpoint/status.rs
cryptoAtwill: basically just display a bunch of metrics/fields related to bottom up checkpointing.

PR #750, opened 2024-02-26
FIX: ipc-cli OpenSSL problem

aakoshh:
Fixes the following error during the multiarch docker build:

```
200.3 error: failed to run custom build command for `openssl-sys v0.9.101`
200.3 
200.3 Caused by:
200.3   process didn't exit successfully: `/app/target/release/build/openssl-sys-48695aebe880d626/build-script-main` (exit status: 101)
200.3   --- stdout
200.3   cargo:rerun-if-env-changed=AARCH64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR
200.3   AARCH64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR unset
200.3   cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
200.3   OPENSSL_LIB_DIR unset
200.3   cargo:rerun-if-env-changed=AARCH64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR
200.3   AARCH64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR unset
200.3   cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
200.3   OPENSSL_INCLUDE_DIR unset
200.3   cargo:rerun-if-env-changed=AARCH64_UNKNOWN_LINUX_GNU_OPENSSL_DIR
200.3   AARCH64_UNKNOWN_LINUX_GNU_OPENSSL_DIR unset
200.3   cargo:rerun-if-env-changed=OPENSSL_DIR
200.3   OPENSSL_DIR unset
200.3   cargo:rerun-if-env-changed=OPENSSL_NO_PKG_CONFIG
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_aarch64-unknown-linux-gnu
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_aarch64_unknown_linux_gnu
200.3   cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_ALLOW_CROSS
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_aarch64-unknown-linux-gnu
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_aarch64_unknown_linux_gnu
200.3   cargo:rerun-if-env-changed=TARGET_PKG_CONFIG
200.3   cargo:rerun-if-env-changed=PKG_CONFIG
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_aarch64-unknown-linux-gnu
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_aarch64_unknown_linux_gnu
200.3   cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_SYSROOT_DIR
200.3   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
200.3   run pkg_config fail: pkg-config has not been configured to support cross-compilation.
200.3 
200.3   Install a sysroot for the target platform and configure it via
200.3   PKG_CONFIG_SYSROOT_DIR and PKG_CONFIG_PATH, or install a
200.3   cross-compiling wrapper for pkg-config and set it via
200.3   PKG_CONFIG environment variable.
200.3 
200.3   --- stderr
200.3   thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/openssl-sys-0.9.101/build/find_normal.rs:190:5:
200.3 
200.3 
200.3   Could not find directory of OpenSSL installation, and this `-sys` crate cannot
200.3   proceed without this knowledge. If OpenSSL is installed and this crate had
200.3   trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
200.3   compilation process.
200.3 
200.3   Make sure you also have the development packages of openssl installed.
200.3   For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.
200.3 
200.3   If you're in a situation where you think the directory *should* be found
200.3   automatically, please open a bug at https://github.com/sfackler/rust-openssl
200.3   and include information about your system as well as this message.
200.3 
200.3   $HOST = x86_64-unknown-linux-gnu
200.3   $TARGET = aarch64-unknown-linux-gnu
200.3   openssl-sys = 0.9.101
200.3 
200.3 
200.3   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
200.3 warning: build failed, waiting for other jobs to finish...
206.4 error: failed to compile `ipc-cli v0.1.0 (/app/ipc/cli)`, intermediate artifacts can be found at `/app/target`.
206.4 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
------
Dockerfile:84
--------------------
  83 |     # Do the final build.
  84 | >>> RUN set -eux; \
  85 | >>>   case "${TARGETARCH}" in \
  86 | >>>   amd64) ARCH='x86_64'  ;; \
  87 | >>>   arm64) ARCH='aarch64' ;; \
  88 | >>>   esac; \
  89 | >>>   cargo install --locked --root output --path fendermint/app --target ${ARCH}-unknown-linux-gnu ; \
  90 | >>>   cargo install --locked --root output --path ipc/cli --target ${ARCH}-unknown-linux-gnu
  91 |     # syntax=docker/dockerfile:1
--------------------
ERROR: failed to solve: process "/bin/sh -c set -eux;   case \"${TARGETARCH}\" in   amd64) ARCH='x86_64'  ;;   arm64) ARCH='aarch64' ;;   esac;   cargo install --locked --root output --path fendermint/app --target ${ARCH}-unknown-linux-gnu ;   cargo install --locked --root output --path ipc/cli --target ${ARCH}-unknown-linux-gnu" did not complete successfully: exit code: 101
make: *** [Makefile:85: docker-build] Error 1
```

To reproduce the problem on Ubuntu I ran the following:

```shell
PROFILE=ci BUILDX_FLAGS="--platform linux/arm64" make docker-build
```

This took me ages to "solve" when I first did the multiarch Docker image. I couldn't even make it work on `debian:bookworm` IIRC, but on `ubuntu:latest` at the end adding a dependency on `openssl` with the `vendored` feature worked: that way the library did not have to look for the C headers, they were included in the package. 

The reason we had a problem with `ipc-cli` is that it only depends `openssl` transitively. I spotted this comparing the outputs of the following:

```shell
cargo tree --package fendermint_app -e=no-dev 
cargo tree --package ipc-cli -e=no-dev 
cargo tree --package openssl-sys --invert --no-dev-dependencies
```

From `fendermint_app` we reached `openssl` directly, while from `ipc-cli` it was deeply embedded. This gave me the idea to add the dependency to `ipc-cli` as well, and then it compiles it with the headers and doesn't require it to be found on the machine.

### DIFF
--- a/fendermint/docker/builder.ci.Dockerfile
+++ b/fendermint/docker/builder.ci.Dockerfile
@@ -46,7 +46,7 @@ ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc \
 WORKDIR /app
 
 # Update the version here if our `rust-toolchain.toml` would cause something new to be fetched every time.
-ARG RUST_VERSION=1.75
+ARG RUST_VERSION=1.76
 RUN rustup install ${RUST_VERSION} && rustup target add wasm32-unknown-unknown
 
 # Defined here so anything above it can be cached as a common dependency.
@@ -70,8 +70,9 @@ RUN set -eux; \
   *) echo >&2 "unsupported architecture: ${TARGETARCH}"; exit 1 ;; \
   esac; \
   rustup show ; \
-  cargo build --locked --release -p fendermint_app --target ${ARCH}-unknown-linux-gnu &&\
+  cargo build --locked --release -p fendermint_app --target ${ARCH}-unknown-linux-gnu ; \
   cargo build --locked --release -p ipc-cli --target ${ARCH}-unknown-linux-gnu
+
 
 # Now copy the full source.
 COPY . .
@@ -85,5 +86,5 @@ RUN set -eux; \
   amd64) ARCH='x86_64'  ;; \
   arm64) ARCH='aarch64' ;; \
   esac; \
-  cargo install --locked --root output --path fendermint/app --target ${ARCH}-unknown-linux-gnu &&\
+  cargo install --locked --root output --path fendermint/app --target ${ARCH}-unknown-linux-gnu ; \
   cargo install --locked --root output --path ipc/cli --target ${ARCH}-unknown-linux-gnu
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.8"
+version = "0.8.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42cd52102d3df161c77a887b608d7a4897d7cc112886a9537b738a887a03aaff"
+checksum = "d713b3834d76b85304d4d525563c1276e2e30dc97cc67bfb4585a4a29fc2c89f"
 dependencies = [
  "cfg-if",
  "getrandom",
@@ -148,9 +148,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.11"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e2e1ebcb11de5c03c67de28a7df593d32191b44939c482e97702baaaa6ab6a5"
+checksum = "96b09b5178381e0874812a9b157f7fe84982617e48f71f4e3235482775e5b540"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -196,9 +196,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.79"
+version = "1.0.80"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "080e9890a082662b09c1ad45f567faeeb47f22b5fb23895fbe1e651e718e25ca"
+checksum = "5ad32ce52e4161730f7098c077cd2ed6229b5804ccf99e5366be1ab72a98b4e1"
 dependencies = [
  "backtrace",
 ]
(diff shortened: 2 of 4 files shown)


mb1896 (approved):

fendermint/docker/builder.ci.Dockerfile
```diff
@@ -85,5 +86,5 @@ RUN set -eux; \
   amd64) ARCH='x86_64'  ;; \
   arm64) ARCH='aarch64' ;; \
   esac; \
-  cargo install --locked --root output --path fendermint/app --target ${ARCH}-unknown-linux-gnu &&\
+  cargo install --locked --root output --path fendermint/app --target ${ARCH}-unknown-linux-gnu ; \
```
fridrik01, line 89: hmm, why did you change this (and above)? I guess there is no possibility of hiding errors I guess but still the previous version _feels_ safer :S

fridrik01 (commented):
Linking with openSSL is always so fun :)

fendermint/docker/builder.ci.Dockerfile
aakoshh: I just wanted to see if it makes any difference, although didn't really expect it to do. The `set -eux;` in the beginning causes all this script to fail if anything goes wrong, and because the previous steps were all separated by `;` I thought we should use the same throughout. I don't have a strong opinion, though.

PR #752, opened 2024-02-28
FIX: Do not attempt to bootstrap from self

aakoshh:
See problem report on [Slack](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1709120541444859):

### Error
I'm trying to use kademlia for node discovery.
I set one node as a bootstrap node with
```
FM_RESOLVER__DISCOVERY__STATIC_ADDRESSES=""
```
And all other nodes with
```
FM_RESOLVER__DISCOVERY__STATIC_ADDRESSES="/dns4/fendermint-test-0.service.consul/tcp/27659/p2p/16Uiu2HAm9DLMnvQGuzzo2BTiF21nzAMtYQxXSmqev2zG6PCnsZwC"
```
First node fails to start with
```
2024-02-28T11:41:03.088692Z ERROR fendermint/app/src/main.rs:75: failed to execute Options { home_dir: "/local", log_dir: None, log_file_prefix: None, mode: "dev", log_level: Info, global: GlobalArgs { network: Testnet }, command: Run(RunArgs) }: error creating IPLD Resolver Service

Caused by:
    0: Error in the discovery configuration
    1: no bootstrap address
```

### Investigation

We add seeds nodes [here](https://github.com/consensus-shipyard/ipc/blob/3bcef484642775c3c259c603af534708ef3d329f/ipld/resolver/src/behaviour/discovery.rs#L137-L143) after parsing them [here](https://github.com/consensus-shipyard/ipc/blob/3bcef484642775c3c259c603af534708ef3d329f/ipld/resolver/src/behaviour/discovery.rs#L101-L110) so they shouldn't be empty addresses (this [test](https://github.com/consensus-shipyard/ipc/blob/3bcef484642775c3c259c603af534708ef3d329f/fendermint/app/settings/src/lib.rs#L411-L418) also ensures empty lists are handled). 

[Kademlia::add_address](https://github.com/libp2p/rust-libp2p/blob/1a9cf4f7760724032b729c43165716c7ecd842ad/protocols/kad/src/behaviour.rs#L519) can ignore an address if it's the _self address_, and then [Kademila::bootstrap](https://github.com/libp2p/rust-libp2p/blob/1a9cf4f7760724032b729c43165716c7ecd842ad/protocols/kad/src/behaviour.rs#L872) would indeed fail if it has no peers to bootstrap from. 

So if we configured a single static address for the node, and that address happened to be using its own peer ID, then we would see this error. Maybe all nodes get the same env var, or maybe they all share the same network key. I don't see how a truly empty list of static addresses could lead to this, though. 

### Fix

The PR skips static addresses that use the local peer ID, which should prevent bootstrapping if there are no other candidates.

### DIFF
--- a/ipld/resolver/src/behaviour/discovery.rs
+++ b/ipld/resolver/src/behaviour/discovery.rs
@@ -98,12 +98,18 @@ impl Behaviour {
             return Err(ConfigError::InvalidNetwork(nc.network_name));
         }
 
+        let local_peer_id = nc.local_peer_id();
+
         // Parse static addresses.
         let mut static_addresses = Vec::new();
         for multiaddr in dc.static_addresses {
             let mut addr = multiaddr.clone();
             if let Some(Protocol::P2p(peer_id)) = addr.pop() {
-                static_addresses.push((peer_id, addr))
+                // Trying to add the local peer to the routing table would be ignored,
+                // but if it's the only one we could falsely believe we can bootstrap.
+                if peer_id != local_peer_id {
+                    static_addresses.push((peer_id, addr));
+                }
             } else {
                 return Err(ConfigError::InvalidBootstrapAddress(multiaddr));
             }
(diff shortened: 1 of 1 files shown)


fridrik01 (approved):
Nice!

PR #757, opened 2024-02-29
Topdown lost commits

cryptoAtwill:
Recover lost fendermint topdown commits from previous archived repo. Some major changes includes:
- Remove sync pointers that simplifies the polling logic
- Integrate with `sync_many` flag from main

Back then the requirement was only pushing the parent view to cache if the next block is a non-null block. In this case, we need to track: 
1. The highest block pull (head)
2. The previous non-null block (tail)
We were using a head and tail to pull data.
Now we dont have this requirement, we just need the latest height fetched in cache to determine the next height to poll. The data is pushed to cache regardless if it's null or non-null. This way it simplifies the logic.


fendermint/vm/topdown/src/finality/null.rs
```diff
@@ -73,15 +73,8 @@ impl FinalityWithNull {
...
         if let Some((block_hash, validator_changes, top_down_msgs)) = maybe_payload {
-            emit!(
-                EventType::NewParentView,
-                is_null = false,
-                height,
-                block_hash = hex::encode(&block_hash)
-            );
             self.parent_block_filled(height, block_hash, validator_changes, top_down_msgs)
         } else {
-            emit!(EventType::NewParentView, is_null = true, height);
```
aakoshh, line 84: Why remove these now? They were actually useful because they show up in `docker logs`, while looking at the debug logs currently is a bit of a PITA because of all the extra logs from 3rd party libraries like tower ABCI.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -41,75 +38,50 @@ where
...
+        tracing::debug!(chain_head, latest_height_fetched, "syncing heights");
 
-        if self.detected_reorg_by_height(chain_head) {
+        if latest_height_fetched > chain_head {
             tracing::warn!(
-                pointers = self.sync_pointers.to_string(),
                 chain_head,
-                "reorg detected from height"
+                latest_height_fetched,
+                "chain head went backwards, potential detected from height"
```
aakoshh, line 74:
```suggestion
                "chain head went backwards, potential reorg detected from height"
```

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -143,16 +128,54 @@ where
...
-    ///
-    /// Returns the height which was polled.
-    ///
-    /// This method is only expected to be called after the caller has checked that the chain
-    /// is supposed to have data at the next height.
-    async fn poll_next(&mut self) -> Result<BlockHeight, Error> {
-        let height = self.sync_pointers.head() + 1;
-        let parent_block_hash = self.non_null_parent_hash().await;
+    /// Get the data needed to pull the next block
+    async fn get_from_cache(&self) -> (BlockHeight, BlockHash) {
```
aakoshh, line 132: Can we find a better name for this method? Something that mentions what it is getting?

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -143,16 +128,54 @@ where
...
+        // we are getting the latest height fetched in cache along with the first non null block
+        // that is stored in cache.
+        // we are doing two fetches in one `atomically` as if we get the data in two `atomically`,
+        // the cache might be updated in between the two calls. `atomically` should guarantee atomicity.
```
aakoshh, line 136: Just wondering: who else might be updating the cache?

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -164,9 +187,20 @@ where
...
-
-                    self.sync_pointers.advance_head();
+                    tracing::debug!(
+                        height,
+                        "detected null round at height, inserted None to cache"
+                    );
+
+                    atomically_or_err::<_, Error, _>(|| {
+                        self.provider.new_parent_view(height, None)?;
+                        emit!(EventType::NewParentView, is_null = true, height);
```
aakoshh, line 197: It's okay that we do debug logs inside `atomically`, but I would try not to `emit!` any events because these blocks can be retried on conflict, so any side effecting should be avoided. I think we should treat `emit!` a bit more seriously than throwaway debug logs, and try to make it consistent; emitting the same event twice in a row would be weird, and could throw off metrics too.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -184,48 +218,24 @@ where
             return Err(Error::ParentChainReorgDetected);
         }
 
-        if let Some((to_confirm_height, to_confirm_hash)) = self.sync_pointers.tail() {
-            tracing::debug!(
+        let data = self.fetch_data(height, block_hash_res.block_hash).await?;
+        atomically_or_err::<_, Error, _>(|| {
+            self.provider.new_parent_view(height, Some(data.clone()))?;
+            emit!(
```
aakoshh, line 224: This too should be outside the `atomically_or_err`.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -143,16 +128,54 @@ where
...
+    async fn get_from_cache(&self) -> (BlockHeight, BlockHash) {
+        // we are getting the latest height fetched in cache along with the first non null block
+        // that is stored in cache.
+        // we are doing two fetches in one `atomically` as if we get the data in two `atomically`,
+        // the cache might be updated in between the two calls. `atomically` should guarantee atomicity.
+        atomically(|| {
+            let latest_height = if let Some(h) = self.provider.latest_height()? {
+                h
+            } else {
+                unreachable!("guaranteed to have last committed finality, report bug please")
```
aakoshh, line 141:
```suggestion
                unreachable!("guaranteed to have latest height, report bug please")
```

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -184,48 +218,24 @@ where
...
+                block_hash = hex::encode(&data.0),
+                num_topdown_messages = data.2.len(),
+                num_validator_changes = data.1.len(),
             );
-
-            let data = self.fetch_data(to_confirm_height, to_confirm_hash).await?;
-
-            atomically_or_err::<_, Error, _>(|| {
-                // we only push the null block in cache when we confirmed a block so that in cache
-                // the latest height is always a confirmed non null block.
```
aakoshh, line 198: So this is no longer true.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -119,14 +91,27 @@ where
                 break;
             }
 
-            let synced_height = self.poll_next().await?;
+            let synced_height = match self
+                .poll_next(latest_height_fetched + 1, first_non_null_parent_hash)
```
aakoshh, line 95: Maybe we can massage the return value so that instead of returning the height, which we don't need now, since we know that it's going to be `latest_height_fetched + 1`, we return the _next parent hash_ to use, which will the the same as the input if we found a null round, or it's going to be the hash of the non-null block we fetched. With this value you can eliminate the call to `get_from_cache()` at the end of the loop.

aakoshh (commented):
Great to see the code being simplified. Could you provide some description of what exactly changed? What replaces the pointers, why were they needed before but not any more?

fendermint/vm/topdown/src/sync/syncer.rs
cryptoAtwill: when the finality is committed, some of the cache might be purged

fendermint/vm/topdown/src/finality/null.rs
cryptoAtwill: I shifted its location, so that more info will be emitted

fendermint/vm/topdown/src/sync/syncer.rs
cryptoAtwill: location shifted

fendermint/vm/topdown/src/sync/syncer.rs
cryptoAtwill: nope, we just need to insert whatever we pulled from parent into the cache

aakoshh (approved):
LGTM, thanks!

I couldn't explain why it should be faster, though :thinking:

PR #758, opened 2024-02-29
FIX: Add (optional) timeout to parent sync Ethereum API queries

aakoshh:
Related to https://github.com/consensus-shipyard/ipc/pull/755

The setting is added as optional so that existing deployments won't get an error if it's missing, and that I don't have to hardcode some default. Instead it's added in the infra scripts, so we can see if it has any effect.

It also fixes the default majority to be larger than 2/3, which is standard in Byzantine settings.


ipc/provider/src/manager/evm/manager.rs
```diff
@@ -67,7 +67,7 @@ const ETH_PROVIDER_POLLING_TIME: Duration = Duration::from_secs(1);
 const TRANSACTION_RECEIPT_RETRIES: usize = 200;
 
 /// The majority vote percentage for checkpoint submission when creating a subnet.
-const SUBNET_MAJORITY_PERCENTAGE: u8 = 60;
+const SUBNET_MAJORITY_PERCENTAGE: u8 = 67;
```
maciejwitowski, line 70: In contract [this default](https://github.com/consensus-shipyard/ipc/blob/main/contracts/test/invariants/handlers/SubnetRegistryHandler.sol#L24) is set as 70. Would it make sense to unify it one way or another so it's at least simpler to answer the question "what is the default?"

maciejwitowski (approved):

ipc/provider/src/manager/evm/manager.rs
aakoshh: From Byzantine literature it has to be `n * 2/3 + 1`, which if we go from 1 to 100 is 67. Not sure where 60 came from, or why it should be 70.

PR #763, opened 2024-02-29
ENG-770: Accept log filter expressions

aakoshh:
The PR adds support for more complex log filters to be passed and uses an [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html) instead of the simple log levels. 

## Original

Is it possible to configure log lever of individual components? I want info level overall but want to see debug metrics of kademlia to check if nodes are connected.
Like it is done in cometbft:
```
CMT_LOG_LEVEL  = "main:info,abci-client:debug,*:error"
```
Is this possible with fendermint?

## Changes

* The default env var we look for is now updated to `FM_LOG_LEVEL`, but `LOG_LEVEL` is still accepted as well. 
* This log level is now applied on the console log, that is, it's no longer restricted to `INFO` level
* Added a separate `FM_LOG_FILE_LEVEL` that governs the filter on the log file. If it's not present, we default to the level of the console. 
* Prints the console logs to `stderr` as it [used to be](https://github.com/consensus-shipyard/fendermint/pull/432/files#diff-8a4e452aee36822f580e7e6fbb305327636546a0f2302b116c3178ee71c0699cL20), so that we can print e.g. JSON in the CLI commands and they don't get mixed up with logs. 
* The log expressions can look like `info,tower_abci=warn,libp2p=warn` - see the docs of [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html). 

The `target` of the filter can be a module like `fendermint_vm_topdown::sync` or it should work just as well on a [prefix](https://github.com/tokio-rs/tracing/blob/0e3577f6f3995b92accee21e0737c25ef0f1953c/tracing-subscriber/src/filter/directive.rs#L247), e.g. just `fendermint` should cover all of our crates, even though there is no explicit `fendermint` module.

### DIFF
--- a/fendermint/app/options/src/lib.rs
+++ b/fendermint/app/options/src/lib.rs
@@ -118,12 +118,12 @@ pub struct Options {
 }
 
 impl Options {
-    /// Tracing filter for the console, unless it's turned off.
+    /// Tracing filter for the console.
     ///
     /// Coalescing everything into a filter instead of either a level or a filter
     /// because the `tracing_subscriber` setup methods like `with_filter` and `with_level`
     /// produce different static types and it's not obvious how to use them as alternatives.
-    pub fn log_console_filter(&self) -> anyhow::Result<Option<EnvFilter>> {
+    pub fn log_console_filter(&self) -> anyhow::Result<EnvFilter> {
         self._log_level
             .as_ref()
             .unwrap_or(&self.log_level)
@@ -131,7 +131,7 @@ impl Options {
     }
 
     /// Tracing filter for the log file.
-    pub fn log_file_filter(&self) -> anyhow::Result<Option<EnvFilter>> {
+    pub fn log_file_filter(&self) -> anyhow::Result<EnvFilter> {
         if let Some(ref level) = self.log_file_level {
             level.to_filter()
         } else {
@@ -202,21 +202,18 @@ mod tests {
             opts.log_console_filter().expect("filter should parse")
         };
 
-        assert!(
-            parse_filter("fendermint --log-level debug run").is_some(),
-            "standard log level"
-        );
-        assert!(
-            parse_filter("fendermint --log-level off run").is_none(),
-            "no logging"
-        );
-        assert!(
-            parse_filter("fendermint --log-level libp2p=warn,debug run").is_some(),
-            "complex filter"
-        );
+        let assert_level = |cmd: &str, level: LevelFilter| {
+            let filter = parse_filter(cmd);
+            assert_eq!(filter.max_level_hint(), Some(level))
+        };
 
-        let filter = parse_filter("fendermint --log-level info run").unwrap();
-        assert_eq!(filter.max_level_hint(), Some(LevelFilter::INFO));
+        assert_level("fendermint --log-level debug run", LevelFilter::DEBUG);
+        assert_level("fendermint --log-level off run", LevelFilter::OFF);
+        assert_level(
+            "fendermint --log-level libp2p=warn,error run",
+            LevelFilter::WARN,
+        );
+        assert_level("fendermint --log-level info run", LevelFilter::INFO);
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


fendermint/app/options/src/log.rs
```diff
@@ -0,0 +1,84 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use clap::{builder::PossibleValue, ValueEnum};
+
+use lazy_static::lazy_static;
+use tracing_subscriber::EnvFilter;
+
+/// Standard log levels, or something we can pass to <https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html>
+///
+/// To be fair we could pass of everything except "off" as a filter.
```
aakoshh, line 11: That's not true actually, `EnvFilter` uses a different type which handles "off" too.

aakoshh: But I thought I'd leave it like this also to try how to implement a `ValueEnum` manually.

fendermint/app/options/src/lib.rs
```diff
@@ -87,15 +80,30 @@ pub struct Options {
...
-        env = "LOG_LEVEL"
+        env = "FM_LOG_LEVEL",
+        help = "Standard log levels, or a comma separated list of filters, e.g. 'debug,tower_abci=warn,libp2p::gossipsub=info'",
+        value_parser = parse_log_level,
+    )]
+    log_level: LogLevel,
+
+    /// Fallback for the `log_level` with a legacy env var name.
+    #[arg(hide = true, value_enum, env = "LOG_LEVEL", value_parser = parse_log_level,)]
+    _log_level: Option<LogLevel>,
```
fridrik01, line 97: Do we need to keep legacy env around? Maybe we can just remove them now, otherwise they become much difficult to get rid off

fendermint/app/options/src/log.rs
```diff
@@ -0,0 +1,88 @@
...
+/// and perhaps we have simpler usecases where we only want to simply match levels.
+#[derive(Debug, Clone)]
+pub enum LogLevel {
+    Off,
+    Error,
+    Warn,
+    Info,
+    Debug,
+    Trace,
+    Filter(String),
```
fridrik01, line 22: oh that's nice way to implement this!

fridrik01 (approved):
nice!

fendermint/app/options/src/lib.rs
aakoshh: I don't know, I didn't want to throw a monkey wrench in Fluence's deployment scripts without giving some heads up. I figured it can't get more difficult, we just remove this line, it's not like any code depends on the field itself. It would be nice if there were env aliases supported by `clap`, but as far as I saw there can be only one name.

fendermint/app/src/main.rs
```diff
@@ -8,61 +8,64 @@ use tracing_appender::{
...
+
+            Some(file_layer)
+        }
+        _ => None,
     });
 
-    // we also log all traces with level INFO or higher to stdout
-    let registry = registry.with(
+    // we also log all traces to stderr (reserving stdout for any actual output such as from the CLI commands)
+    let registry = registry.with(console_filter.map(|filter| {
```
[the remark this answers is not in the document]
aakoshh: You are right, apparently you can! However I think that would not be what `None` was supposed to mean, which is basically `OFF`. I suspect if we call `.with_filter(None)` then the effect is like not calling the method at all, which means it will log something, whereas I wanted to say that if you configure `"off"` which is turned into `None` then there are no logs at all. 

This is where we could simplify the thing and just return a non-optional `EnvFilter`. I didn't know what was better: not configuring anything, or configuring something with and `off` filter. 

But if I'm honest we'll probably never have a case of not logging anything at all, so we might as well simplify what we can, since this part of the code is just more complicated to maintain for little benefit.

PR #766, opened 2024-03-04
Skip empty bottom up checkpoint

cryptoAtwill:
Initial implementation of skipping bottom up checkpoint. 

Currently, if the child is not active enough, there are not enough to fill each bottom up checkpoint and mostly, it's a large number of empty bottom up checkpoint with no cross message nor validator changes.

This has two draw backs:
1. Wasting gas as checkpoint signatures validaton is not cheap and cost grows with number of validators.
2. Slows down relayer catching up, doing useless work.

This PR materializes previous discuss by adding a previous checkpoint height to `BottomUpCheckpoint` struct so that bottom up checkpoints are chained but not required to be submitted at fixed interval.
```solidity
/// @notice A bottom-up checkpoint type.
struct BottomUpCheckpoint {
    /// @dev The height of the previous bottom up checkpoint
    uint256 prevHeight;
   
   // other fields not changed
   ... 
}
```

The bottom up checkpoint period is still kept for batch purposes. The new flow is as follows:
1. Caller sends a bottom up message to gateway
2. Gateway stores the bottom up message to the next bottom up epoch with message batch created/updated
(The above two steps are the same as before)
3. `fendermint` creates a checkpoint only when bottom up message batch is not empty or configuration number is updated (in follow up PR, not implemented yet as it does not impact contract).
4. Validators submit signatures
5. Relayer forward quorum to parent
6. Parent gateway checks if the `checkpoint.prevHeight == storage.lastCommittedHeight`


contracts/src/gateway/router/CheckpointingFacet.sol
```diff
@@ -52,9 +52,42 @@ contract CheckpointingFacet is GatewayActorModifiers {
...
+        // checkpoint.msgs.length != 0 || s.prevConfigNumber < checkpoint.nextConfigurationNumber
+        //
+        // so that no empty checkpoint will be created. The drawback is we might have long block intervals with
+        // no checkpoints at all. We can get around this with some `maxBUGap` or something along this line, so that
+        // when no bottom up checkpoints, we dont allow long gap where there is no bottom up checkpoints. This also
+        // has sequential submission restriction
+        //
+        // Or we dont enforce any check at all here, so that we let the child subnet decide whatever checkpoint
+        // the validator wants to create, but as long as quorum is reached and submitted to the parent, the parent
+        // will enforce checkpoints are sequential. This initial implementation took this approach.
```
cryptoAtwill, line 89: But the safest approach is enforcing:
```
checkpoint.prevHeight != s.prevBottomUpCheckpointHeight
```
If somehow the child decided to the `checkpoint.prevHeight` is not the same as that in the parent, then, the checkpointing is blocked.

contracts/src/gateway/router/CheckpointingFacet.sol
```diff
@@ -52,9 +52,42 @@ contract CheckpointingFacet is GatewayActorModifiers {
...
+        // }
+        //
+        // First of all, the check is buggy in the first place as it does not support
+        // a full message batch before a checkpoint period is reached.
+        // Then, `prevBottomUpCheckpointHeight` is introduced, as long as BU checkpoints
+        // can form a chain, it should be fine.
+        //
+        // Instead, one can add the following check:
+        //
+        // if (checkpoint.prevHeight != s.prevBottomUpCheckpointHeight) {
```
aakoshh, line 70:
```suggestion
        // if (checkpoint.prevHeight != s.lastBottomUpCheckpointHeight) {
```
I remember finding our wording confusing somewhere, I think it was _current_ vs _last_ validator set. 

Anyway, I think the `s` means the current state, and in that context it's "the last bottom-up checkpoint we received". If we use `prev` it's a bit less clear to me if there is something newer.

contracts/src/gateway/router/CheckpointingFacet.sol
```diff
@@ -52,9 +52,42 @@ contract CheckpointingFacet is GatewayActorModifiers {
...
+        //     revert InvalidCheckpointEpoch();
+        // }
+        //
+        // However, we are restricting to ONLY sequential creation
+        // of bottom up checkpoints. One can only submit the next one if the current
+        // checkpoint quorum is reached. This would slow down the creation process.
```
aakoshh, line 76: You are saying the relayers can only submit checkpoints with full quorum. I think that's reasonable, a checkpoint without a quorum cannot be executed, so what's the point of it? Either there is another checkpoint missing in front of it, or it's an equivocation. The relayer has the infromation to reconstruct the right sequence.

contracts/src/gateway/router/CheckpointingFacet.sol
```diff
@@ -52,9 +52,42 @@ contract CheckpointingFacet is GatewayActorModifiers {
...
+        //
+        // Or we can enforce:
+        //
+        // checkpoint.msgs.length != 0 || s.prevConfigNumber < checkpoint.nextConfigurationNumber
+        //
+        // so that no empty checkpoint will be created. The drawback is we might have long block intervals with
+        // no checkpoints at all. We can get around this with some `maxBUGap` or something along this line, so that
+        // when no bottom up checkpoints, we dont allow long gap where there is no bottom up checkpoints. This also
```
aakoshh, line 84: Yes I think the formula we discussed in the ticket you created included this as an option. 
Checkpointing does have value on its own, so I wouldn't come up with a rule that says no, we don't do any empty checkpoints, period. This is a good time to make it configurable.

Can you include a link to the ticket with the formula?

contracts/src/gateway/router/CheckpointingFacet.sol
aakoshh: Do you mean _reverting_ if that condition is true, or enforcing that it _is_ true? 

I don't understand the "But". You would the parent enforce sequential checkpoints without this check?

aakoshh: I am slightly confused by the comment, but to be clear I think sequential checkpoints should be enforced by the checks here, we should not admit out-of-order checkpoints.

aakoshh: Ah, okay, I see this method is running on the child subnet :upside_down_face:

contracts/src/gateway/router/CheckpointingFacet.sol
```diff
@@ -109,13 +142,19 @@ contract CheckpointingFacet is GatewayActorModifiers {
...
             height: height,
             membershipProof: membershipProof,
             weight: weight,
             signature: signature
         });
+
+        // if a new quorum is reached, the checkpoint should be forwarded to
+        // the parent and a new checkpoint with a bigger height should be created
+        if (isQuorumReached) {
+            s.prevBottomUpCheckpointHeight = height;
```
aakoshh, line 156: I don't think this is correct here. Checkpoints are created any time the conditions are right: 
* Too many messages in the buffer
* Some messages in the buffer have been waiting too long
* Too much time has passed without a chekpoint

Under any of these conditions the ledger creates a checkpoint. The next checkpoint should point at that one, the `lastbottomUpCheckpointHeight` should be updated as soon as the checkpoint is created. 

Then, validators start submitting signatures to the checkpoints, but there can be any number of them that need signing, we don't know which one reaches quorum first. They could go in opposite order of creation. This variable should not be updated just because some checkpoint reached a quorum. An event should be emitted, maybe the checkpoint flagged as completed, but that's it.

cryptoAtwill: Yeah, understand. Initlally I was wondering if there are chances of validator not agreeing on the quorum, so that only a quorum is reached, we can conclude that the checkpoint is created and set previous checkpoint height.

contracts/src/gateway/router/CheckpointingFacet.sol
cryptoAtwill: is there a formula? I was not aware of this.

contracts/src/gateway/router/CheckpointingFacet.sol
cryptoAtwill: Yeah, relayer should submit sequential. Also `fendermint` node will create bottom up checkpoint and immediately sets the `lastBottomUpCheckpointHeight` and this will allow future checkpoints to be `created`. Signature collection will be done later.

PR #767, opened 2024-03-04
Cache address-to-ActorType translation

mb1896:
This closes ENG-616

### DIFF
--- a/fendermint/eth/api/src/cache.rs
+++ b/fendermint/eth/api/src/cache.rs
@@ -143,11 +143,13 @@ mod tests {
     use fendermint_rpc::FendermintClient;
     use fvm_shared::address::Address;
     use std::str::FromStr;
-    use tendermint_rpc::{MockClient};
+    use tendermint_rpc::MockClient;
 
     #[test]
     fn test_read_and_write_addr_to_actor_type() {
-        let client = FendermintClient::new(MockClient::new(tendermint_rpc::MockRequestMethodMatcher::default()).0);
+        let client = FendermintClient::new(
+            MockClient::new(tendermint_rpc::MockRequestMethodMatcher::default()).0,
+        );
         let addr_cache = AddressCache::new(client, 1000);
 
         let address1 = Address::from_str("f410fivboj67m6ut4j6xx3lhc426io22r7l3h6yha5bi").unwrap();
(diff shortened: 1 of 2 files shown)


mb1896:
@aakoshh Please take a look when you get a chance!

fendermint/eth/api/src/cache.rs
```diff
@@ -107,4 +110,54 @@ where
...
+    use crate::state::ActorType;
+    use cid::Cid;
+    use fendermint_rpc::FendermintClient;
+    use fvm_shared::address::Address;
+    use std::str::FromStr;
+    use tendermint_rpc::HttpClient;
+
+    #[test]
+    fn test_read_and_write_addr_to_actor_type() {
+        let client = FendermintClient::new(HttpClient::new("http://localhost").unwrap());
```
aakoshh, line 137: There is a `MockClient` for them, e.g. [here](https://github.com/consensus-shipyard/ipc/blob/c3e4c020d52decfbc57eb6e2618c947f3f7b29a7/fendermint/vm/snapshot/src/manager.rs#L483).

fendermint/eth/api/src/state.rs
```diff
@@ -442,6 +443,12 @@ where
         height: FvmQueryHeight,
     ) -> JsonRpcResult<ActorType> {
         let addr = to_fvm_address(*address);
+
+        if let Some(actor_type) = self.addr_cache.get_actor_type_from_addr(&addr) {
+            log::debug!("cache contains the result: {:?}, return it", actor_type);
```
aakoshh, line 448:
```suggestion
            tracing::debug!(addr, actor_type, "cache hit");
```
I'm not sure if there's a difference between using `tracing::log::debug!` but usually we use `tracing::debug!`. The macro supports structured logging like `debug!(primitive_value, key2 %= value_with_display, ?value_with_debug, "static message")` which means those values will turn up in a way that is easier to filter for in log aggregators. We should prefer them to interpolating values into the message.

fendermint/eth/api/src/state.rs
```diff
@@ -457,6 +464,8 @@ where
             Some((typ, _)) => ActorType::Known(Cow::Owned(typ)),
             None => ActorType::Unknown(actor_type_cid),
         };
+        log::debug!("put result into cache: {:?} => {:?}", addr, ret);
```
aakoshh, line 467: Here too you can use structured logging.

aakoshh (commented):
Nice! Have you considered caching the CID-to-type relationship as well? 

It makes sense to store something on the address level because that's what we are asking about, and first we have to retrieve the state of the actor to get its code CID, however, there are only a handful of CIDs, so caching that second lookup could eliminate half of the queries on its own.

PR #772, opened 2024-03-04
FIX: Reject non-eth message.to if message.from is eth

aakoshh:
Fixes a vulnearability in the signature checking where `message.to` was converted to an Ethereum `H160` address type even if it was an f1 or f2 address. This would have allowed an attacker to alter the recipient address type of the FVM message and send funds to potentially irrecoverable addresses.

The reason it was like this was because originally the method was made for the purpose of displaying FVM transactions in the Ethereum API _somehow_, ie. even if the address was an f2 address, it was happy to show it so the transaction can be browsed.

### DIFF
--- a/fendermint/eth/api/src/conv/from_tm.rs
+++ b/fendermint/eth/api/src/conv/from_tm.rs
@@ -319,8 +319,8 @@ pub async fn to_eth_receipt(
         transaction_index,
         block_hash: Some(block_hash),
         block_number: Some(block_number),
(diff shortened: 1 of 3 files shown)


maciejwitowski (approved):

maciejwitowski:
This makes sense to me but would definitely be good to have more eyes on this.

jsoares (approved):

raulk (commented):
Review came in a little late, but this LGTM.

PR #776, opened 2024-03-07
[HOLD] Fix child subnet gateway lib staking not initialized

### LINKED ISSUE #775: Validator update in child not matched with that in the parent
### Issue type

Bug

### Have you reproduced the bug with the latest dev version?

Yes

### Version

latest

### Custom code

No

### OS platform and distribution

_No response_

### Describe the issue

The subnet gateway's `validatorTracker` is not initialized correctly to track that in the parent. This will lead to incorrect validator update in the child or even cannot be updated.

### Repro steps

1. After the subnet is created in the parent with `collateral` staking mode, i.e. say minimal collateral required is 0.01 fil.
2. The validator join with collateral `0.02`, the subnet will now be boostrapped.
3. The validator then increase its stake by `0.02` using `ipc-cli stake ...`.
4. `fendermint` will crash or invalidate membership information.

Membership after step 2 in fendermint prints:
```
current membership membership=Membership { validators: [Validator { weight: 20000000000000000, addr: 0x1a79385ead0e873fe0c441c034636d3edf7014cc, metadata: Bytes(0x047efe505fb55f56756514db73ff1e3a8d7fc08f7c5bbc3cbf10d646be71c2593766d6a8785f468ed6701c427d9b2a6a8d8a7d7146bc77a7e7a94c49bbcbd39f7f) }], configuration_number: 0 }
```
The weight is proportional to `0.02 fil` by some `scale_factor`(not critial in this issue), `configuration_number` is 0, which are correct. So the membership information is correct.

After `stake` is relayed to the child, it is mapped to `deposit` operation, after this operation is committed in the child, the membership now prints:
```
current membership membership=Membership { validators: [Validator { weight: 20000000000000000, addr: 0x1a79385ead0e873fe0c441c034636d3edf7014cc, metadata: Bytes(0x) }], configuration_number: 1 }
```
`configuration_number` increased to 1, which is correct due to `deposit` operation. However, `weight` is not changed, which it should be doubled, and metadata is set to zero address.

This is because in the child gateway, only membership information is initialised but not the `validatorTracker.validators` , see [line](https://github.com/consensus-shipyard/ipc/blob/adcbc084972d662dddfdb70229928233dac67f42/contracts/src/GatewayDiamond.sol#L77) . So no validator information are setup at all, they are all 0. 

So when the membership is updated collateral in [line](https://github.com/consensus-shipyard/ipc/blob/adcbc084972d662dddfdb70229928233dac67f42/contracts/src/gateway/router/TopDownFinalityFacet.sol#L52), the collateral is 0 + 0.02 instead of 0.02 + 0.02. For metadata, it's updated in [line](https://github.com/consensus-shipyard/ipc/blob/adcbc084972d662dddfdb70229928233dac67f42/contracts/src/gateway/router/TopDownFinalityFacet.sol#L61), which `info.metadata` is actually address(0).

When initial membership is not length 0, one should also setup the initial staking information for `validatorTracker.validators` so that the child is aligned with the parent.

### Relevant log output

_No response_

cryptoAtwill:
In the child gateway, the `validatorsTracker` is not properly initialised to track the parent, see #775 for details.

This PR adds:
1. A new `GenesisValidator` struct that tracks the `staking` information. Might be slightly verbose now, suggestions are welcomed.
2.  The initialisation of the `validatorTracker` based on the genesis validator strcuts
3. Add a getter function to check the child validator staking information so that one can compare with the parent.
4. Test case coverage.

Two follow up PRs:
- [x] #778 
- [x] #784 

Closes #775.


contracts/src/structs/Subnet.sol
```diff
@@ -143,6 +143,17 @@ struct IPCAddress {
     FvmAddress rawAddress;
 }
 
+/// @notice Validator struct for gateway constructor.
+struct GenesisValidator {
```
aakoshh, line 147: Wasn't there already a type for this? Isn't this what the constructor passes?

contracts/src/structs/Subnet.sol
```diff
@@ -143,6 +143,17 @@ struct IPCAddress {
...
+    /// The initial collateral deposited in the parent, 0 if none.
+    uint256 collateral;
```
aakoshh, line 149: I don't think there is a need for a separate `weight` and `collateral` field. At the moment `weight` _is_ the collateral, it's in tokens. The voting power is just a projection of it made in Rust.

contracts/src/structs/Subnet.sol
```diff
@@ -143,6 +143,17 @@ struct IPCAddress {
...
+    /// The initial federated power set by super admin in the parent, 0 if none.
+    uint256 federatedPower;
```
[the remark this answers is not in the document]
cryptoAtwill: Yeah, I will add them in a follow up.

contracts/src/structs/Subnet.sol
cryptoAtwill: oh, I was not aware of this, I thought it's another parameter updated in some way.

cryptoAtwill:
@aakoshh The `ferderatedPower` is part of the federated permission mode, it's also part of the `LibStaking` as it represents the `collateral` assigned or moderated from superadmin

aakoshh:
> The ferderatedPower is part of the federated permission mode, it's also part of the LibStaking as it represents the collateral assigned or moderated from superadmin

Thanks, I get that, I just don't understand why it needs to be a separate fields when we already have `weight`, which so far has been interpreted as collateral. Now it's not collateral, but it was never being paid out on the subnet anyway, so why do we need two fields?

cryptoAtwill:
I think currently for `federated power` permission mode, the weight of the validators will be 0 because weight is now just  related to collateral. But `federatedPower` should also be forwarded to `fendermint` as part of the weight. For my understanding, a validator's total power is the collateral staked + the federated power set by admin. It's just usually collateral and federated power are not all set to be greater than 0.

maciejwitowski:
> For my understanding, a validator's total power is the collateral staked + the federated power set by admin

@cryptoAtwill this isn't correct. Depending on the mode (collateral vs federated) the power is based on either collateral OR federated power, it's never a sum of those. Refer to [the original requirements](https://www.notion.so/pl-strflt/Federated-validation-6290874b58904818ac4fdddb5128d01b?pvs=4#b9aa0e930c714ae1baaa0c80c1b4888f)

cryptoAtwill:
@maciejwitowski Yeah, I understand, that's controlled by the Subnet Actor. But behind the scene, it's all controlled by LibStaking.

aakoshh (commented):
This looks correct to my untrained eye. 

My general comments about the federated power are [here](https://github.com/consensus-shipyard/ipc/pull/784#pullrequestreview-1930783460). The TL;DR is that I don't understand why this separation is necessary, other than the fact that `recordDeposit` sounds strange when we talk about the federated model, but that suggests to me that perhaps the `collateral` itself is a misnomer here, an artifact of using the same `LibStaking` on both the parent and the child, but not necessarily that we need to differentiate between various ways that `weight` can be assigned, but rather to generalise `collateral` into something that can increase and decrease, and is expressed in `atto`.

aakoshh:
FWIW when I had a similar dilemma in Rust I decided to make the [Validator](https://github.com/consensus-shipyard/ipc/blob/d28073b6d5d92871be48d44a5ab5c35abd5bbddb/fendermint/vm/genesis/src/lib.rs#L201) generic in how it's `power` is expressed. In the Genesis file it was turned into `Collateral` because the contract wanted the token amount, scaled to a `u64` value of type `Power` for CometBFT. 

Perhaps `LibStaking` could similarly use `power` (in `atto`) instead of `collateral` and then all would be well?

contracts/src/structs/Subnet.sol
raulk: @aakoshh this is the struct used to track validators pre-bootstrap. During this phase, we need to represent both collateral and explicit weight, if we ever want to support a hybrid model. If not, we could just represent a single weight here and have the relevant "power mutator" (stake, weight) update it.

contracts/src/structs/Subnet.sol
raulk: This is redundant with `ValidatorInfo`. It just contains the address, which we could extend the `ValidatorInfo` to contain too if needed.

PR #777, opened 2024-03-08
forge requires :ContractName to execute

snissn:
This change lets the forge scripts execute for me


fridrik01:
@snissn It would be good to know what errors this fixed for you :)

jsoares (approved):

raulk (changes requested):
This isn't true for me (both with an old version of Foundry and having just updated). What error are you seeing?

jsoares:
@raulk by "not true", do you mean you didn't have the issue before or that it doesn't work with the changes? I didn't have any issues before or after, fwiw 🤷 

Anyhow, if only @snissn observed this behaviour, then it's probably something to debug rather than brute force.

PR #778, opened 2024-03-08
Update genesis validator lifecycle

cryptoAtwill:
Fixing genesis validators not updated when `unstake` and `leave` are called.

Currently in `unstake`, when a validator unstakes all the collateral previuosly staked, then the validator address will not be removed from `genesisValidators`, see [line](https://github.com/consensus-shipyard/ipc/blob/9fe1ee01e85ef430b77aceccc90a934ec947edee/contracts/src/subnet/SubnetActorManagerFacet.sol#L200). The same happens to `leave`.

Note that this PR will fail the CI only when https://github.com/consensus-shipyard/ipc/pull/784 is merged into this branch.

### DIFF
--- a/contracts/src/lib/LibGenesis.sol
+++ b/contracts/src/lib/LibGenesis.sol
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity ^0.8.23;
+
+import {GenesisValidator, SubnetGenesis, ValidatorInfo, ValidatorSet} from "../structs/Subnet.sol";
+import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
+
+/// @notice Provides convinient util methods to handle genesis states of the subnet
+library LibGenesis {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    /// @notice Dumps the validator's genesis information
+    function getValidatorInfo(SubnetGenesis storage self, address validator) internal view returns(GenesisValidator memory info){
+        info = self.validatorInfo[validator];   
+    }
+
+    function addValidator(SubnetGenesis storage self, address validator) internal {
+        self.validators.add(validator);   
+    }
+
+    function removeValidator(SubnetGenesis storage self, address validator) internal {
+        self.validators.remove(validator);   
+    }
+
+    /// @notice Handles the genesis state when the subnet is bootstrapped. From this point onwards,
+    ///         no genesis state of the subnet can be changed.
+    /// @param validatorInfo The validator staking information from LibStaking
+    function bootstrap(SubnetGenesis storage self, ValidatorSet storage validatorInfo) internal {
+        finalizeValidatorInfo(self, validatorInfo);
+    }
+
+    // ============ Interal functions ==============
+
+    /// @notice Finalizes the genesis validator information as the subnet is bootstrapped. After 
+    ///         this point, the genesis validator info can no longer be changed.
+    /// @param validatorInfo The validator staking information from LibStaking
+    function finalizeValidatorInfo(SubnetGenesis storage self, ValidatorSet storage validatorInfo) internal {
+        address[] memory validators = self.validators.values();
+        
+        for (uint256 i = 0; i < validators.length; ) {
+            address addr = validators[i];
+
+            ValidatorInfo memory info = validatorInfo.validators[addr];
+            GenesisValidator memory genesis = GenesisValidator({
+                collateral: info.totalCollateral,
+                federatedPower: info.federatedPower,
+                addr: addr,
+                metadata: info.metadata
+            });
+
+            self.validatorInfo[addr] = genesis;
+
+            unchecked {
+                i++;
+            }
+        }
+    }
+}
--- a/contracts/src/lib/LibStaking.sol
+++ b/contracts/src/lib/LibStaking.sol
@@ -594,7 +594,7 @@ library LibValidatorTracking {
         // empty validator change logs
         self.changes.startConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
 
-        initValidators(self, validators);
+        return initValidators(self, validators);
     }
 
     function initValidators(
--- a/contracts/src/lib/LibStaking.sol
+++ b/contracts/src/lib/LibStaking.sol
@@ -9,6 +9,7 @@ import {LibStakingChangeLog} from "./LibStakingChangeLog.sol";
 import {PermissionMode, StakingReleaseQueue, StakingChangeLog, StakingChange, StakingChangeRequest, StakingOperation, StakingRelease, ValidatorSet, AddressStakingReleases, ParentValidatorsTracker, GenesisValidator, Validator} from "../structs/Subnet.sol";
 import {WithdrawExceedingCollateral, NotValidator, CannotConfirmFutureChanges, NoCollateralToWithdraw, AddressShouldBeValidator, InvalidConfigurationNumber} from "../errors/IPCErrors.sol";
 import {Address} from "openzeppelin-contracts/utils/Address.sol";
+import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
 
 library LibAddressStakingReleases {
     /// @notice Add new release to the storage. Caller makes sure the release.releasedAt is ordered
@@ -377,6 +378,7 @@ library LibStaking {
     using LibMaxPQ for MaxPQ;
     using LibMinPQ for MinPQ;
     using Address for address payable;
+    using EnumerableSet for EnumerableSet.AddressSet;
 
     uint64 internal constant INITIAL_CONFIGURATION_NUMBER = 1;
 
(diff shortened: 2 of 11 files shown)


contracts/src/lib/LibSubnetActor.sol
```diff
@@ -5,16 +5,18 @@ import {VALIDATOR_SECP256K1_PUBLIC_KEY_LENGTH} from "../constants/Constants.sol"
...
+import {LibGenesis} from "../lib/LibGenesis.sol";
 import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
 import {LibSubnetActorStorage, SubnetActorStorage} from "./LibSubnetActorStorage.sol";
 
 library LibSubnetActor {
     using EnumerableSet for EnumerableSet.AddressSet;
+    using LibGenesis for SubnetGenesis;
 
-    event SubnetBootstrapped(Validator[]);
+    event SubnetBootstrapped(address[]);
```
aakoshh, line 19: Why does this have to change?

contracts/src/lib/LibSubnetActor.sol
```diff
@@ -98,16 +101,16 @@ library LibSubnetActor {
...
+            s.genesis.addValidator(validators[i]);
 
             unchecked {
                 ++i;
             }
         }
 
         s.bootstrapped = true;
-        emit SubnetBootstrapped(s.genesisValidators);
+        s.genesis.bootstrap(s.validatorSet);
```
aakoshh, line 112: Should `bootstrapped` be moved into this new `SubnetGenesis` struct?

contracts/src/lib/LibSubnetActor.sol
aakoshh: I mean the subscriber would be interested in the actual powers, wouldn't it?

contracts/src/lib/LibSubnetActor.sol
```diff
@@ -98,16 +101,16 @@ library LibSubnetActor {
 
             LibStaking.setMetadataWithConfirm(validators[i], publicKeys[i]);
             LibStaking.setFederatedPowerWithConfirm(validators[i], powers[i]);
-
-            s.genesisValidators.push(Validator({addr: validators[i], weight: powers[i], metadata: publicKeys[i]}));
+            s.genesis.addValidator(validators[i]);
```
aakoshh, line 104: Is there are "remove" variant?

contracts/src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -55,8 +55,40 @@ contract SubnetActorGetterFacet {
...
+            return validators;
+        }
+
+        for (uint256 i = 0; i < total; ) {
+            address addr = s.genesis.validators.at(i);
+            ValidatorInfo memory info = getValidator(addr);
+            validators[i] = GenesisValidator({
+                // for genesis validators, totalCollateral == confirmedCollateral
+                collateral: info.totalCollateral,
+                federatedPower: info.federatedPower,
```
aakoshh, line 84: Can these be 0? I can't see where a validator is removed from `SubnetGenesis`, but I assume that they can leave before genesis and remove their collateral?

contracts/src/lib/LibSubnetActor.sol
aakoshh: Okay, found it at https://github.com/consensus-shipyard/ipc/pull/778/files#diff-db15c5a391bbc67defc15a7661eed42bf8488188847a6c72b0aa042e995c65baR206 

Not sure why it's in a different Solidity file, but it exists :+1:

contracts/src/lib/LibStaking.sol
```diff
@@ -611,15 +586,15 @@ library LibValidatorTracking {
         ParentValidatorsTracker storage self,
         uint16 activeValidatorsLimit,
         GenesisValidator[] memory validators
-    ) internal returns (Validator[] memory membership) {
```
aakoshh, line 614: Where was `membership` assigned before? :thinking:

contracts/src/lib/LibStaking.sol
```diff
@@ -611,15 +586,15 @@ library LibValidatorTracking {
...
+    ) internal returns (Validator[] memory) {
         self.validators.activeLimit = activeValidatorsLimit;
         // Start the next configuration number from 1, 0 is reserved for no change and the genesis membership
         self.changes.nextConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
         // The startConfiguration number is also 1 to match with nextConfigurationNumber, indicating we have
         // empty validator change logs
         self.changes.startConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
 
-        initValidators(self, validators);
+        return initValidators(self, validators);
```
aakoshh, line 597: Did Solidity not realise that there was nothing returned before :scream: ?

cryptoAtwill: yeah, that's the thing, solidity does not report an error, it's very easy to make a mistake, lol.

contracts/src/subnet/SubnetActorGetterFacet.sol
cryptoAtwill: yeah, when they `unstake` or `leave` with zero collateral, they are removed from genesis validator.

contracts/src/lib/LibSubnetActor.sol
cryptoAtwill: yeah, actually I think a lot of the parameters can be moved to `SubnetGenesis` struct so that we have better encapsulation, now it's scattered around. If this struct is ok, then I will create more PRs to migrate the fields one by one.

contracts/test/integration/SubnetActorDiamond.t.sol
```diff
@@ -90,6 +90,79 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
...
+
+        saDiamond.manager().unstake(validator1Stake / 2);
+        validators = saDiamond.getter().genesisValidators();
+        require(validators[0].addr == validator1);
+        require(validators[0].collateral == validator1Stake / 2, "collateral not match");
+
+        // unstake everything else
+        saDiamond.manager().unstake(validator1Stake / 2);
+        require(saDiamond.getter().genesisValidators().length == 0, "should not be validator");
+    }
```
aakoshh, line 164: Great! I couldn't write any of this, but it helps to understand the flow :pray:

contracts/src/lib/LibSubnetActor.sol
aakoshh: I'm not a contract designer, so I am possibly completely off the mark on this: I thought events should ideally be self-contained so you learn what you have to without the need for further queries, and if you _do_ need a query, then I'm not sure why it should return even the list of addresses, why not just a signal to flag up the block where the query can be made?

contracts/src/lib/LibStaking.sol
```diff
@@ -469,31 +470,6 @@ library LibStaking {
...
-            for (uint256 i; i < length; ) {
-                if (s.genesisValidators[i].addr == validator) {
-                    alreadyValidator = true;
-                    break;
-                }
-                unchecked {
-                    ++i;
-                }
-            }
-            if (!alreadyValidator) {
```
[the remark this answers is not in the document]
aakoshh: Good catch! 
Also :exploding_head:

aakoshh (approved):
This looks good to me, apart from my [general comments](https://github.com/consensus-shipyard/ipc/pull/784#pullrequestreview-1930783460) about `federated_power` itself.

contracts/src/lib/LibStaking.sol
raulk: What's worse is that none of the 100 linters, static analyzers, nor auditors caught this 💩

contracts/src/lib/LibGenesis.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    // ============ Interal functions ==============
+
+    /// @notice Finalizes the genesis validator information as the subnet is bootstrapped. After 
+    ///         this point, the genesis validator info can no longer be changed.
+    /// @param validatorInfo The validator staking information from LibStaking
+    function finalizeValidatorInfo(SubnetGenesis storage self, ValidatorSet storage validatorInfo) internal {
+        address[] memory validators = self.validators.values();
+        
+        for (uint256 i = 0; i < validators.length; ) {
```
raulk, line 39: Can we harden this by asserting that validatorInfo contains **_exactly_** the validators we expect (and hold in storage)?

contracts/src/lib/LibStaking.sol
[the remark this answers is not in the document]
raulk: Do we have a authoritative state diagram showing the lifecycle stages of a subnet, all transition edges, _and_ all possible actions within each stage?

contracts/src/lib/LibSubnetActor.sol
raulk: This event payload seems somewhat arbitrary. I wouldn't dwell too much on it right now, we'll need to change it going forward anyway to serve practical observability purposes. There are many ways to skin a cat, but subjective best practices:
* Keep events as small as they can be.
* Indexed fields: only publish identification data that subscribers will be interested to filter by through equality match (as that's the only predicate supported in Eth JSON-RPCs), e.g. publish addresses, IDs, kinds, categories, classes, etc. but do not publish balances, amounts, etc. as indexed fields.
* In the payload, only publish data that refers to the transaction itself, which would otherwise NOT be available in state (e.g. deltas, actioning address, etc.) Public state data can be fetched from the contract directly, so it's superfluous to publish here. (Draw parallels with event streaming principles)

contracts/src/structs/Subnet.sol
```diff
@@ -153,6 +154,17 @@ struct GenesisValidator {
...
 
+/// @notice Maintains the genesis information related to a particular subnet
+struct SubnetGenesis {
+    /// The genesis validators
+    EnumerableSet.AddressSet validators;
+    /// Tracks the validator info. This is only populated when the subnet is bootstrapped
+    mapping(address => GenesisValidator) validatorInfo;
+
+    /// TODO: migrating all the genesis related fields to this struct so that one can handle
+    ///       them all in a library.
```
raulk, line 165: Does it make sense to do this now, to minimise the migration logic we'll need to write to upgrade early adopters?

contracts/src/structs/Subnet.sol
```diff
@@ -153,6 +154,17 @@ struct GenesisValidator {
     bytes metadata;
 }
 
+/// @notice Maintains the genesis information related to a particular subnet
+struct SubnetGenesis {
```
raulk, line 158: Nice work -- this type makes a ton of sense.

contracts/test/integration/SubnetActorDiamond.t.sol
```diff
@@ -90,6 +90,79 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
         );
     }
 
+    /// @notice Testing the genesis validator's collateral
+    function testSubnetActorDiamond_GenesisValidators() public {
```
raulk, line 94: Great work writing these tests!

raulk (approved):

raulk:
Going to merge this because it's way easier to reason about the complete changeset all at once in the base branch.

PR #784, opened 2024-03-11
Update fendermint to match genesis changes in solidity

cryptoAtwill:
Update fendermint to align with the changes in solidity contracts. Some key changes are:
1. New type changes, i.e. adding `GenesisValidator` parameter that matches that in the contract. 
2. The validator's power in fendermint is the sum of the validator's collateral + federated power from the parent. See `LibStakingPower` for references.
3. Update `ipc` get genesis from parent to use `GenesisValidator`.
4. Update genesis CLI commands to reflect changes on the contract.

This PR is compilable and integration tested, but it's pending updates in unit tests.


fendermint/vm/genesis/src/lib.rs
```diff
@@ -95,6 +95,45 @@ pub struct Actor {
...
+    pub fn into_power(self: LibStakingPower, scale: PowerScale) -> Power {
+        let atto_per_power = Collateral::atto_per_power(scale);
+        let atto = self.federated_power.atto() + self.collateral.atto();
+        // Rounding away from zero, so with little collateral (e.g. in testing)
+        // we don't end up with everyone having 0 power and then being unable
+        // to produce a checkpoint because the threshold is 0.
+        let power = atto.div_ceil(&atto_per_power);
+        let power = power.min(BigInt::from(u64::MAX));
+        Power(power.try_into().expect("clipped to u64::MAX"))
+    }
```
aakoshh, line 125: There shouldn't be a need to copy-paste all this. The original is covered by tests, this isn't. 

Since both `federated_power` and `collateral` are `TokenAmount`, you can just add the together, wrap the result in `Collateral` and call `into_power` on it: `Collateral(self.federated_power + self.collateral).into_power(scale)`.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -126,6 +126,8 @@ impl<DB: Blockstore + Clone> GatewayCaller<DB> {
     ) -> anyhow::Result<()> {
         // Construct a Merkle tree from the power table, which we can use to validate validator set membership
         // when the signatures are submitted in transactions for accumulation.
+        tracing::debug!("power_table: {power_table:?}");
```
aakoshh, line 129:
```suggestion
        tracing::debug!(?power_table, "constructing Merkle tree of powers");
```

ipc/api/src/validator.rs
```diff
@@ -10,48 +10,90 @@ use crate::{
...
+//             Ok(Validator {
+//                 addr: ethers_address_to_fil_address(&validator.addr)?,
+//                 weight: eth_to_fil_amount(&validator.weight)?,
+//                 metadata: validator.metadata.to_vec(),
+//             })
+//         })
+//         .collect();
+//
+//     result
+// }
```
aakoshh, line 99: Why do we need this commented code?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -14,7 +14,7 @@ use ipc_actors_abis::{
     subnet_actor_reward_facet,
 };
 use ipc_api::evm::{fil_to_eth_amount, payload_to_evm_address, subnet_id_to_evm_addresses};
-use ipc_api::validator::from_contract_validators;
+use ipc_api::validator::vec_try_from;
```
aakoshh, line 17: Since this is a public and completely generic function, just a convenience for some library builtins, it would be worth moving it into a utility module. It's strange to see it being imported from `validator` when it has nothing validator specific about it.

aakoshh: I don't even see it used anywhere else, so maybe just move it into this module as a private function.

aakoshh (commented):
I can see Maciej commenting [here](https://github.com/consensus-shipyard/ipc/pull/776#issuecomment-1985893239) that it was never a requirement to support collateral and federation-assigned power at the same time. Your response, as I understand, is that this is implementation detail of `LibStaking`, ie. the lower level component does the book keeping this way, but this is not visible to the user (or the `SubnetActor` in your answer). 

I don't know the details of `LibStaking`, why it needs to separate the two concepts. Is it because if you have `collateral` then if you `leave` you get something back? But isn't that decided by the parent? Isn't `collateral` just a label in the subnet? Couldn't the parent just look at the permission mode and decide whether to release tokens to the validator during unbonding or not?

In any case, in contrast to what I thought was an implementation detail, here the distinction proliferates through the API and the CLI, and now the user can assign both at the same time. I'm not sure if this is desirable. Maybe we just shouldn't call it `collateral`, but keep it as a single field? 

My problem is that voting power can come from many places: it could be a consensus where PoW+PoS gives you power, or it could be coming from delegations. What we I think know is that for a subnet, it comes from the parent (and for rootnet, well, we don't do rootnets outside tests). I totally agree that `collateral` is misleading, but a proliferation of fields doesn't seem to be all that helpful to me.

On a different note, if we just wanted to make a distinction between tokens that can be returned to the users and other sources, we could perhaps call the fields `funded` and `unfunded`, with the latter being _not backed by locked tokens_.

ipc/api/src/validator.rs
cryptoAtwill: will remove them

raulk:
I agree with @aakoshh. We do not need to track both collateral and weight.

But the problem is deeper: these should NOT be used in conjunction. There should just be _weight_, which is either set explicitly (`federated`) or as a projection of collateral (`collateral`), either or, both not the sum of both. This confusion came up in [this thread](https://filecoinproject.slack.com/archives/C05J93G9G4F/p1710417471480019) and was corrected there, but the implementation hasn't been updated.

The "hybrid" model that uses both _both_ collateral and federated is very unsound. There is no unit enforced for manually-assigned weights. So if the user assigns equal weights of 1, 1, 1, an attacker could come in and deposit 10 FIL and take over the network in a clean sweep.

raulk (changes requested):
Let's unblock by doing the following.

1. Parent subnet actor needs to track collateral and static power for every validator, _before_ and _after_ genesis.
2. Remove the `GenesisValidator` type and simply use `ValidatorInfo` type; unless I'm missing something, these types are redundant. @aakoshh also pointed it out in his review.
3. That should allow us to get rid of weirdness like this:
    - [the transformation in `finalizeValidatorInfo`](https://github.com/consensus-shipyard/ipc/blob/97799c480a91b2eaf6d943b9b5d9d1893066f99f/contracts/src/lib/LibGenesis.sol#L36).
    - [this conditional in genesisValidator()](https://github.com/consensus-shipyard/ipc/blob/97799c480a91b2eaf6d943b9b5d9d1893066f99f/contracts/src/subnet/SubnetActorGetterFacet.sol#L63). 
5. The subnet actor translates the combination of validator collateral and static power into a weight, depending on the `permissionMode`.
    - If `collateral`, use only collateral.
    - If `federated`, use only the static power.
    - If in the future we want to add a hybrid, we'll add the hybrid. But neither collateral nor federated should be hybrid.
6. The validator set we pull in genesis and we subsequently send as configuration changes MUST NOT include collateral nor static power, just the weight, i.e. it should be this type: https://github.com/consensus-shipyard/ipc/blob/97799c480a91b2eaf6d943b9b5d9d1893066f99f/contracts/src/structs/Subnet.sol#L169.

raulk:
Actually, I just realised that the child gateway receives individual validator commands in the form of:

- set_federated_power(new_power)
- deposit(amount)
- withdraw(amount)
- ...

It also stores full [ValidatorInfo objects](https://github.com/consensus-shipyard/ipc/blob/97799c480a91b2eaf6d943b9b5d9d1893066f99f/contracts/src/structs/Subnet.sol#L87) including the power, collateral, etc.

The child gateway then needs to reconcile all this info and recalculate the power locally based on the `permissionMode`. I think all of this is unnecessarily complex.

Ideally all this acccounting would be encapsulated in the source of truth: subnet actor at the parent. That actor would just emit new weights (computed based on its permissionMode) that propagate down to the child gateway to apply them in the subnet.

Unfortunately all of this is going to require a more profound discussion/architectural change to adjust. Ideally we would've had this discussion _before_ implementing federated validation (at that time we just had collateral, which directly translated into weight), but that ship has sailed. At this stage, we might be better off just biting the bullet and refactoring later.

raulk:
So I think we're just going to have the bullet and accept this change until we rearchitect to decouple the collateral and static power tracking at the parent, from their translation to a mere weight at the child.

That said, we should implement (2), (3) and (4) from above before we merge this PR.

> 2. Remove the `GenesisValidator` type and simply use `ValidatorInfo` type; unless I'm missing something, these types are redundant. @aakoshh also pointed it out in his review.
> 3. That should allow us to get rid of weirdness like this:
>     - [the transformation in `finalizeValidatorInfo`](https://github.com/consensus-shipyard/ipc/blob/97799c480a91b2eaf6d943b9b5d9d1893066f99f/contracts/src/lib/LibGenesis.sol#L36).
>     - [this conditional in genesisValidator()](https://github.com/consensus-shipyard/ipc/blob/97799c480a91b2eaf6d943b9b5d9d1893066f99f/contracts/src/subnet/SubnetActorGetterFacet.sol#L63). 
> 4. The subnet actor translates the combination of validator collateral and static power into a weight, depending on the `permissionMode`.
>     - If `collateral`, use only collateral.
>     - If `federated`, use only the static power.
>     - If in the future we want to add a hybrid, we'll add the hybrid. But neither collateral nor federated should be hybrid.

raulk:
Merged in the interest of landing the fixes in the base branch onto main. Will discuss next steps with the team and @cryptoAtwill on Monday.

PR #785, opened 2024-03-11
Set subnet owner explicitly

cryptoAtwill:
Currently the ownership of the subnet is the `registry`, this PR explicitly sets the ownership as a parameter and controlled by the deployer. The deployer could set this to its own address or some other address instead of forcing it to be the `msg.sender`.


contracts/src/subnetregistry/RegisterSubnetFacet.sol
```diff
@@ -58,7 +58,7 @@ contract RegisterSubnetFacet is ReentrancyGuard {
         });
 
         // slither-disable-next-line reentrancy-benign
-        subnetAddr = address(new SubnetActorDiamond(diamondCut, _params));
+        subnetAddr = address(new SubnetActorDiamond(diamondCut, _params, msg.sender));
```
aakoshh, line 61: It looks like the registry will set it to the sender, which seems reasonable, however my impression was that the creator should be able to specify a multisig address if they wanted to, so the power over the federated weights aren't in one hand.

cryptoAtwill: Maybe registry should also accept it as a parameter?

aakoshh: I understand that requires modifying the Rust code as well, so I'll defer to others for what to do here.

contracts/src/SubnetActorDiamond.sol
```diff
@@ -38,7 +38,7 @@ contract SubnetActorDiamond {
         SubnetID parentId;
     }
 
-    constructor(IDiamond.FacetCut[] memory _diamondCut, ConstructorParams memory params) {
+    constructor(IDiamond.FacetCut[] memory _diamondCut, ConstructorParams memory params, address owner) {
```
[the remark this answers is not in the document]
aakoshh: Hm, but it seems to go against the grain of it, doesn't it? To have all parameters, plus this on the side?

raulk (approved):
This is fine, but it won't be straightforward to create the subnet with a multisig with our current tooling. We should implement `OwnershipFacet` so we can reuse our existing tooling (and submit the creation from an EOA), and _then_ transfer ownership to a multisig.

contracts/src/SubnetActorDiamond.sol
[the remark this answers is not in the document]
raulk: The owner indeed tends to be a "meta" parameter as it governs the contracts themselves, not the logic of the subnet. We should rename `ConstructorParams` to `SubnetConfig` (as it represents the subnet configuration and it's not the totality of the constructor params) to make it clearer down the line.

That said, I would move the owner to the 2nd place to have all contract metadata params leading, _then_ followed by the subnet config.

contracts/src/subnetregistry/RegisterSubnetFacet.sol
cryptoAtwill: I added a `Ownable` contract that handles transfer ownership.

contracts/test/unit/Ownable.t.sol
```diff
@@ -0,0 +1,35 @@
...
+
+        vm.prank(secondOwner);
+        vm.expectRevert(LibDiamond.NotOwner.selector);
+        o.transferOwnership(secondOwner);
+
+        // owner can perform transfer
+        vm.prank(firstOwner);
+        o.transferOwnership(secondOwner);
+        require(o.owner() == secondOwner, "second owner not correct");
+    }
```
raulk, line 34: Need zero address test, since that's a scenario we control for.

contracts/src/GatewayDiamond.sol
```diff
@@ -13,14 +13,15 @@ import {LibGateway} from "./lib/LibGateway.sol";
...
+import {Ownable} from "./Ownable.sol";
 
 error FunctionNotFound(bytes4 _functionSelector);
 
 bool constant FEATURE_MULTILEVEL_CROSSMSG = false;
 bool constant FEATURE_GENERAL_PUPRPOSE_CROSSMSG = true;
 uint8 constant FEATURE_SUBNET_DEPTH = 2;
 
-contract GatewayDiamond {
+contract GatewayDiamond is Ownable {
```
raulk, line 24: Don't we need to add the Ownable methods to the diamond function selectors somehow? Otherwise they might not be externally invokable?

raulk:
@cryptoAtwill why break the diamond abstraction model and not go with an [`OwnershipFacet`](https://github.com/mudgen/diamond-1-hardhat/blob/3da037b378195b4108b977cf2014600eeddbf67d/contracts/facets/OwnershipFacet.sol#L7) like I suggested above?

contracts/src/GatewayDiamond.sol
snissn: fyi here's how the diamond Upgrade determines if the user making the request is the "contract owner" https://github.com/consensus-shipyard/ipc/blob/b718ddb2dc8f8d849f61cb9992ac7b565e03082d/contracts/src/diamond/DiamondCutFacet.sol#L19 It is using the method LibDiamond.enforceIsContractOwner(); 

the upgrade facet is helpful but it only adds in two methods - one that allows for explicitly changing the ownership (seems helpful and worth adding!) and the other returns the owner. I don't think we want to add is Ownable to the contract here because Owning happens in LibDiamond

cryptoAtwill:
@raulk @snissn OwnableFacet works for me as well. Will migrate this.

PR #787, opened 2024-03-11
CLI: Add env var for config path

aakoshh:
> Also would be awesome so be able to specify ipc config path with env variable

Adds the `IPC_CLI_CONFIG_PATH` env var for the `config.toml` path and an `IPC_NETWORK` as a prefixed version of `NETWORK`. 

Fendermint has `FM_NETWORK` which I admit isn't ideal. I think people would frown on `FM_` prefixes in the `ipc-cli`, so this PR also changes `fendermint` to look for some environment variable aliases, preferring the `FM_` version but also accepting `IPC_` in this case, as well as the non-prefixed version. The same treatment is added to the `FM_LOG_LEVEL`, which allowed me to remove the `_log_level` field; it also looks for `RUST_LOG` now, which is the variable the `ipc-cli` uses (and might be tried based on intuition for those used to Rust libraries).

### DIFF
--- a/ipc/cli/src/lib.rs
+++ b/ipc/cli/src/lib.rs
@@ -34,13 +34,18 @@ pub trait CommandLineHandler {
 pub struct GlobalArguments {
     #[arg(
         long,
-        help = "The toml config file path for IPC Agent, default to ${HOME}/.ipc-agent/config.toml"
+        help = "The toml config file path for IPC Agent, default to ${HOME}/.ipc/config.toml",
+        env = "IPC_CLI_CONFIG_PATH"
     )]
     config_path: Option<String>,
 
     /// Set the FVM Address Network. It's value affects whether `f` (main) or `t` (test) prefixed addresses are accepted.
-    #[arg(long, default_value = "testnet", env = "NETWORK", value_parser = parse_network)]
-    pub network: Network,
+    #[arg(long, default_value = "testnet", env = "IPC_NETWORK", value_parser = parse_network)]
+    _network: Network,
+
+    /// Legacy env var for network
+    #[arg(hide = true, env = "NETWORK", value_parser = parse_network)]
+    __network: Option<Network>,
 }
 
 impl GlobalArguments {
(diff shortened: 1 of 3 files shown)


ipc/cli/src/lib.rs
```diff
@@ -34,13 +34,18 @@ pub trait CommandLineHandler {
...
+        help = "The toml config file path for IPC Agent, default to ${HOME}/.ipc/config.toml",
+        env = "IPC_CLI_CONFIG_PATH"
     )]
     config_path: Option<String>,
 
     /// Set the FVM Address Network. It's value affects whether `f` (main) or `t` (test) prefixed addresses are accepted.
-    #[arg(long, default_value = "testnet", env = "NETWORK", value_parser = parse_network)]
-    pub network: Network,
+    #[arg(long, default_value = "testnet", env = "IPC_NETWORK", value_parser = parse_network)]
+    _network: Network,
```
raulk, line 44: This is a good time for me to comment on this. Let's revisit the name `network` here. IUC:
* It only affects string parsing/formatting for Addresses.
* It refers to a network _kind/category/class_ (mainnet or testnet), not to a concrete _network_.
* We inherited the name as-is from fvm_shared, where it's poorly named at origin.
* `network` is an ridiculously overloaded term, and out of all the other usages, this is probably the least legitimate one, so I'd release the usage here so we can use it elsewhere (e.g. in the `--subnet` flag which should be aliased to `--network` because some commands can run against non-IPC subnets, e.g. ipc wallet balances).

raulk (approved):

ipc/cli/src/lib.rs
aakoshh: I totally agree, when I see `NETWORK` env var I have no idea what it might be for. The prefix helps with that. But I also didn't like the contract calling the subnet ID `network` here and there; I didn't understand why we need two names for the same thing, and thought "subnet" is good enough, no need to be ambiguous about it when "network" can be mixed with this. 

If you have a better name for it, we can change this of course. I didn't know if this was the right time to introduce breaking changes, so kept it backwards compatible.

PR #789, opened 2024-03-12
Workflow to auto deploy and test linked token contracts

mb1896:
Workflow link: https://github.com/consensus-shipyard/ipc/actions/workflows/deploy-linked-token-to-dedicated-host.yaml

Closes ENG-750


.github/workflows/deploy-linked-token-to-dedicated-host.yaml
```diff
@@ -0,0 +1,44 @@
+name: Deploy LinkedToken contracts to dedicated host running IPC subnet
+
+on:
+  workflow_dispatch:
+  pull_request:
```
mb1896, line 5: Will remove `pull_request:` before PR merges.

jsoares:
Closes ENG-750

I don't think this works in a comment, but I'll test anyway :)

raulk:
The LinkedToken is now an upgradable contract, so we'll have to check if these deployment scripts require change.

snissn:
@raulk I have been working on changes here to support the upgrade semantics - https://github.com/consensus-shipyard/ipc/pull/854 the process is a bit frustrating because of the slow timeline to run each test and I think the set up is flaky if multiple tests run concurrently

PR #791, opened 2024-03-13
FIX: Allow early checkpoint submission when msg batch is full

cryptoAtwill:
Currently when the msg batch is full, one cannot submit the bottom up checkpoint because the parent gateway is expecting submission at the checkpoint period. 

However, if a message batch is created that means the checkpoint epoch will not be at the checkpoint period, but slightly early, hence changing the `==` to `<=`.


raulk (commented):
@cryptoAtwill how does Fendermint behave when an earlier checkpoint is submitted to flush a message batch? Will it generate the next "idle" checkpoint exactly `bottomUpCheckPeriod` blocks after the last one, or will it stay on its "original" schedule?

For example, if `bottomUpCheckPeriod` is 10, and we've generated checkpoints 10, 20, 30, and then we generate a checkpoint at 35 to flush messages, will Fendermint generate the next checkpoint at 40 (original schedule), or at 45 (10 blocks after the last checkpoint)?

cryptoAtwill:
@raulk This is not controlled in fendermint, it's controlled in the [contract](https://github.com/consensus-shipyard/ipc/blob/c45a13c3d23788e312bca724f824a26779f15dec/contracts/src/lib/LibGateway.sol#L272). It will take all the messages at the checkpoint period and copy over to the current epoch. Say the next checkpoint period is 100, then we are currently at 99. Then the checkpoint at 100 is full, then all messages are copied to 99 and checkpoint 100 is cleared.

raulk:
@cryptoAtwill yes, I understand that. However, it's not clear to me if Fendermint will still produce the next (empty) checkpoint at 100 (original schedule), or at 99+10=109 (if `bottomUpCheckPeriod=10`). With this change, the subnet actor will accept both. Without this change, the subnet actor will only accept the latter IUC.

cryptoAtwill:
@raulk Fendermint will still create the checkpoint, see this [section](https://github.com/consensus-shipyard/ipc/blob/c45a13c3d23788e312bca724f824a26779f15dec/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L323). Basically `fenermint` checks both message batch and checkpoint period.

raulk:
@cryptoAtwill thanks for that pointer! From what I see [here](https://github.com/consensus-shipyard/ipc/blob/c45a13c3d23788e312bca724f824a26779f15dec/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L328), Fendermint will submit the next checkpoint at the original schedule (at 100).

So we will have:

- 99: checkpoint submitted due to message batch being full
- 100: empty checkpoint submitted on original schedule

raulk (commented):
This looks good, but two things:
- somehow this makes contract tests for federated validation fail
- have you confirmed that this works end-to-end live? (saturating a message batch, then checking we continue to produce checkpoints correctly)

raulk (commented):
We'll also want Solidity tests for this.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -32,7 +32,13 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
 
         bytes32 checkpointHash = keccak256(abi.encode(checkpoint));
 
-        if (checkpoint.blockHeight == s.lastBottomUpCheckpointHeight + s.bottomUpCheckPeriod) {
+        if (checkpoint.blockHeight <= s.lastBottomUpCheckpointHeight) {
```
cryptoAtwill, line 35: We need to add this to prevent `old` checkpoint to be resubmitted and overwrite `newer` checkpoints. This means past checkpoints cannot be submitted again.

contracts/test/IntegrationTestBase.sol
```diff
@@ -851,10 +851,8 @@ contract IntegrationTestBase is Test, TestParams, TestRegistry, TestSubnetActor,
             signatures[i] = abi.encodePacked(r, s, v);
         }
 
-        for (uint256 i = 0; i < n; i++) {
-            vm.prank(validators[i]);
-            saDiamond.checkpointer().submitCheckpoint(checkpoint, validators, signatures);
-        }
+        vm.prank(validators[0]);
```
cryptoAtwill, line 854: just like relayer, only need to submit once, not in a loop.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -32,7 +32,13 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
...
+            revert BottomUpCheckpointAlreadySubmitted();
+        }
+
+        // When a bottom up msg batch becomes full, a bottom up checkpoint will be
+        // created before the next checkpoint period is reached.
+        if (checkpoint.blockHeight <= s.lastBottomUpCheckpointHeight + s.bottomUpCheckPeriod) {
```
raulk, line 41: We should throw an error when the checkpoint's block height is higher than expected. Right now we will succeed silently. 

So this method should look like a traditional validate-then-execute pattern.
1. validate checkpoint is not below or equal to latest checkpointed height; if not, error.
2. validate checkpoint is within range; if not, error.

Then perform the logic outside of an if block.

contracts/test/IntegrationTestBase.sol
raulk: Note that this is a digression from the previous model. Previously we would attempt to reward all relayers that submitted a checkpoint, although the reward path was unimplemented, and the overall concept somewhat questionable.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -32,7 +32,13 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
```
raulk, line 32: We already have an `ensureValidCheckpoint` function that performs height validation based on the subnet actor state. We seem to be replicating some logic here, and in fact this logic is laxer than what we have there.

Please move all validation logic there and reconcile it.

raulk (changes requested):
The checkpoint validation logic is way too scattered, and we apply stricter validation first inside `ensureValidCheckpoint`, which means that the laxer validation in `submitCheckpoint` is partly not having an effect.

contracts/src/subnet/SubnetActorCheckpointingFacet.sol
raulk: This has been done in the follow-up PR!

raulk (approved):

PR #793, opened 2024-03-13
ENG-771: Add upgrade-info json file to determine when upgrades are applied + rpc + cli

fridrik01:
## Context
This PR updates fendermint to support future [ipcvisor](https://linear.app/interplanetary-consensus/issue/ENG-771/investigate-ipc-process-management-for-supporting-breaking-upgrades#comment-cfa0fc1d) tool we are building to make upgrades easier to orchistrate for our partners (see [post](https://filecoinproject.slack.com/archives/C05UA7RALFP/p1708432761811339)).

In order to facilitate the orchistration, this PR updates fendermint to:
- Split the upgrade_scheduler into hardcoded `Upgrade` migrations and separate upgrade schedule (a list of `UpgradeInfo` structs) which maps block heights to the hardcoded upgrade migrations. 
- Stop processing when it reaches a block height which has an associated hardcoded `Upgrade` which is `required` it does not have the source code of.
- Have RPC methods to query the node state (i.e check if its frozen)
- CLI commands to manage the upgrade-info.json file and check node state

This PR introduces all that functionality which should be enough for the `ipcvisor` to do its thing.

Note, the following are not implemented in this PR:
- Add verification of how the app version progresses progresses (ie. should it be strictly incrementing, can it decrease, etc)
- Verify CometBFT version at startup

### Test new functionality, including that fendermint should freeze and that rpc/cli work as expected
By default when fendermint starts it will automatically create an empty upgrade-migration.json file:

```
# start fenderming
cargo run -p fendermint_app --release -- run
...
2024-03-13T13:54:55.515089Z  INFO fendermint/app/src/cmd/run.rs:118: loaded upgrade_schedule path="/Users/fridrik/.fendermint/upgrade_info.json" schedule=UpgradeSchedule { schedule: {} }
...
CTRL-C

# confirm the upgrade_info.json file was created
cat ~/.fendermint/upgrade_info.json
[]
```

Lets try adding a new upgrade info which we don't have the hardcoded Upgrade for, lets also make it `required` to make things interisting

```
# add a new required upgrade
cargo run -p fendermint_app --release -- upgrade --upgrade-file=/Users/fridrik/.fendermint/upgrade_info.json  add-upgrade --height=1000 --new-app-version=1 --cometbft-version=0.37 --required

# confirm that the upgrade info was added
cat /Users/fridrik/.fendermint/upgrade_info.json
[
  {
    "height": 1000,
    "new_app_version": 1,
    "cometbft_version": "0.37",
    "required": true
  }
]
```

Now observe that fenderming freezes at the specifiec upgrade height since it does not have the hardcoded upgrade:

```
cargo run -p fendermint_app --release -- run
...
2024-03-13T14:01:42.959472Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=998
2024-03-13T14:01:43.994470Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=999
2024-03-13T14:01:45.034063Z  INFO fendermint/vm/interpreter/src/fvm/exec.rs:223: upgrade scheduled app_version=0 new_app_version=1 height=1000 upgrade_info=UpgradeInfo { height: 1000, new_app_version: 1, cometbft_version: "0.37", required: true }
2024-03-13T14:01:45.034080Z  WARN fendermint/vm/interpreter/src/fvm/exec.rs:242: upgrade not found height=1000 upgrade_info=UpgradeInfo { height: 1000, new_app_version: 1, cometbft_version: "0.37", required: true }
2024-03-13T14:01:45.034084Z ERROR fendermint/vm/interpreter/src/fvm/exec.rs:254: node frozen, please restart with a newer version which has the required upgrade height=1000 upgrade_info=UpgradeInfo { height: 1000, new_app_version: 1, cometbft_version: "0.37", required: true }
```

Lets query the rpc and confirm its frozen:

```
cargo run -p fendermint_app --release -- rpc query node-state
{
  "response": {
    "height": "999",
    "value": {
      "frozen": true
    }
  }
}
```

### Verify that updating fendermint with new version can successfully proceed

In the previous test, we confirmed that fendermint stopped once it progressed to a block height which had a `required` upgrade associated that it didn't have.

Here we want to test that when this happens, that when we stop the running (frozen) fendermint with a newer version which has the upgrade, that it will successfully proceed from where the previous version stopped. 

To test this, I created a [test-required-upgrade](https://github.com/consensus-shipyard/ipc/compare/upgrades-and-scheduler...test-required-upgrade?expand=1) branch which employs a custom kernel with new syscall as well with adding support for upgrading to app version 1. 

```
# stop the previous frozen fendermint
CTRL+C

# start the fendermint from the test-required-upgrade branch:
cargo run -p fendermint_app --release -- run
...
2024-03-13T15:42:09.886900Z  INFO fendermint/vm/interpreter/src/fvm/exec.rs:223: upgrade scheduled app_version=0 new_app_version=1 height=1010 upgrade_info=UpgradeInfo { height: 1010, new_app_version: 1, cometbft_version: "0.37", required: true }
2024-03-13T15:42:09.886931Z  INFO fendermint/vm/interpreter/src/fvm/exec.rs:239: upgrade successful app_version=1
2024-03-13T15:42:09.886940Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=1010
2024-03-13T15:42:10.249421Z  INFO fendermint/app/src/app.rs:670: event=ProposalProcessed is_accepted=true height=1011 size=0 hash="7EDD19B7B473CA1B006B4D694BF3B5D7771702C72D8F7F3352AC9245A5D6FD4C" proposer="9CDE756FDEFC4FAC219F01844D80309D8D800C32"
2024-03-13T15:42:12.614231Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=1011
2024-03-13T15:42:12.642061Z  INFO fendermint/app/src/app.rs:670: event=ProposalProcessed is_accepted=true height=1012 size=0 hash="EF9CC72C12A5DC33430CCBF2F4D6CD37BE0A80B71C5B722EF94C6B6BBEA0873F" proposer="9CDE756FDEFC4FAC219F01844D80309D8D800C32"
2024-03-13T15:42:12.675509Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=1012
2024-03-13T15:42:13.683986Z  INFO fendermint/app/src/app.rs:670: event=ProposalProcessed is_accepted=true height=1013 size=0 hash="A1D6DB8BF88F616426783667CEA3415D12F42875340F310F99E41391FE3920C7" proposer="9CDE756FDEFC4FAC219F01844D80309D8D800C32"
2024-03-13T15:42:13.710567Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=1013
2024-03-13T15:42:14.712736Z  INFO fendermint/app/src/app.rs:670: event=ProposalProcessed is_accepted=true height=1014 size=0 hash="CEC377F99B00787E909939D3C845DBE68463FE491834F4DB215D0DF0FF817AB5" proposer="9CDE756FDEFC4FAC219F01844D80309D8D800C32"
2024-03-13T15:42:14.736308Z  INFO fendermint/app/src/app.rs:771: event=NewBlock height=1014
...
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/793)

### DIFF
--- a/fendermint/app/options/src/rpc.rs
+++ b/fendermint/app/options/src/rpc.rs
@@ -94,6 +94,8 @@ pub enum RpcQueryCommands {
     StateParams,
     /// Query the current upgrade schedule
     UpgradeSchedule,
+    /// Get the current node state
+    NodeState,
 }
 
 #[derive(Subcommand, Debug, Clone)]
--- a/fendermint/app/options/src/rpc.rs
+++ b/fendermint/app/options/src/rpc.rs
@@ -92,6 +92,8 @@ pub enum RpcQueryCommands {
     },
     /// Get the slowly changing state parameters.
     StateParams,
+    /// Query the current upgrade schedule
+    UpgradeSchedule,
 }
 
 #[derive(Subcommand, Debug, Clone)]
--- a/fendermint/app/options/src/upgrade.rs
+++ b/fendermint/app/options/src/upgrade.rs
@@ -1,0 +1,43 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::path::PathBuf;
+
+use clap::{Args, Subcommand};
+
+#[derive(Args, Debug)]
+pub struct UpgradeArgs {
+    /// Path to the upgrade_info JSON file.
+    #[arg(long)]
+    pub upgrade_file: PathBuf,
+
+    #[command(subcommand)]
+    pub command: UpgradeCommands,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum UpgradeCommands {
+    AddUpgrade(AddUpgrade),
+}
+
+#[derive(Args, Debug)]
+pub struct AddUpgrade {
+    /// the height at which to schedule the upgrade
+    #[arg(long)]
+    pub height: u64,
+
+    /// the application version the upgrade will update to
+    #[arg(long)]
+    pub new_app_version: u64,
+
+    /// the required cometbft version for the upgrade
+    #[arg(long)]
+    pub cometbft_version: String,
+
+    /// whether this is a required upgrade or not. A required upgrade
+    /// will cause the node to freeze and not process any more blocks
+    /// if it does not have the corresponding Upgrade defined to
+    /// migrate to that version
+    #[arg(long)]
+    pub required: bool,
+}
--- a/fendermint/app/settings/src/lib.rs
+++ b/fendermint/app/settings/src/lib.rs
@@ -206,6 +206,8 @@ pub struct Settings {
     builtin_actors_bundle: PathBuf,
     /// Custom actors CAR file.
     custom_actors_bundle: PathBuf,
+    /// upgrade_info JSON file.
+    pub upgrade_info: PathBuf,
 
     /// Where to reach CometBFT for queries or broadcasting transactions.
     tendermint_rpc_url: Url,
@@ -229,7 +231,8 @@ impl Settings {
         snapshots_dir,
         contracts_dir,
         builtin_actors_bundle,
-        custom_actors_bundle
+        custom_actors_bundle,
+        upgrade_info
     );
 
     /// Load the default configuration from a directory,
--- a/fendermint/app/settings/src/lib.rs
+++ b/fendermint/app/settings/src/lib.rs
@@ -207,7 +207,7 @@ pub struct Settings {
     /// Custom actors CAR file.
     custom_actors_bundle: PathBuf,
     /// upgrade_info JSON file.
-    pub upgrade_info: PathBuf,
+    upgrade_info: PathBuf,
 
     /// Where to reach CometBFT for queries or broadcasting transactions.
     tendermint_rpc_url: Url,
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -109,10 +109,10 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
 
     let upgrade_schedule = if settings.upgrade_info().exists() {
         // load existing upgrade schedule
-        UpgradeSchedule::from_file(&settings.upgrade_info())?
+        UpgradeSchedule::from_file(settings.upgrade_info())?
     } else {
         // create an empty upgrade schedule
-        UpgradeSchedule::new().to_file(&settings.upgrade_info())?;
+        UpgradeSchedule::new().to_file(settings.upgrade_info())?;
         UpgradeSchedule::new()
     };
     tracing::info!(
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -107,14 +107,7 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         ValidatorContext::new(sk, broadcaster)
     });
 
-    let upgrade_schedule = if settings.upgrade_info().exists() {
-        // load existing upgrade schedule
-        UpgradeSchedule::from_file(settings.upgrade_info())?
-    } else {
-        // create an empty upgrade schedule
-        UpgradeSchedule::new().to_file(settings.upgrade_info())?;
-        UpgradeSchedule::new()
-    };
+    let upgrade_schedule = UpgradeSchedule::get_or_create(settings.upgrade_info())?;
     tracing::info!(
         path = settings.upgrade_info().to_string_lossy().into_owned(),
         schedule = ?upgrade_schedule,
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -11,7 +11,7 @@ use fendermint_crypto::SecretKey;
 use fendermint_rocksdb::{blockstore::NamespaceBlockstore, namespaces, RocksDb, RocksDbConfig};
 use fendermint_vm_actor_interface::eam::EthAddress;
 use fendermint_vm_interpreter::chain::ChainEnv;
-use fendermint_vm_interpreter::fvm::upgrades::UpgradeScheduler;
+use fendermint_vm_interpreter::fvm::upgrades::{UpgradeSchedule, Upgrades};
 use fendermint_vm_interpreter::{
     bytes::{BytesMessageInterpreter, ProposalPrepareMode},
     chain::{ChainMessageInterpreter, CheckpointPool},
@@ -114,7 +114,8 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         settings.fvm.gas_overestimation_rate,
         settings.fvm.gas_search_step,
         settings.fvm.exec_in_check,
-        UpgradeScheduler::new(),
+        UpgradeSchedule::new(),
+        Upgrades::new(),
     );
     let interpreter = SignedMessageInterpreter::new(interpreter);
     let interpreter = ChainMessageInterpreter::<_, NamespaceBlockstore>::new(interpreter);
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -107,6 +107,20 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         ValidatorContext::new(sk, broadcaster)
     });
 
+    let upgrade_schedule = if settings.upgrade_info().exists() {
+        // load existing upgrade schedule
+        UpgradeSchedule::from_file(&settings.upgrade_info())?
+    } else {
+        // create an empty upgrade schedule
+        UpgradeSchedule::new().to_file(&settings.upgrade_info())?;
+        UpgradeSchedule::new()
+    };
+    tracing::info!(
+        path = settings.upgrade_info().to_string_lossy().into_owned(),
+        schedule = ?upgrade_schedule,
+        "loaded upgrade_schedule"
+    );
+
     let interpreter = FvmMessageInterpreter::<NamespaceBlockstore, _>::new(
         tendermint_client.clone(),
         validator_ctx,
@@ -114,7 +128,7 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         settings.fvm.gas_overestimation_rate,
         settings.fvm.gas_search_step,
         settings.fvm.exec_in_check,
-        UpgradeSchedule::new(),
+        upgrade_schedule,
         Upgrades::new(),
     );
     let interpreter = SignedMessageInterpreter::new(interpreter);
--- a/fendermint/app/src/cmd/upgrade.rs
+++ b/fendermint/app/src/cmd/upgrade.rs
@@ -30,7 +30,7 @@ cmd! {
             self.required,
         ))?;
 
-        us.to_file(upgrade_file)?;
+        us.export_file(upgrade_file)?;
 
         Ok(())
   }
--- a/fendermint/app/src/cmd/upgrade.rs
+++ b/fendermint/app/src/cmd/upgrade.rs
@@ -22,15 +22,7 @@ cmd! {
 
 cmd! {
     AddUpgrade(self, upgrade_file: PathBuf) {
-        let mut us = if upgrade_file.exists() {
-            // load existing upgrade schedule
-            UpgradeSchedule::from_file(&upgrade_file)?
-        } else {
-            // create an empty upgrade schedule
-            UpgradeSchedule::new().to_file(&upgrade_file)?;
-            UpgradeSchedule::new()
-        };
-
+        let mut us = UpgradeSchedule::get_or_create(&upgrade_file)?;
         us.add(UpgradeInfo::new(
             self.height.try_into().unwrap(),
             self.new_app_version,
--- a/fendermint/app/src/cmd/upgrade.rs
+++ b/fendermint/app/src/cmd/upgrade.rs
@@ -1,0 +1,45 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::path::PathBuf;
+
+use fendermint_app_options::upgrade::AddUpgrade;
+use fendermint_app_options::upgrade::UpgradeArgs;
+use fendermint_app_options::upgrade::UpgradeCommands;
+use fendermint_vm_interpreter::fvm::upgrades::UpgradeSchedule;
+use fendermint_vm_message::ipc::UpgradeInfo;
+
+use crate::cmd;
+
+cmd! {
+    UpgradeArgs(self) {
+        let upgrade_file = self.upgrade_file.clone();
+        match &self.command {
+            UpgradeCommands::AddUpgrade(args) => args.exec(upgrade_file).await,
+        }
+    }
+}
+
+cmd! {
+    AddUpgrade(self, upgrade_file: PathBuf) {
+        let mut us = if upgrade_file.exists() {
+            // load existing upgrade schedule
+            UpgradeSchedule::from_file(&upgrade_file)?
+        } else {
+            // create an empty upgrade schedule
+            UpgradeSchedule::new().to_file(&upgrade_file)?;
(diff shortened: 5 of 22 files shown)


aakoshh:
> Stop processing when it reaches a block height which has an associated hardcoded Upgrade

Haven't seen it yet, but I expected that it will be looking for heights that it does _not_ have hardcoded, heights looking from the upgrade JSON file that it _learned_ about, but doesn't know how to execute.

fendermint/app/options/src/upgrade.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// the required cometbft version for the upgrade
+    #[arg(long)]
+    pub cometbft_version: String,
+
+    /// whether this is a required upgrade or not. A required upgrade
+    /// will cause the node to freeze and not process any more blocks
+    /// if it does not have the corresponding Upgrade defined to
+    /// migrate to that version
+    #[arg(long)]
+    pub required: bool,
```
aakoshh, line 42: Just thinking when an upgrade wouldn't be required. I know we talked about upgrades where you can download the new software and replace the existing one with it _now_, because the binary is backwards compatible, and it will apply the upgrade when the time comes. Why the upgrade list would know this though..

OTOH if some nodes update the `app_version` then everyone must do it, because it's part of the ledger. So an optional upgrade must not update the app version.

aakoshh: I'm also not sure how an optional upgrade would be handled if the node with the upgrade is started later than the height. Nothing will trigger that upgrade, so we don't know if it's been applied or not?

fendermint/app/settings/src/lib.rs
```diff
@@ -206,6 +206,8 @@ pub struct Settings {
     builtin_actors_bundle: PathBuf,
     /// Custom actors CAR file.
     custom_actors_bundle: PathBuf,
+    /// upgrade_info JSON file.
+    pub upgrade_info: PathBuf,
```
aakoshh, line 210:
```suggestion
    /// upgrade_info JSON file.
    upgrade_info: PathBuf,
```

Don't make it public, because if it's `home_relative!` then people should access it through the `settings.upgrade_info()` method, and that should be the only way, to avoid confusion.

fendermint/app/options/src/rpc.rs
```diff
@@ -92,6 +92,10 @@ pub enum RpcQueryCommands {
     },
     /// Get the slowly changing state parameters.
     StateParams,
+    /// Query the current upgrade schedule
+    UpgradeSchedule,
```
aakoshh, line 96: I am not exactly sure what an upgrade schedule is; maybe the docstring could explain a bit more. 

For example I understand it can be the contents of the JSON file, but I was wondering if it could also include the versions that the node is capable of executing.

fendermint/app/src/cmd/run.rs
```diff
@@ -107,14 +107,29 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         ValidatorContext::new(sk, broadcaster)
     });
 
+    let upgrade_schedule = if settings.upgrade_info().exists() {
+        // load existing upgrade schedule
+        UpgradeSchedule::from_file(settings.upgrade_info())?
+    } else {
+        // create an empty upgrade schedule
+        UpgradeSchedule::new().to_file(settings.upgrade_info())?;
+        UpgradeSchedule::new()
```
aakoshh, line 116: It looks odd to have to create two instances :monocle_face:

fendermint/app/src/cmd/upgrade.rs
```diff
@@ -0,0 +1,45 @@
...
+cmd! {
+    AddUpgrade(self, upgrade_file: PathBuf) {
+        let mut us = if upgrade_file.exists() {
+            // load existing upgrade schedule
+            UpgradeSchedule::from_file(&upgrade_file)?
+        } else {
+            // create an empty upgrade schedule
+            UpgradeSchedule::new().to_file(&upgrade_file)?;
+            UpgradeSchedule::new()
+        };
```
aakoshh, line 32: Maybe a `UpgradeSchedule::load_or_init` or `get_or_create` could be used to encapsulate this? It appears in the same way further up.

fendermint/app/src/cmd/upgrade.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        us.add(UpgradeInfo::new(
+            self.height.try_into().unwrap(),
+            self.new_app_version,
+            self.cometbft_version.clone(),
+            self.required,
+        ))?;
+
+        us.to_file(upgrade_file)?;
```
aakoshh, line 41: Usually the `to_this` and `into_that` methods imply conversions that take a reference or self. We're not converting the upgrade to a `File`, so maybe something like `write` or `export` could be used? Although we use `from_file` in the exact same way you do, so I'm not sure why this seems unusal.

fendermint/testing/contract-test/tests/run_upgrades.rs
```diff
@@ -235,7 +237,7 @@ async fn test_applying_upgrades() {
         tester.commit().await.unwrap();
 
         // check that the app_version was upgraded to 1
```
aakoshh, line 239: This comment is outdated.

fendermint/testing/contract-test/tests/run_upgrades.rs
```diff
@@ -235,7 +237,7 @@ async fn test_applying_upgrades() {
...
-        assert_eq!(tester.state_params().app_version, 1);
+        assert_eq!(tester.state_params().app_version, block_height as u64 + 10);
```
aakoshh, line 240: I think `block_height + 10` as `app_version` is puzzling. I kept wondering why that should be true, and it's true in this particular set of hardcoded values, but it looks like some kind of rule that one has to make sense of. Perhaps it would be more straight forward to just iterate over the known list of (height, version) pairs and assert them, rather than have a formula for it that works now, but it's not clear where the upgrades are defined that this offset _must_ be adhered to.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -19,6 +23,9 @@ use super::{
     FvmMessage, FvmMessageInterpreter,
 };
 
+/// Indicates whether the node has been frozen due to not having a required upgrade.
+pub static IS_FROZEN: AtomicBool = AtomicBool::new(false);
```
aakoshh, line 27: Can this live inside the `FvmMessageInterpreter` instead of being a public static global field? The way it is, it can be set from anywhere, by anyone. There has been no rule so far that there can be only one interpreter instance. The application only creates one, so there is no problem with multiple conflicting values.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,21 +219,47 @@ where
...
+                app_version = state.app_version(),
+                new_app_version = upgrade_info.new_app_version,
+                height = state.block_height(),
+                upgrade_info = ?upgrade_info,
+                "upgrade scheduled"
+            );
+
+            match self.upgrades.get(upgrade_info.new_app_version) {
+                Some(upgrade) => {
+                    let new_app_version = upgrade.execute(&mut state).context("upgrade failed")?;
```
aakoshh, line 233: It's not clear why we make the app version computed, when we said the upgrade will set it to the new version. Maybe `execute` can return nothing for now?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,21 +219,47 @@ where
...
+                        height = state.block_height(),
+                        upgrade_info = ?upgrade_info,
+                        "upgrade not found"
+                    );
+
+                    if upgrade_info.required {
+                        // mark the node as frozen
+                        IS_FROZEN.store(true, Ordering::Relaxed);
+
+                        // sleep forever, we can't proceed any further using our current fendermint version
```
aakoshh, line 252: Maybe this can be in `begin` instead of `end` to minimise work; like if we know we won't be able to finish the block, why even start executing it?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,21 +219,47 @@ where
...
+                        loop {
+                            tracing::error!(
+                                height = state.block_height(),
+                                upgrade_info = ?upgrade_info,
+                                "node frozen, please restart with a newer version which has the required upgrade"
+                            );
+                            sleep(std::time::Duration::from_secs(60));
```
aakoshh, line 259: I reckon this should be using `tokio::time::sleep(...).await;` since this is an async method. We shouldn't block a tokio thread if we don't have to.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -212,21 +219,47 @@ where
...
-
-            // there is an upgrade scheduled for this height, lets run the migration
-            let res = upgrade.execute(&mut state).context("upgrade failed")?;
-            if let Some(new_app_version) = res {
-                state.update_app_version(|app_version| {
-                    *app_version = new_app_version;
-                });
-
-                tracing::info!(app_version = state.app_version(), "upgraded app version");
+        if let Some(upgrade_info) = self.upgrade_schedule.get(state.block_height()) {
```
aakoshh, line 222: I was thinking whether we should also add this check to the `ProposalInterpreter` implementations (which this interpreter does not implement), but I think you are right, this should be sufficient. If a validator proposes a block without knowing whether it will be able to execute it, it can still be finalized and cause all nodes to freeze there during delivery. 

If we live with this model, then it's also probably right to keep the execution of the upgrade in `end`, because the decision about what transactions to put into the block has been made the the current version of the software... but no, that doesn't make any difference, if the new version will have to execute it, it will have to be able to handle all the transactions that the previous one already put into the block. 

So _that_ could be prevented by adding the freeze already to the proposal handling: if for example we wanted to get rid of a certain transaction type, we could just say "we're no longer eligible to create the next block", rather than put it in, then consider it invalid. 

If we want to handle such cases by marking these transactions as invalid, then I reckon the upgrade should be done in `begin` after all, which signifies: once this block begins, it's the new rules, even if the transactions are old. Versus suggesting that "until the end, it's the old rules", which could leave the new version in a conundrum whether to be backwards compatible just for the sake of a single block.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -155,6 +164,16 @@ where
                 let (state, ret) = state.builtin_actors().await?;
                 Ok((state, FvmQueryRet::BuiltinActors(ret)))
             }
+            FvmQuery::UpgradeSchedule => {
+                let ret = self.upgrade_schedule.get_all();
+                Ok((state, FvmQueryRet::UpgradeSchedule(ret)))
+            }
+            FvmQuery::NodeState => {
+                let ret = NodeState {
+                    frozen: IS_FROZEN.load(Ordering::Relaxed),
```
aakoshh, line 173: So this one is executed by the same interpreter, it can access it as a member variable.

aakoshh: Maybe it could tell on which upgrade it has been frozen, instead of just a bool.

fendermint/vm/interpreter/src/fvm/upgrades.rs
```diff
@@ -1,54 +1,94 @@
...
+
+        let mut us = UpgradeSchedule::new();
+        for upgrade_info in schedules {
+            us.add(upgrade_info)?;
+        }
+
+        Ok(us)
+    }
+
+    pub fn to_file(&self, path: impl AsRef<Path>) -> anyhow::Result<()> {
```
aakoshh, line 47: If `to_file` doesn't consume `self` then I really don't understand why there are two `new()` where it was created during startup.

fendermint/vm/interpreter/src/fvm/upgrades.rs
```diff
@@ -1,54 +1,94 @@
...
+use super::state::FvmExecState;
+use fvm_shared::clock::ChainEpoch;
 
-#[derive(PartialEq, Eq, Clone)]
-struct UpgradeKey(ChainID, BlockHeight);
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct UpgradeSchedule {
+    // schedule of upgrades to be executed at given height
+    schedule: BTreeMap<ChainEpoch, UpgradeInfo>,
+}
```
aakoshh, line 20: What I see missing here is a file system watcher. The JSON is only read when the node starts. We want it to stop when a new upgrade is announced, and the height arrives with the old version still running. If it doesn't refresh the file, then it won't know. 

I reckon it could keep a reference to the file and start a background loop that checks it every now and then, then loads new items into the `schedule`. 

This is also important because without a governance module, the upgrade schedule somehow has to make it to all full nodes, which is why I think the visor should be able to download it and overwrite the file on the disk. Not sure how to make sure that doesn't happen for the guy who is adding upgrades by hand.

fendermint/vm/interpreter/src/fvm/upgrades.rs
```diff
@@ -1,54 +1,94 @@
...
+    pub fn to_file(&self, path: impl AsRef<Path>) -> anyhow::Result<()> {
+        let json = serde_json::to_string_pretty(&self.get_all())
+            .context("failed to serialize upgrade_info")?;
+        std::fs::write(path, json).context("failed to write to upgrade_info file")?;
+
+        Ok(())
+    }
+
+    pub fn add(&mut self, upgrade_info: UpgradeInfo) -> anyhow::Result<()> {
+        match self.schedule.entry(upgrade_info.height) {
```
aakoshh, line 56: Maybe `add` should check not only that the height is unoccupied, but that the version fits between the previous and next upgrades. I'm not sure what kind of semantics we put on the app version, but since it identifies the upgrade, I reckon making it strictly monotonic is a reasonable default.

fendermint/vm/interpreter/src/fvm/upgrades.rs
```diff
@@ -1,54 +1,94 @@
...
+
+    pub fn get(&self, height: ChainEpoch) -> Option<&UpgradeInfo> {
+        self.schedule.get(&height)
+    }
+
+    pub fn get_all(&self) -> Vec<UpgradeInfo> {
+        self.schedule
+            .values()
+            .cloned()
+            .collect::<Vec<UpgradeInfo>>()
```
aakoshh, line 75: Maybe return them in reverse, so the top of the file shows useful upcoming info?

aakoshh (commented):
Great :clap:  

I added a lot of exploratory comments about design. Curious to hear others' opinions on the semantics of app version for example, as any upgrade that touches it is consensus breaking, and I find hard to reason about `required` because of this. 

What I'd be curious about as an operator is whether I an upgrade my node to any particular future version and expect it to work _before_ the upgrades are applied. For example maybe a migration changes the data schema of an IPC actor. Until it's executed, we cannot expect the new bindings to work, so we couldn't start a fendermint that has new bindings ahead of time. Maybe fendermint could have a _minimum application version_ that it checks at the start, and quits if it's not time for this particular binary to run yet.

fridrik01:
Closing while working on addressing remaining comments

jsoares:
As a general practice, I find it best to revert to draft rather than close. It keeps things observable while clearly signalling that it's not ready for review.

fridrik01:
> As a general practice, I find it best to revert to draft rather than close. It keeps things observable while clearly signalling that it's not ready for review.

I agree and this is the first time I have closed a PR instead of putting to draft. Reason was that I have on several occations had my PRs reviewed while they were in draft and I thought this time that might happen again as I had pushed several commits and didn't want someone to review it before I push the remaining commits :)

karlem:
@consensus-shipyard/ipc-core-devs  It looks like using a Visor might be a useful concept for future implementation of upgrades, but I would imagine something like the UpgradeScheduler would be a built-in on-chain actor. In that case, I think this draft can be safely closed for now.

PR #796, opened 2024-03-14
ENG-629: emit! macro with event structs

aakoshh:
This PR ended up as a slight refactoring of the `emit!` macro introduced [here](https://github.com/consensus-shipyard/ipc/commit/4d3f8901c8be2116cbab9c3416e1aba8bbf16724) 

Instead of the `EventType` being an enumeration of possible event labels, we introduce a flat `struct` for each of the events. They aren't in an enum because at the moment there is no need for them to be. The new `enum!` macro call looks like the construction of the struct (and it has to work as such!) but in reality it's a call to `tracing::event!`. The struct is only there for two reasons: 1) to make sure we are passing all the fields, with the right types and 2) the type name is used as `name` in the trading `Metadata` instead of `"event <file>:<line>"` which would be the default. Note that the filename and line are still available in other metadata fields. The `target` remained the same: the module name. So did the log level of INFO. 

This way we get a sort of schema we can look up for the events that get emitted, although we aren't really told up front what these events are going to be.

Given this, we'll have to create a `tracing::Subscriber` in the app to look for these events by name, and e.g. maintain Prometheus metrics. The subscriber will have to use the visitor pattern to look for fields such as `block_height` to increment gauges.

### Tried but failed

Originally I wanted to create a `Tracer<T>` type and pass around an instance of it, which would be called with  a `&T` _by a macro_ that would also call `tracing::Event::dispatch` after constructing the `Metadata` and `Callsite`. We would have defined an implementation with a `tracer!` macro that gets a list of fields, which can feed the metdata and the `ValueSet`. 

This way the only place where we'd define which fields we want to trace would have been the application root.

Unfortunately this didn't quite work. For performance reasons `tracing` never copies the data, and the `Metadata` is only created once as a `static` variable. For that it must know the list of field names statically, but right now it's not possible to have `const` functions on a trait, so the two could not be reconciled.


fendermint/app/src/events.rs
raulk, line 1: It'd be nicer to centralise all events in a single file/module, but I understand there may be type dependencies that make it cumbersome.

raulk (approved):
Nice work -- taking the simple road and thinking outside the box. When I looked at this, I tried to crack the `Valuable` mechanism, but this is simpler and it works.

fendermint/app/src/events.rs
aakoshh: Yeah, I can see the ups and downs with both. I thought it was nice to see what events are emitted by a crate inside the crate, and consumers who know about the crate will know about its events too. It is _also_ nice to be able to look up all events in a single place, rather than hunt them down. Perhaps the `fendermint_app` crate, if that's where we define our metrics, can re-export them, as a sort of Table-of-Contents. 

But I think centrally defining events and importing them into the emitter crate is a sort of anti-pattern. The producer knows what the events relevant to it are, the consumers can only pick what they want from that, not really the other way around, although of course there is a conversion, if the producer is interested in something extra. 

For example I left the `fendermint_vm_events` crate as events from everything under the "vm" namespace.. okay. But I wouldn't want them to know about the events emitted by `fendermint_app` during the ABCI lifecycle. Or for example we could emit events from the `fendermint_eth_api` package, which have no reason to be in the same module as the VM or ABCI events.

So I thought we can make this call as as it makes sense, to not hide them away in the two dozen crates, but also not make it into a totally centralised taxonomy.

aakoshh:
I think `Valuable` will be cool when it's stable, it's more intuitive than having to flatten everything.

PR #797, opened 2024-03-15
Support dry-run for create subnet

cryptoAtwill:
Closes https://github.com/consensus-shipyard/ipc/issues/949

This PR supports `dry-run` that, at the current stage, prints the calldata/params that sends to the blockchain. When the `dry-run` is set, a `DryRunProvider` is used instead to provide more behaviours.

While implementing `dry-run`, existing default parameters and hardcoded parameters are scattered, there is no one place that handles all the default parameters. This makes consistent implementation difficult. This PR also makes some code changes so that one can update or check the default parameters in one place.

When this PR is merged, other methods will be supported.


ipc/api/src/evm.rs
```diff
@@ -232,6 +233,60 @@ impl TryFrom<SupplySource> for register_subnet_facet::SupplySource {
...
+        let active_validators_limit = params
+            .active_validators_limit
+            .ok_or_else(|| anyhow!("active validators limit is not defined"))?;
+        let power_scale = params
+            .power_scale
+            .ok_or_else(|| anyhow!("power scale is not defined"))?;
+        let consensus = params
+            .consensus
+            .ok_or_else(|| anyhow!("consensus is not defined"))? as u64
+            as u8;
```
aakoshh, line 252: Can this just be `as u8` instead of `as u64 as u8`?

ipc/provider/src/dryrun.rs
```diff
@@ -0,0 +1,53 @@
...
+impl IPCDryRunProvider {
+    pub fn create_subnet(&self, params: ConstructParams) -> anyhow::Result<()> {
+        let params = self.preflight.create_subnet(params)?;
+        let params = register_subnet_facet::ConstructorParams::try_from(params)?;
+
+        let data = to_fvm_calldata(
+            &register_subnet_facet::REGISTERSUBNETFACET_ABI.functions,
+            "newSubnetActor",
+            params,
+        )?;
```
aakoshh, line 29: This suggests this dryrun functionality should live under the `evm` module.

ipc/api/src/subnet.rs
```diff
@@ -67,18 +67,50 @@ pub enum SupplyKind {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ConstructParams {
+    // optional fields, set to default if not provided
+    pub ipc_gateway_addr: Option<Address>,
+    pub power_scale: Option<i8>,
+    pub consensus: Option<ConsensusType>,
+    pub majority_percentage: Option<u8>,
+    pub active_validators_limit: Option<u16>,
```
aakoshh, line 75: Can you explain why these needed to be made optional because of the `--dry-run`?

ipc/cli/src/commands/subnet/create.rs
```diff
@@ -42,34 +42,23 @@ impl CreateSubnet {
...
-    type Arguments = CreateSubnetArgs;
+        let params = ConstructParams::new(
+            parent,
+            arguments.min_validators,
+            f64_to_token_amount(arguments.min_validator_stake)?,
+            arguments.bottomup_check_period,
+            f64_to_token_amount(arguments.min_cross_msg_fee)?,
+            arguments.permission_mode,
+            supply_source,
+        );
```
aakoshh, line 54: There doesn't seem to be a way to set the non-default parameters. I'm wondering if it would be cleaner to make those fields non-optional, and pass the defaults to this method, so that it can fill them unconditionally, and then the caller has the option to overwrite them if they have the means to do so. I think it would be better than leaving them potentially empty and throwing errors in the final constructor transformation if you forgot to call this pre-flight method.

ipc/cli/src/commands/subnet/create.rs
```diff
@@ -42,34 +42,23 @@ impl CreateSubnet {
...
         };
-        let addr = provider
-            .create_subnet(
-                from,
-                parent,
-                arguments.min_validators,
-                f64_to_token_amount(arguments.min_validator_stake)?,
-                arguments.bottomup_check_period,
-                arguments
-                    .active_validators_limit
```
aakoshh, line 53: It looks like this field isn't used any more :eyes:

aakoshh (commented):
Looking at how the provider / manager code is structured, I think the way I'd approach this would be to add a _property_ to the `IPCProvider`, which would be `false` by default, but `--dry-run` would set it to true, and then in the `connection()` method it would create a version of the `EvmManager` that is dry-run compatible - or return an error if the manager doesn't support dry runs. 

Then, from the manager, I'd print what needs to be printed and call `System::exit`, like `--help` does. I suppose it could even be a dry-run compatible HTTP client that performs read-only queries to do all the gas, but then quits at the first side-effecting method. 

That sounds hacky, but it avoids the issue with the non-existing return type, makes it easier to satisfy the type system without having to switch in the command handler. 

Maybe I'm completely wrong, I never thought about how do dry-runs sanely and nicely before.

ipc/provider/src/dryrun.rs
```diff
@@ -0,0 +1,53 @@
...
+        log::info!(
+            "params: {}",
```
aakoshh, line 31: I have now re-read the ticket, and it looks like the purpose of this story is that other tools can pick up the output. 

As I have said multiple times, there is a difference between logging and output. Logging should go to _stderr_ and purposeful output to _stdout_, so that you can pipe it to the next tool without being conflated with logs, or having to parse the thing out from among timestamps, or as here, English prefix. 

For this purpose I always recommended using `println!` with `json!`, like in `fendermint`, which is human readable as well as easily parseable with `jq`. A Base64 string in the logs is not human readable and I imagine difficult to parse as well.

aakoshh: I'm not sure if you have seen https://linear.app/interplanetary-consensus/issue/ENG-780/implement-dry-run-for-ipc-cli-commands#comment-b4ad2e7d about the list of fields that @raulk thought this tool should print. 

This is where it's not entirely clear what `--dry-run` was supposed to be. Just thinking about it, I would have guessed that it would print out what it would send to the ethereum API, something you could copy and use with `curl` for example, so you have a sense of everything, the destination, the sending user, etc. 

Instead, the ticket is talking about the multisig, printing stuff that can be signed and sent to the `Multisig` actor, which means an FVM `Message` needs to be constructed, which the `ipc-cli` doesn't do, although the `params` constructed here would be part of it. However, as we discussed it would potentially be helpful for them to know what the other knowable fields wouldbe: the `to`, the `method_number`, `value`; we can print `from` to be this signer, even though it's useless for multisig, just to keep it general. 

So, just wondering if this field is enough, as it keeps the `ipc-cli` from having to construct a `Message`, but (probably) still makes the use of `msig` possible, or do you want it to list more fields.

aakoshh: OTOH you can forget what I said about doing this on the HTTP level as it would print a signed transaction request, which isn't what is ultimately being sent to the EVM actor. Although maybe it _would_ make sense to detect the JSON-RPC request going to `eth_sendRawTransaction` and just replicate [these lines](https://github.com/consensus-shipyard/ipc/blob/ae750fe9778d2b6dda10768c3b6966ba3c648238/fendermint/eth/api/src/apis/eth.rs#L606-L621) and print every field from the `Message`.

aakoshh: Maybe this is what I'd do: instead of `--dry-run` begin an on/off flag, it could be a `--dry-run=[message|raw]` format selector, which, if set, would result in the `eth_sendRawTransaction` begin hijacked and the transaction printed either as raw RLP, or as parsed `Message` fields in JSON. 

This would allow all other calls such as fee estimation and nonce checking to go through, so you would get a fully configured transaction to print.

The only problem with it in our case is that `eth_call` and `eth_estimateGas` would fail because they actually execute the transaction, and the caller would not be authorized to actually set the federated powers, only the multisig would, so I assume we'd never get to the point of actually sending anything. You would have to prevent gas estimation by setting some dummy values. But that can be achieved, because the `ipc-cli` could know that when it's asked to set the federated powers, to avoid triggering an estimation.

ipc/api/src/subnet.rs
cryptoAtwill: no no, it's just previously they are hardcoded or scattered around. From the old code, I inferred that they are optional, so I put them in one place.

ipc/cli/src/commands/subnet/create.rs
cryptoAtwill: Then we need a builder perhaps

aakoshh: Your pattern of not-really-optional-but-optional-during-build is used in ethers, but that's because if not set, they can trigger all sorts of calculations to fill them in (nonce lookup, gas estimation, fee estimation). I could be wrong but our usecase seems simpler here.

cryptoAtwill: so a check is performed before sending to chain, I think I added into the data conversion. The setting optional part is not added yet as they are not exposed in `ipc-cli`.

cryptoAtwill:
@aakoshh Initially I thought of another implementation of `EvmManager`, but then a lot of the methods do not really need a `dry-run`, such as a getter methods or like `get_current_chain_head`. Also their return response will be different, so at a provider level might be more appropriate.

ipc/provider/src/dryrun.rs
cryptoAtwill: I'm not sure if we want to perform any nonce check because `ethers` should have checked for us and dry-run, mostly should be just print out what you are sending?

I think `--dry-run=[message|raw]` works for me, but curious why not set it as a separate field?

ipc/provider/src/dryrun.rs
cryptoAtwill: it's kind of awkward in the sense that the calldata for fvm is derived from evm.

cryptoAtwill: it has to go through evm first, then convert to fvm.

cryptoAtwill:
Probably a good way is to create a `DrynRun` trait, then the `connection()` method returns the `DryRun` trait instance if `dry-run` is set.

ipc/provider/src/dryrun.rs
cryptoAtwill: btw, when I'm implementing it, I feel `raw` somehow does not make sense without the context, like `evm` or `fvm`, I kind of feel it's two params, `network` and `format`. Without the network, I dont really know what does `raw` mean. I kind of feel if we know the network, i.e. `dry-run=[evm|fvm]` and prints the fields of the txn, it should be enough for now?

ipc/cli/src/commands/subnet/create.rs
aakoshh: I know the check is there before sending to the chain. Sorry, what I was trying to say is that I think this is a bit convoluted, when all that happens is that the defaults will be set by this provider implementation or that provider implementation.. might as well be set when the record is created in `new` and then there is no need for checks or calling anything, _unless_ you want to override something. Since none of these fields require compilicated estimations _iff_ they are empty, I don't see what benefits leaving them as `None` has.

ipc/provider/src/dryrun.rs
aakoshh: Yeah, `evm|fvm` would work great :+1:

aakoshh: Or, well, maybe not. You see the `fvm` is a given, we always send a `Message` type at the end. But `evm` is specific to subnets: some can be of the `fevm` type (which is the only one supported right now), but you have a `Provider` abstraction in case it's not. That's what I meant by `raw`: print out what you would send to the API that corresponds to the subnet. I don't want an non-evm provider to produce evm format, only it's own, or potentially `fvm`.

aakoshh: I'm just trying to make sense of how to interpret this whole `--dry-run` in a generic way. Obviously every EVM request can be printed, which, as you say, "should be just print out what you are sending", however the FVM version is _not_ what _you_ are sending, it's something extra, but it is what Raul was actually asking for, but it doesn't sit well with `--dry-run` as a general concept IMO. That's why I thought `--dry-run=fvm` can give the hint. 

Whether it's one fields or not: if it would be two fields, then `--dry-run` could be the default "API native" format, ie. EVM in this case, and you could use a `--dry-run-format=fvm` to differ from that.

ipc/provider/src/dryrun.rs
aakoshh: The `params` for the `Message` are indeed derived from the `calldata` of the EVM transaction request. However the fact that this even knows that it's calling one of the IPC Solidity contracts means it knows that the subnet is implemented by `fevm` (the only type currently, but not necessarily forever), so _this_ particular thing would _only_ work as a generic parameter to print _iff_ we already know the subnet is implemented with Solidity, therefore _this_ dryrun implementation should also be under the `evm` namespace.

ipc/cli/src/commands/subnet/create.rs
cryptoAtwill: Ok, I see where you are coming from. Yeah, I can do this too. I thought we might have to config those `Optional` parameter. But we can do those later too.

cryptoAtwill:
@aakoshh I think maybe we can try implementing a `DryRun` version of `SubnetManager`, like you mentioned above in `connection` we can return the `DryRun` version based on the `DryRun` targeting network(evm/fvm). 

The only thing we have to accept is the `DryRun` versions should return just defaults, as the `SubetManager` interfaces are set. Say the `create_subnet` method returns the address of the subnet created, then for the `DryRun` version, we just return `address(0)`. For `join_subnet` method, it returns the epoch that the sender joins the subnet, then for `DryRun` version, we jut return epoch `0`.

In this way, the interfaces are satisfied and the abstractions will not break.

ipc/api/src/evm.rs
```diff
@@ -232,6 +233,45 @@ impl TryFrom<SupplySource> for register_subnet_facet::SupplySource {
...
+        let ipc_gateway_addr = payload_to_evm_address(gateway.payload())?;
+
+        let min_validator_stake = params
+            .min_validator_stake
+            .atto()
+            .to_u128()
+            .ok_or_else(|| anyhow!("invalid min validator stake"))?;
+
+        let route = subnet_id_to_evm_addresses(&params.parent)?;
+        log::debug!("root SubnetID as Ethereum type: {route:?}");
```
aakoshh, line 252: Do we really need to keep this debug log in a conversion function?

aakoshh: I mean I can imagine it aided you during development, but does it serve an ongoing purpose for the operator interacting with the CLI?

ipc/cli/src/commands/subnet/create.rs
aakoshh: This field is still not passed on to the constructor, nor is it set after.

ipc/provider/src/lib.rs
```diff
@@ -41,6 +37,15 @@ pub mod manager;
 const DEFAULT_REPO_PATH: &str = ".ipc";
 const DEFAULT_CONFIG_NAME: &str = "config.toml";
 
+/// The vm type the dry run performs in
+#[derive(Debug, Clone, Copy, strum::EnumString)]
+pub enum VMType {
+    #[strum(serialize = "Evm", serialize = "evm")]
```
aakoshh, line 43: `clap` has its own `ValueEnum` type. Why do we need `strum` here? 

What does it do with multiple `serialize` options?

aakoshh: Ah, different crate :confused:

ipc/provider/src/manager/evm/dryrun.rs
```diff
@@ -0,0 +1,405 @@
...
+
+#[derive(Serialize)]
+struct MockedEVMTxn {
+    from: ethers::types::Address,
+    to: ethers::types::Address,
+    value: U256,
+    #[serde(serialize_with = "serialize_bytes_to_str")]
+    calldata: Vec<u8>,
+    #[serde(serialize_with = "serialize_bytes_to_str")]
+    method: Selector,
```
aakoshh, line 59: Can you add a docstring about the relationship between `method` and `calldata`? I see `method` is not used in the conversion below.

ipc/provider/src/manager/evm/dryrun.rs
```diff
@@ -0,0 +1,405 @@
...
+        params: ConstructParams,
+    ) -> anyhow::Result<Address> {
+        let evm = self.create_subnet_evm(&from, params)?;
+
+        let s = match self.vm_type {
+            VMType::Evm => serde_json::to_string(&evm)?,
+            VMType::Fvm => serde_json::to_string(&MockedFVMTxn::try_from(evm)?)?,
+        };
+
+        println!("txn to send: \n {}", s);
```
aakoshh, line 99: If we want to pipe the output to `jq` then it would be better not to mix in non-JSON lines into the output. If you must, you can use logging (which goes to stderr), or maybe wrap the JSON into another one like `{"create_subnet": s}` in case you intend to print multiple outputs (because your dry-run doesn't stop at the first but keeps going with fake data).

ipc/provider/src/manager/evm/dryrun.rs
```diff
@@ -0,0 +1,405 @@
...
+            (converted,),
+        )?;
+
+        Ok(MockedEVMTxn {
+            from,
+            to,
+            value: U256::zero(),
+            method,
+            calldata,
+        })
```
aakoshh, line 381: This seems like a lot of code for the dry-run, which, AFAICS, will never actually be verified for correctness. That is, unlike what I suggested, which was:
1. Do everything normally, let the `ipc-cli` construct the _real_ calldata and send it to the API...
2. ... except in dry-run mode short-circuit before hitting the `eth_sendRawTransaction` endpoint and just print the parameters. 
3. This way we _know_ that what we print is correct, because we test it with _non-dry-run_ mode all the time. 

By contrast, here the dry-run mode is entirely separate not just in execution, but in its construction of the call parameters. It's nice that the ABI checks some of the compatibility things :crossed_fingers: but I'm not sure what would happen if e.g. we forgot to pass the argument as a tuple.

aakoshh: Hopefully it's fine as `encode_function_data` does type checking, so for methods which have very different input this should be okay.

ipc/provider/src/manager/evm/dryrun.rs
```diff
@@ -0,0 +1,405 @@
...
+        &self,
+        from: &Address,
+        params: ConstructParams,
+    ) -> anyhow::Result<MockedEVMTxn> {
+        let converted = register_subnet_facet::ConstructorParams::try_from(params)?;
+        log::debug!("converted constructor params: {converted:?}");
+
+        let to = converted.ipc_gateway_addr;
+        let from = payload_to_evm_address(from.payload())?;
+        log::debug!("from: {}, to: {}", from, to);
```
aakoshh, line 366: Do we need this debug? They will be printed as part of the dry-run anyway.

ipc/provider/src/manager/evm/dryrun.rs
```diff
@@ -0,0 +1,405 @@
...
+const INVOKE_CONTRACT: MethodNum = 3844450837;
+
+#[derive(Serialize)]
+struct MockedFVMTxn {
+    #[serde(serialize_with = "serialize_address_to_str")]
+    from: Address,
+    #[serde(serialize_with = "serialize_address_to_str")]
+    to: Address,
+    /// The value, display as string to align with evm display
+    value: String,
```
aakoshh, line 45: Using `String` here seem strange. I don't understand how it's aligned with EVM display, what does that mean?

aakoshh (commented):
Thanks for changing the defaults, just a minor comment about a missing field. 

You obviously came to a different conclusion than me on the issue of how to deal with static typing and dry-run: instead of short circuiting and exiting after printing the genuine call parameters as I suggested, you are mocking and faking return data to make it type check. 

If we only think about showing what kind of things would happen during a call, this seems reasonable. I'm not sure how much sense it makes in the context of multisig signing of transactions, it might be confusing that after the first transaction, every subsequent one will be filled with fake data and be useless for this purpose. I guess you can say that whoever wants to use this for multisig has to understand that, and that chaining transactions is rare anyway. 

Were it me I'd baulk at the sheer amount of duplicate work this version of dry-run has to do. With my recommendation you would be done, covered all possible transaction types with one change. With this, you have enabled dry-run for just one command, and it will need careful construction of parameters for all other commands to support it. Seems like a lot of effort. 

I'll leave it up to @raulk to accept as he asked for the feature.

ipc/api/src/evm.rs
cryptoAtwill: yeah, we can remove them, just copy pasted from previous code.

ipc/provider/src/manager/evm/dryrun.rs
cryptoAtwill: When I tried it, actually we dont really need this line. I will remove it.

cryptoAtwill:
@aakoshh I think your suggestion at the middleware level is very cool. But I need to do some research here because after the `send_transaction` is called, the caller is expecting some kind response. I'm not sure how that plays out with dry-run. We have to do some mocking of the response based on the txn sent in. I need to some checks first.

cryptoAtwill:
@aakoshh Interesting one! I think it should work, just `dry-run` will no longer work in non-cli environments.

aakoshh:
> just dry-run will no longer work in non-cli environments.

Yes, absolutely; I'm not sure that was a requirement though. My "dry run" in the materializer made every return value and input a generic parameter, so instead of returning e.g. a real subnet with fake values, it returned a "thing" that you had to pass back next time you wanted to use it. It's quite complicated. I'm clearly just suggesting a quick-and-dirty trick here, if that is compatible with the vision.

cryptoAtwill:
@raulk might need your input here, are we going to use `dry-run` outside of CLI?

raulk:
We never got to consensus here and this became stale. It's not worth investing time in this now, so closing for now.

PR #798, opened 2024-03-15
FIX: filter event by contract

cryptoAtwill:
Closes ENG-785.

This PR fixes a bug from calling `ethers` event query. Currently the contract address is not specified, which means other subnets' events can be queried as well or even attackers faking events. This PR solves this issue by fixing the contract address.

### DIFF
--- a/ipc/provider/src/manager/evm/manager.rs
+++ b/ipc/provider/src/manager/evm/manager.rs
@@ -37,7 +37,7 @@ use ethers::prelude::k256::ecdsa::SigningKey;
 use ethers::prelude::{Signer, SignerMiddleware};
 use ethers::providers::{Authorization, Http, Middleware, Provider};
 use ethers::signers::{LocalWallet, Wallet};
-use ethers::types::{Eip1559TransactionRequest, I256, U256};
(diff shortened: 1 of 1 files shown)


ipc/provider/src/manager/evm/manager.rs
```diff
@@ -1102,7 +1104,8 @@ impl BottomUpCheckpointRelayer for EthSubnetManager {
         let ev = contract
             .event::<lib_quorum::QuorumReachedFilter>()
             .from_block(height as u64)
-            .to_block(height as u64);
+            .to_block(height as u64)
+            .address(ValueOrArray::Value(contract.address()));
```
raulk, line 1108: Just to confirm: this won't really matter because this flow only watches a single subnet anyway: the one represented by the gateway. Right? But it's safer 👍

raulk (approved):

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: It should be fine, quorum reached is emitting from the gateway of the subnet. It's not really about its children. It's just one subnet, no other subnets.

aakoshh (commented):
LGTM!

PR #801, opened 2024-03-15
FIX: emit event name as a field

aakoshh:
Followup for https://github.com/consensus-shipyard/ipc/pull/796

Adds back the name of the event as `event = <name>` because it turns out that there is [no way to configure](https://github.com/tokio-rs/tracing/blob/908cc432a5994f6e17c8f36e13c217dc40085704/tracing-subscriber/src/fmt/format/mod.rs#L930) the subscriber to display the `name` - which would be useless in other cases anyway as it defaults to filename and line which are already displayed.

### DIFF
--- a/fendermint/tracing/src/lib.rs
+++ b/fendermint/tracing/src/lib.rs
@@ -10,6 +10,14 @@
 /// Once the [valuable](https://github.com/tokio-rs/tracing/discussions/1906) feature is stable,
 /// we won't have the restriction of flat events.
 ///
+/// The emitted [tracing::Event] will contain the name of the event twice:
+/// in the [tracing::metadata::Metadata::name] field and under the `event` key in the [tracing::field::ValueSet].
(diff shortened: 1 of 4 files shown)


fendermint/tracing/src/lib.rs
```diff
@@ -38,7 +38,7 @@ macro_rules! emit {
         tracing::event!(
             name: stringify!($event),
             tracing::Level::INFO,
-            { $($field $(= $value)?),* }
+            { event = tracing::field::display(stringify!($event)), $($field $(= $value)?),* }
```
raulk, line 41: Isn't this duplicative with `name`? Happy to approve, but maybe drop a note on why we need both?

raulk (approved):

fendermint/tracing/src/lib.rs
aakoshh: Thanks, added the docs

PR #805, opened 2024-03-16
fix: missing enum variant for StakingOperation.

raulk:
## Problem

The IPC API module was missing enum variant StakingOperation#SetFederatedPower. The fallback branch of the u8 to enum ended up representing unrecognized ABI values as SetMetadata operations. Consequently, SetFederatedPower commands were being wrongly translated and applied in the subnet gateway as SetMetadata commands.

## Solution

Add the missing enum variant and make the conversion typesafe by using try_from. To avoid being pedantic, use num_enum macros. This create was anyway part of the build graph already, so we might as well use it.

## Credits

Shoutout to @mikevoronov and @folex from Fluence Labs for finding and reporting this!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4962,6 +4962,7 @@ dependencies = [
  "lazy_static",
  "log",
  "num-traits",
+ "num_enum",
  "serde",
  "serde_json",
  "serde_tuple",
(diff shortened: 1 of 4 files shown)


mikevoronov (approved):
LGTM

raulk:
Merging this myself, as main is strictly worse off without this fix.

PR #807, opened 2024-03-16
fix(federated validation): make subnet gateway use federated power to calculate power table

raulk:
Federated validation is a bit messy right now. There are two phases:

1. Pre-bootstrap and pre-genesis: the sum of collateral and explicit power determine the weight of a validator.
2. Post-bootstrap and post-genesis: only setFederatedPower calls are allowed. Collateral-based [changes are disallowed](https://github.com/consensus-shipyard/ipc/blob/0855f0fd0b267942e3e2e4b3bf5ff8d5fbe0596d/contracts/src/subnet/SubnetActorManagerFacet.sol#L161).

This fixes the weight calculation in the child within phase 2, summing the explicit power to the collateral-based power (fixed at genesis).

Note: This calculation should not happen inside the subnet. See design discussion at https://github.com/consensus-shipyard/ipc/pull/784.


raulk:
Going ahead and merging this, as we confirmed the bug live.

PR #808, opened 2024-03-17
ci: use GHA docker cache.

raulk:


jsoares:
I fixed the typo that was erroring out, but the github action cache can't really be used in this way as you're not passing a url and token.

The recommended method seems to be to use [docker/build-push-action](https://docs.docker.com/build/cache/backends/gha/#using-dockerbuild-push-action) in CI given these parameters are not exposed, and the workarounds are [pretty bad](https://github.com/orgs/community/discussions/42856).

jsoares (commented):
I fixed the minor typo that was erroring out, but the github action cache can't really be used in this way as you're not passing a url and token.

The recommended method seems to be to use [docker/build-push-action](https://docs.docker.com/build/cache/backends/gha/#using-dockerbuild-push-action) in CI given these parameters are not exposed, and the workarounds are [pretty bad](https://github.com/orgs/community/discussions/42856).

PR #812, opened 2024-03-17
[workaround]: count federated power into collateral ignoring `permissionMode`

cryptoAtwill:
The weight of a validator does not include the federated power in the child gateway. This is causing the validator's weight to be calculated wrongly in the child. This PR updates the weight to include federated power.


raulk (commented):
@cryptoAtwill this description is inaccurate, and this implementation is a temporary workaround. The federated power is being accounted for, it's just that the genesis permission mode is not being correctly set in the child gateway. Therefore the current if does not match. The correct solution under the current design is to correctly propagate the configuration to the child gateway, so that the permission mode is filled as expected.

I suggest we keep the if block but comment it out, and add a TODO. Also adjust the PR description to describe the actual problem.

raulk:
This was tested successfully by Fluence, so proceeding to merge.

raulk (approved):

PR #813, opened 2024-03-17
fix(cli): set-federated-power: use correct signer middleware.

raulk:
We received multiple complaints that `ipc-cli subnet set-federated-power` was invariably failing due to a contract revert. It turns out we were not correctly setting the `from` account as the signer of the transaction.

This PR also applies other minor fixes.


raulk:
@mb1896 this probably worked in your tests because your default wallet account was the creator of the subnet, or something like that.

mb1896 (approved):

PR #814, opened 2024-03-17
extras/axelar-token: improve error handling and add recovery paths.

raulk:
1. We introduce the notion of an "admin" address as the recovery operator.
2. We now handle failures when parsing semi-untrusted input, and when invoking the IPC gateway. If any of those fail, we increment the allowance of the admin so they can operate the otherwise locked tokens.
3. When the subnet reports a failure crediting the deposit (e.g. such as if the recipient was a smart contract / actor that reverted), we increment the allowance of the admin also. This is more failsafe and easier to operate than crediting the funds to the recipient on the L1 (who wouldn't be able to rescue them if it was a smart contract, unless it could manage to deploy itself onto the same address it had in the subnet, plausible if it used Deploy3).
4. As an ultimate backstop, we introduce a method so that the operator can increase its allowance on any token. This should only be used iff something unexpected failed, like the error handling paths themselves.

### DIFF
--- a/extras/axelar-token/src/IpcTokenHandler.sol
+++ b/extras/axelar-token/src/IpcTokenHandler.sol
@@ -31,14 +31,15 @@ contract IpcTokenHandler is InterchainTokenExecutable, IpcHandler {
 
     error NothingToWithdraw();
 
-    TokenFundedGateway public _ipcGateway;
-    mapping(address beneficiary => mapping(address token => uint256 value)) private _claims;
-
     event SubnetFunded(SubnetID indexed subnet, address indexed recipient, uint256 value);
     event FundingFailed(SubnetID indexed subnet, address indexed recipient, uint256 value);
 
-    constructor(address axelarIts, address ipcGateway) InterchainTokenExecutable(axelarIts) {
+    TokenFundedGateway public _ipcGateway;
+    address public _admin;
+
+    constructor(address axelarIts, address ipcGateway, address admin) InterchainTokenExecutable(axelarIts) {
         _ipcGateway = TokenFundedGateway(ipcGateway);
+        _admin = admin;
     }
 
     // @notice The InterchainTokenExecutable abstract parent contract hands off to this function after verifying that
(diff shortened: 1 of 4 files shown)


extras/axelar-token/src/IpcTokenHandler.sol
```diff
@@ -52,13 +53,38 @@ contract IpcTokenHandler is InterchainTokenExecutable, IpcHandler {
...
         uint256 amount
     ) internal override {
-        (SubnetID memory subnet, address recipient) = abi.decode(data, (SubnetID, address));
-
         IERC20 token = IERC20(tokenAddr);
         require(token.balanceOf(address(this)) >= amount, "insufficient balance");
 
         // Authorize the IPC gateway to spend these tokens on our behalf.
-        token.approve(address(_ipcGateway), amount);
+        token.safeIncreaseAllowance(address(_ipcGateway), amount);
```
snissn, line 60: I wasn't aware of safeIncreaseAllowance - it is helpful!

extras/axelar-token/src/IpcTokenHandler.sol
```diff
@@ -52,13 +53,38 @@ contract IpcTokenHandler is InterchainTokenExecutable, IpcHandler {
...
+        if (!success) {
+            // Restore the original allowance.
+            token.safeDecreaseAllowance(address(_ipcGateway), amount);
+
+            // Increase the allowance of the admin address so they can retrieve these otherwise lost tokens.
+            token.safeIncreaseAllowance(_admin, amount);
+
+            // Emit a FundingFailed event; we can't associate a specific subnet or recipient since parsing may have failed.
+            SubnetID memory nilSubnet;
+            emit FundingFailed(nilSubnet, address(0), amount);
```
snissn, line 77: Can we want to emit msg.sender or tx.origin here? Or will that tend to be not helpful information like "gateway" or "ipc subnet" would be the callers?

extras/axelar-token/src/IpcTokenHandler.sol
```diff
@@ -78,38 +104,28 @@ contract IpcTokenHandler is InterchainTokenExecutable, IpcHandler {
...
-        }
-
-        delete _claims[msg.sender][token];
-        IERC20(token).safeTransfer(msg.sender, available);
+    // @notice The ultimate backstop in case the error-handling logic itself failed unexpectedly and we failed to
+    //         increase the recovery allowances of the admin address.
+    function adminTokenIncreaseAllowance(address token, uint256 amount) external {
+        require(msg.sender == _admin, "unauthorized");
+        IERC20(token).safeIncreaseAllowance(_admin, amount);
     }
```
snissn, line 129: If this is our back pocket trump card to unroll any errors, can we get a simple test for the recovery?

snissn (approved):
I think the changes are productive and aim to properly increase security and create recovery against potential issues. I would love to see a test for the admin token recovery flow, but on face value it is simple and appears to be correct

extras/axelar-token/src/IpcTokenHandler.sol
raulk: Done here: https://github.com/consensus-shipyard/ipc/pull/814/commits/f0a035aa8765517a30c9ebf1e31535522a6cc7ab

extras/axelar-token/src/IpcTokenHandler.sol
raulk: I refactored the solution a bit, and now we're able to pass actual data here.

crystalbit (approved):
looks good

PR #815, opened 2024-03-18
Prefund with token

cryptoAtwill:
Adding `prefundWithToken` method that allows users to deposit funds into the genesis when the `supplySource` is `ERC20`.

Integration testing in progress.

Closes ENG-799.


raulk (commented):
The subnet actor will need to transfer its genesis circulating supply to the gateway actor when bootstrapping.

cryptoAtwill:
Currently this PR is working on its own, but it's not working in integration tests due to this [line](https://github.com/consensus-shipyard/ipc/blob/a33cfee763cafaafdd00ad38a1367ee262e75fa4/contracts/src/gateway/GatewayManagerFacet.sol#L40). The gateway is still expected the genesis circulation in native tokens. If this part is not updated, then this PR will not work.

jsoares:
@raulk :  aim to land this and then refactor gateway; use the supplyhelper instead of checking msg.value

PR #816, opened 2024-03-18
ENG-797: Add parent_http_auth_token to topdown settings

aakoshh:
Exposes the `auth_token` setting of the IPC provider in the fendermint settings.

### DIFF
--- a/fendermint/app/settings/src/lib.rs
+++ b/fendermint/app/settings/src/lib.rs
@@ -236,9 +236,15 @@ impl Settings {
 
     /// Load the default configuration from a directory,
     /// then potential overrides specific to the run mode,
-    /// then overrides from the local environment.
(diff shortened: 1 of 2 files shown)


raulk (approved):
Nice! Also copying from Slack:

> The ipc-cli supports the header in the config.toml It's part of the [EvmSubnet](https://github.com/consensus-shipyard/ipc/blob/c4c210b595acf234362d64600d49e41e940ee028/ipc/provider/src/config/subnet.rs#L89)  definition and gets set [here](https://github.com/consensus-shipyard/ipc/blob/c4c210b595acf234362d64600d49e41e940ee028/ipc/provider/src/manager/evm/manager.rs#L1014-L1022).

@aakoshh the pesky CI config test error is back...

PR #818, opened 2024-03-18
FIX: Use f410 delegated address in EAM `can_deploy`

aakoshh:
The `FvmRuntime` returns an `ActorID` wrapped as an `Address`, but this is almost impossible to tell up front.


raulk:
Superseded by #819.

PR #819, opened 2024-03-18
fix(eam): normalize allowlist addresses in can_deploy.

raulk:
The current logic is unable to match deployments because `msg.caller()` returns an ID address, whereas users add f1, f2, f3, and f410 addresses to the allowlist. This PR introduces address normalization.


fendermint/actors/eam/src/state.rs
```diff
@@ -51,16 +53,23 @@ impl State {
...
+                let deployers =
+                    DeployerMap::load(rt.store(), cid, DEFAULT_HAMT_CONFIG, "verifiers")?;
+                let mut allowed = false;
+                deployers.for_each(|k, _| {
+                    // Normalize allowed addresses to ID addresses, so we can compare any kind of allowlisted address.
+                    // This includes f1, f2, f3, etc.
+                    // We cannot normalize the allowlist at construction time because the addresses may not be bound to IDs yet (counterfactual usage).
+                    // Unfortunately, API of Hamt::for_each won't let us stop iterating on match, so this is more wasteful than we'd like. We can optimize later.
+                    // Hamt has implemented Iterator recently, but it's not exposed through Map2 (see ENG-800).
+                    allowed |= rt.resolve_address(&k) == Some(deployer);
```
aakoshh, line 69: NB this does not prevent the execution of `resolve_address` after `allowed` has already been set to `true`.

aakoshh (approved):

fendermint/actors/eam/src/state.rs
raulk: That's a good optimization, let me apply it.

aakoshh:
Maybe one of the tests could be changed to use the delegated address, like I did in https://github.com/consensus-shipyard/ipc/pull/818

fendermint/actors/eam/src/state.rs
```diff
@@ -51,16 +53,25 @@ impl State {
...
+                let mut allowed = false;
+                deployers.for_each(|k, _| {
+                    // Normalize allowed addresses to ID addresses, so we can compare any kind of allowlisted address.
+                    // This includes f1, f2, f3, etc.
+                    // We cannot normalize the allowlist at construction time because the addresses may not be bound to IDs yet (counterfactual usage).
+                    // Unfortunately, API of Hamt::for_each won't let us stop iterating on match, so this is more wasteful than we'd like. We can optimize later.
+                    // Hamt has implemented Iterator recently, but it's not exposed through Map2 (see ENG-800).
+                    if !allowed {
+                        allowed = rt.resolve_address(&k) == Some(deployer);
+                    }
```
raulk, line 71: @aakoshh is there a more idiomatic way to do this?

aakoshh: This would work (I tested it):
```
allowed = allowed || rt.resolve_address(&k) == Some(deployer);
```
Which one you prefer is a matter of taste I suppose.

fendermint/actors/eam/src/lib.rs
```diff
@@ -36,10 +36,11 @@ impl IPCEamActor {
     }
 
     fn ensure_deployer_allowed(rt: &impl Runtime) -> Result<(), ActorError> {
-        let caller = rt.message().caller();
+        // The caller is guaranteed to be an ID address.
+        let caller_id = rt.message().caller().id().unwrap();
```
aakoshh, line 40: It's a shame it gets wrapped for no reason.

aakoshh (approved):
LGTM! Thanks for fixing it the right way :+1:

fendermint/actors/eam/src/state.rs
raulk: Good all bool short-circuiting ;-)

PR #823, opened 2024-03-19
feat(eam): add `EAM#UpdateDeployers`.

raulk:
- This method allows updating a deployer allowlist.
- It can only be called when the permissioning mode is Allowlist.
- It must be called by a currently authorized deployer.

This method is bound to an FRC42 method number. Its arguments are CBOR-encoded. It can be called from an f1/f3 account by sending a native transaction, or from an f410 account using this path: https://github.com/consensus-shipyard/ipc/pull/773.

### DIFF
--- a/fendermint/actors/eam/src/lib.rs
+++ b/fendermint/actors/eam/src/lib.rs
@@ -140,8 +140,8 @@ mod tests {
         expect_empty, MockRuntime, ETHACCOUNT_ACTOR_CODE_ID, EVM_ACTOR_CODE_ID,
         SYSTEM_ACTOR_CODE_ID,
     };
+    use fil_actors_runtime::INIT_ACTOR_ADDR;
     use fil_actors_runtime::SYSTEM_ACTOR_ADDR;
-    use fil_actors_runtime::{AsActorError, INIT_ACTOR_ADDR};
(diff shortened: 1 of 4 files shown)


fendermint/actors/eam/src/lib.rs
```diff
@@ -59,6 +68,33 @@ impl IPCEamActor {
 
         Ok(())
     }
+
+    fn update_deployers(rt: &impl Runtime, deployers: Vec<Address>) -> Result<(), ActorError> {
```
cryptoAtwill, line 72: do we do checks on the deployers? Like simple dedup, size limiting?

cryptoAtwill: I think maybe we should restrict to size at least 1 so that we still have someone to update the deployers.

cryptoAtwill: oh yeah, it's a map, so no need to do dedup.

raulk: On 0 check: that would make it impossible to remove all deployment privileges, which is a legitimate use case.

fridrik01 (approved):
lgtm, but there are some lint errors

PR #826, opened 2024-03-19
Remove submission to previous checkpoint in relayer

cryptoAtwill:
With the changes that prevents submission to previous checkpoints, the corresponding logic in the `relayer` is removed.


raulk:
Since the relayer behaviour is not consensus critical, queuing this up for the next days.

raulk:
@cryptoAtwill in the future, please select a reviewer to expedite reviews.

raulk:
This also gets rid of double checkpoint submission (we were submitting every checkpoint twice in the past).

raulk (approved):

PR #828, opened 2024-03-20
Set EAM permissions for genesis

mb1896:
This closes ENG-743

We want to set EAM permission when creating the subnet.

Example usage:

To set allowed address:
```bash
cargo make --makefile infra/fendermint/Makefile.toml -e ALLOWED_ADDR_LIST=<addr1>,<addr2> subnet-genesis-set-eam-permissions
```

If allowed address are not provided, will proceed with `unrestricted` EAM permission.
```bash
cargo make --makefile infra/fendermint/Makefile.toml subnet-genesis-set-eam-permissions
```

It can also be executed as a dependency step for a bigger step, such as `tasks.child-validator`, using the same way of  providing input `-e ALLOWED_ADDR_LIST`.

### DIFF
--- a/infra/fendermint/scripts/subnet.toml
+++ b/infra/fendermint/scripts/subnet.toml
@@ -126,8 +126,8 @@ env = { "CMD" = "genesis --genesis-file /data/genesis.json ipc from-parent --sub
 [tasks.subnet-genesis-set-eam-permissions]
 extend = "fendermint-tool"
 script.pre = """
-echo "ALLOWED_ADDR_LIST=${ALLOWED_ADDR_LIST}"
 if [[ ! -z ${ALLOWED_ADDR_LIST} ]]; then
+  echo "ALLOWED_ADDR_LIST=${ALLOWED_ADDR_LIST}"
   CMD="genesis --genesis-file /data/genesis.json set-eam-permissions --mode allowlist --addresses ${ALLOWED_ADDR_LIST}"
 else
   CMD="genesis --genesis-file /data/genesis.json set-eam-permissions --mode unrestricted"
--- a/infra/fendermint/scripts/subnet.toml
+++ b/infra/fendermint/scripts/subnet.toml
@@ -134,10 +134,6 @@ else
 fi
 """
 
-#[tasks.subnet-genesis-set-eam-permissions]
-#extend = "fendermint-tool"
-#env = { "CMD" = "genesis --genesis-file /data/genesis.json set-eam-permissions --mode unrestricted" }
-
 [tasks.node-report]
 script = """cat << EOF
 #################################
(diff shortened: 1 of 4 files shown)


infra/fendermint/scripts/subnet.toml
```diff
@@ -122,6 +123,10 @@ script.pre = "mkdir -p ${BASE_DIR}/${NODE_NAME}/${KEYS_SUBDIR}; cp ${PRIVATE_KEY
 extend = "fendermint-tool"
 env = { "CMD" = "genesis --genesis-file /data/genesis.json ipc from-parent --subnet-id ${SUBNET_ID} -p ${PARENT_ENDPOINT}  --parent-gateway ${PARENT_GATEWAY}  --parent-registry ${PARENT_REGISTRY} --base-fee ${BASE_FEE} --power-scale ${POWER_SCALE}" }
 
+[tasks.subnet-genesis-set-eam-permissions]
+extend = "fendermint-tool"
+env = { "CMD" = "genesis --genesis-file /data/genesis.json set-eam-permissions --mode unrestricted" }
```
mb1896, line 128: @raulk Should we set to allowlist or unrestricted by default?

aakoshh: Isn't this exactly what `genesis --genesis-file /data/genesis.json ipc from-parent ` [did by default](https://github.com/consensus-shipyard/ipc/blob/ae750fe9778d2b6dda10768c3b6966ba3c648238/fendermint/app/src/cmd/genesis.rs#L327) ?

aakoshh:
I'd just like to point out that the linear link requires login, so if all we say in the description that it closes that issue without describing the change, then it would be nice if the URL was public, or if any issue ready to be worked on would be synced to Github. I usually include some description of what I did and why, just in case.

infra/fendermint/scripts/subnet.toml
mb1896: That's right. So what does this Linear issue ask for? Which addresses should be allowed to deploy contracts in this subnet? @aakoshh @raulk

mb1896:
> I'd just like to point out that the linear link requires login, so if all we say in the description that it closes that issue without describing the change, then it would be nice if the URL was public, or if any issue ready to be worked on would be synced to Github. I usually include some description of what I did and why, just in case.

Good point! Updated the PR description.

jsoares:
We currently have a number of Linear tasks that are not synced to github issues but I think that should be an exceptional event (e.g. security issues). We'll need to discuss & revisit our processes, agreed we don't want super opaque changes.

infra/fendermint/scripts/subnet.toml
jsoares: from @raulk in sync: pass an env variable to `cargo make child-validator`

jsoares: from @raulk in sync: pass permission mode as an env variable to `cargo make child-validator`

PR #835, opened 2024-03-21
ENG-629: Event subscription and metrics 

aakoshh:
The PR subscribes to events emitted by the application via `tracing` and feeds them into `prometheus` metrics. 

It doesn't yet include serving the metrics over HTTP; that will be in a followup PR.


raulk (commented):
Kinda LGTM, but it's sad that we're losing the type safety from the original event schema objects. I wonder if we can shoot for an Trait-based abstraction where we decorate each event type with a `record_metrics(Self, <some
Prometheus context>)` method that we then invoke from the emit event macro. I don't see a strong reason why this _needs_ to be downstream of the tracing subscriber?

raulk:
In the future we could add field attributes to bind event struct fields to concrete metrics. Actually, something like that may already exist?

aakoshh:
The base idea was that you can emit events (ideally nested data structures in the future, like we could with the `valuable` feature, not these purposefully flattened versions), without having to think there and then what metrics you want to maintain and how. Sure, you might have an idea, but it would be up to the consumer if it's prometheus, what the metric is called, etc. 

I agree that it's not nice to lose the static typing. Unfortunately even with `valuable` it's just going to be the equivalent of a JSON object.

Maybe I can work on the definition of finding the values to be checked against the existence of fields, let me see..

raulk:
> The base idea was that you can emit events (ideally nested data structures in the future, like we could with the valuable feature, not these purposefully flattened versions), without having to think there and then what metrics you want to maintain and how. Sure, you might have an idea, but it would be up to the consumer if it's prometheus, what the metric is called, etc.

I see conceptually what you're proposing, but I think the level of indirection adds to the complexity. I'd find it easier as a Fendermint dev to be forced to think about how a particular event is going to transpire into metrics as soon as I add the event schema. If I'm unsure, I can always leave it as a TODO. Honestly I'm not convinced that the indirection buys us anything?

raulk (approved):
BTW, I'm happy to try this out, no need to block on my opinion. We can always refactor later if we agree we can improve the ergonomics.

aakoshh:
> Honestly I'm not convinced that the indirection buys us anything?

It's certainly no panacea. In the "original" it was a statically typed dependency that could be used to emit certain kind of events the module raised, and it was an easy way to say: okay, I want to log this part with this level, turn that part into a metric, without e.g. having to modify the consensus module, re-audit it, etc. Using `tracing` is a bit of a straight jacket.

I think what it buys from my perspective is that there's only one thing to concentrate on at a time: what the domain events are is considered separately from whether they should be logged or turned into metrics. They exist on their own right. For example here, two events aren't turned into metrics because they are already available from CometBFT, so there's no need for it. All the parts are decoupled from each other, which I like.

OTOH I do appreciate how Bitswap comes with its own metrics and we don't have to study its events to map them :shrug: 

You are right, if it doesn't work for us, we can always find a different way. Certainly it's easier to hunt these down than ad-hoc metrics and logs scattered around the code base.

PR #840, opened 2024-03-25
Support submit checkpoint in parallel

mb1896:
This closes ENG-767.

We now support submitting bottum-up checkpoint with limited parallelism, configured by `--max-parallel-submission` flag in `ipc-cli checkpoint relayer` command.

### DIFF
--- a/ipc/provider/src/checkpoint.rs
+++ b/ipc/provider/src/checkpoint.rs
@@ -191,17 +191,14 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
                     .unwrap();
                 all_submit_tasks.push(tokio::task::spawn(async move {
                     let height = event.height;
-                    if let Err(e) =
+                    let result =
                         Self::submit_checkpoint(parent_handler_clone, submitter, bundle, event)
                             .await
-                    {
-                        log::error!("Fail to submit checkpoint at height {height}: {e}");
-                        drop(submission_permit);
-                        Err(e)
-                    } else {
-                        drop(submission_permit);
-                        Ok(())
-                    }
+                            .inspect_err(|err| {
+                                log::error!("Fail to submit checkpoint at height {height}: {err}");
+                            });
+                    drop(submission_permit);
+                    result
                 }));
 
                 count += 1;
--- a/ipc/provider/src/checkpoint.rs
+++ b/ipc/provider/src/checkpoint.rs
@@ -42,6 +42,7 @@ impl<T: BottomUpCheckpointRelayer> BottomUpCheckpointManager<T> {
         child: Subnet,
         parent_handler: T,
         child_handler: T,
+        max_parallel_submission: usize,
     ) -> Result<Self> {
         let period = parent_handler
             .checkpoint_period(&child.id)
@@ -56,7 +57,7 @@ impl<T: BottomUpCheckpointRelayer> BottomUpCheckpointManager<T> {
             parent_handler: Arc::new(parent_handler),
             child_handler,
             finalization_blocks: 0,
-            submission_semaphore: Arc::new(Semaphore::new(4)),
+            submission_semaphore: Arc::new(Semaphore::new(max_parallel_submission)),
         })
     }
 
@@ -71,12 +72,13 @@ impl BottomUpCheckpointManager<EthSubnetManager> {
         parent: Subnet,
         child: Subnet,
         keystore: Arc<RwLock<PersistentKeyStore<EthKeyAddress>>>,
+        max_parallel_submission: usize,
     ) -> Result<Self> {
         let parent_handler =
             EthSubnetManager::from_subnet_with_wallet_store(&parent, Some(keystore.clone()))?;
         let child_handler =
             EthSubnetManager::from_subnet_with_wallet_store(&child, Some(keystore))?;
-        Self::new(parent, child, parent_handler, child_handler).await
+        Self::new(parent, child, parent_handler, child_handler, max_parallel_submission).await
     }
 }
 
--- a/ipc/provider/src/checkpoint.rs
+++ b/ipc/provider/src/checkpoint.rs
@@ -190,7 +190,14 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
                     .await
                     .unwrap();
                 all_submit_tasks.push(tokio::task::spawn(async move {
-                    Self::submit_checkpoint(parent_handler_clone, submitter, bundle, event).await;
+                    match Self::submit_checkpoint(parent_handler_clone, submitter, bundle, event).await {
+                        Ok(()) => {
+                            log::debug!("Successfully submitted checkpoint");
+                        }
+                        Err(_err) => {
+                            log::debug!("Failed to submit checkpoint");
+                        }
+                    }
                     drop(submission_permit);
                 }));
 
@@ -213,7 +220,7 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
         submitter: Address,
         bundle: BottomUpCheckpointBundle,
         event: QuorumReachedEvent,
-    ) {
+    ) -> Result<(), anyhow::Error> {
         let epoch = parent_handler
             .submit_checkpoint(
                 &submitter,
@@ -227,13 +234,13 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
                     "cannot submit bottom up checkpoint at height {} due to: {e:}",
                     event.height
                 )
-            })
-            .unwrap();
+            })?;
 
         log::info!(
             "submitted bottom up checkpoint({}) in parent at height {}",
             event.height,
             epoch
         );
+        Ok(())
     }
 }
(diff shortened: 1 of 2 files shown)


ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,62 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
-                            "cannot submit bottom up checkpoint at height {} due to: {e:}",
-                            event.height
-                        )
-                    })?;
-
-                log::info!(
-                    "submitted bottom up checkpoint({}) in parent at height {}",
-                    event.height,
-                    epoch
+                    .unwrap();
```
cryptoAtwill, line 191: Just curious about the behaviour when the number of checkpoints to submit (say 10) is actually more than number of permits (say 2). So `acquire_owned` will actually block if 2 checkpoints are being submitted until one of 2 actually finishes? Even if `submit_checkpoint ` fails and the thread crashes, `drop(submission_permit)` will be executed and no dead lock here right?

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,62 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                all_submit_tasks.push(tokio::task::spawn(async move {
+                    Self::submit_checkpoint(parent_handler_clone, submitter, bundle, event).await;
```
cryptoAtwill, line 193: Will this actually work in parallel, because there is a `sequential` enforcement in the checkpoint height, so say we are submitting two checkpoints at height 100 and 130. If height 100 is not submitted, 130 cannot be submitted. My question is really about sort of `race` in the sense that if height 100 is still in the memory pool and the transaction for height 130 is submitted, will it pass the gas estimation without errors? Or will it be rejected due to `nonce` not set incrementally?

ipc/provider/src/checkpoint.rs
mb1896: > So acquire_owned will actually block if 2 checkpoints are being submitted until one of 2 actually finishes? 

Yes, it will literally block (the `for` loop just pauses).

> Even if submit_checkpoint  fails and the thread crashes, drop(submission_permit) will be executed and no dead lock here right?

My last version had an issue where the async task can panic before dropping the permit (because the use of `unwrap()`). It's been fixed here so the permit will be dropped no matter it succeeds or fails when submitting the checkpoint.

ipc/provider/src/checkpoint.rs
mb1896: I'm not entirely sure whether the submit of 130 can succeed if the submit of 100 is still in progress...@aakoshh Can you share your opinions here?

aakoshh: I don't know how Lotus works exactly, but let's say that it:
* limits the number of pending transactions in the mempool to 4
* orders the pending transactions by nonce 
* applies them to an in-memory check-state like fendermint

If that were true, then as long as the transactions are _created_ in a logical order, their order could be restored by Lotus when they are included in the block. As for the checks, it probably depends on whether that particular node has seen the preceding transactions are not. 

Note that we would still _send_ the transactions sequentially, we just don't want to _wait_ for the receipt between each submission.

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,71 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
-                    "submitted bottom up checkpoint({}) in parent at height {}",
-                    event.height,
-                    epoch
+                    .unwrap();
+                all_submit_tasks.push(tokio::task::spawn(async move {
+                    match Self::submit_checkpoint(parent_handler_clone, submitter, bundle, event)
+                        .await
+                    {
+                        Ok(()) => {
+                            log::debug!("Successfully submitted checkpoint");
```
aakoshh, line 197: Maybe log which height has been submitted.

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,71 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                        }
+                        Err(_err) => {
+                            log::debug!("Failed to submit checkpoint");
```
aakoshh, line 200:
```suggestion
                        Err(err) => {
                            log::error!(error = err, height = bundle.height, "Failed to submit checkpoint");
```
Not sure how to log the height, but ignoring errors and even demoting the log to `debug` level is a recipe for frustration.

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,71 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+            .submit_checkpoint(
+                &submitter,
+                bundle.checkpoint,
+                bundle.signatures,
+                bundle.signatories,
+            )
+            .await
+            .map_err(|e| {
+                anyhow!(
+                    "cannot submit bottom up checkpoint at height {} due to: {e:}",
```
aakoshh, line 236:
```suggestion
                    "cannot submit bottom up checkpoint at height {} due to: {e:?}",
```
Not sure what `{e:}` is but I suspect it's the same as `{e}`. There is also `{e:#}`.

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,71 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                    event.height
+                )
+            })?;
+
+        log::info!(
+            "submitted bottom up checkpoint({}) in parent at height {}",
+            event.height,
+            epoch
```
aakoshh, line 244:
```suggestion
            height = event.height, epoch, "submitted bottom up checkpoint"
```

ipc/provider/src/checkpoint.rs
aakoshh: I see this is completely redundant with the `info` log [below](https://github.com/consensus-shipyard/ipc/pull/840/files#diff-43df7702e44369062dcd87a6d917f2d9e7a026ef31e8bacedd1248088b064b65R241). 

I would suggest one of two options:
* only do logging here, remove the one below, or
* remove the success logging from here and keep the one below, logging only the errors here: `if let Err(e) = Self::submit_checkpoint(...) { log::error!(error=e, ... }`

ipc/provider/src/checkpoint.rs
mb1896: The `error = err`, `height = bundle.height` do not seem to work. I just log them in a string instead.

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,71 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                    "This round has asynchronously submitted {:} checkpoints",
+                    count
                 );
             }
         }
 
+        log::debug!("Waiting for all submissions to finish");
+        join_all(all_submit_tasks).await;
+
+        Ok(())
```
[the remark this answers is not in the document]
mb1896: Good point! Fixed!

ipc/provider/src/checkpoint.rs
raulk: Indeed, blockchain nodes accept gapped transactions via their RPC and via the mpool, because they live in an eventually consistent universe. Note that any of these can happen, in addition to other scenarios, of course:

* The user submits transactions in sequence to a load-balanced RPC endpoint, and they land gapped in the backend nodes.
* The user submits out of order transactions to the same node, which is just a subcase of the gapped situation.
* The user submits many transactions, all in order, to the same backend node, but each tx takes different gossip propagation routes through the network and it arrives to various nodes in various non-sequential and gapped orders.

ipc/provider/src/checkpoint.rs
raulk: @aakoshh key values were added to the `log` crate in v0.4.21, but our `Cargo.lock` is fixed on v0.4.20. We'll need to upgrade before we can adopt them without using unlocking the `unstable_kv` feature.

ipc/provider/src/checkpoint.rs
```diff
@@ -162,30 +178,70 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                        Self::submit_checkpoint(parent_handler_clone, submitter, bundle, event)
+                            .await
+                    {
+                        log::error!("Fail to submit checkpoint at height {height}: {e}");
+                        drop(submission_permit);
+                        Err(e)
+                    } else {
+                        drop(submission_permit);
+                        Ok(())
+                    }
```
raulk, line 204: If all we're doing is log, we can use [`inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err).

raulk (approved):
This is OK for a start, but it's closer to a batched submission than to actual parallelism. The goal was to have `max-parallelism` active threads submitting checkpoints at all times. Could you please comment on how this was tested?

mb1896:
> ~This is OK for a start, but it's closer to a batched submission than to actual parallelism. The goal was to have `max-parallelism` active threads submitting checkpoints at all times.~
> 
> EDIT: apologies, I think I misread a loop condition. This indeed is performing parallel submissions!
> 
> Could you please comment on how this was tested?

I tested it by running the relayer locally by adding much more debug log (not checked in), and had this observation: 
- Before reaching the parallelism limit, there are no "context switch" ("context switch" here means switching to executing another async code block at `await`, not switching threads) when acquiring the permit from semaphore. I observed is that all permit will be quickly acquired. No submission results ever shows at this point
- Then in the next loop, context switch happened when trying to acquire new permit, the submission result shows, then another context switch happened and a new permit is acquired and a new submission happened. If there's no response from the earliest submission that's currently happening, it will block there waiting for permit
- No deadlock happens and all permit are eventually dropped

I think these observations is what we expected from the change.

PR #842, opened 2024-03-26
ENG-731: Cache transactions in the ethapi

aakoshh:
The PR enhances the `ethapi` to be able to cache ~and buffer~ transactions until they are included in the blockchain, so that it can deal with ~out-of-order submissions and~ plug the gap in the API where an Ethereum transaction is submitted, but it's not found by its hash until it's executed and indexed by its _domain hash_ by CometBFT. 

- [x] Test to show that transaction isn't found by hash
- [x] Add pending transactions to a cache
- [x] Look up transactions in the cache in `eth_getTransactionByHash`
- [x] Subscribe to `NewBlock` events
- [x] Calculate the domain hash of transactions in the `NewBlock` and remove them from the cache

Deferred to future PRs:
- [ ] Look up the maximum nonce in the cache in `eth_getTransactionCount`
- [ ] Detect out-of-order submission error and put it in a buffer
- [ ] Detect transactions enabled in the buffer a newly created block
- [ ] Submit enabled transactions

### Testing

```bash
cargo test -p fendermint_materializer --test docker -- mempool --nocapture
```

### DIFF
--- a/fendermint/eth/api/src/cache.rs
+++ b/fendermint/eth/api/src/cache.rs
@@ -3,27 +3,63 @@
 
 use std::sync::{Arc, Mutex};
 
-use crate::state::ActorType;
 use anyhow::Context;
 use cid::Cid;
-use fendermint_rpc::client::FendermintClient;
-use fendermint_rpc::query::QueryClient;
-use fendermint_vm_message::query::FvmQueryHeight;
+use lru_time_cache::LruCache;
+use tendermint_rpc::Client;
+
 use fvm_shared::{
     address::{Address, Payload},
     ActorID,
 };
-use lru_time_cache::LruCache;
-use tendermint_rpc::Client;
+
+use fendermint_rpc::client::FendermintClient;
+use fendermint_rpc::query::QueryClient;
+use fendermint_vm_message::query::FvmQueryHeight;
+
+use crate::state::ActorType;
+
+// The `LruCache` is wrapped in `Mutex` beause even reading requiers mutation.
+#[derive(Clone)]
+pub struct Cache<K, V> {
+    cache: Arc<Mutex<LruCache<K, V>>>,
+}
+
+impl<K, V> Cache<K, V>
+where
+    K: Ord + Clone,
+    V: Clone,
+{
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            cache: Arc::new(Mutex::new(LruCache::with_capacity(capacity))),
+        }
+    }
+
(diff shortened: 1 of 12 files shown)


fendermint/eth/api/src/cache.rs
```diff
@@ -3,27 +3,73 @@
...
+}
+
+impl<K, V> Cache<K, V>
+where
+    K: Ord + Clone,
+    V: Clone,
+{
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            cache: Arc::new(Mutex::new(LruCache::with_capacity(capacity))),
```
cryptoAtwill, line 35: Does it come with some default time to live as well?

fendermint/eth/api/src/mpool.rs
```diff
@@ -0,0 +1,95 @@
...
+                            tracing::warn!(error=?e, "NewBlocks subscription failed; resubscribing...");
+                            break;
+                        }
+                        Ok(event) => {
+                            if let EventData::NewBlock {
+                                block: Some(block), ..
+                            } = event.data
+                            {
+                                let tx_hashes = collect_tx_hashes(&block, &chain_id);
+                                tx_cache.remove_many(&tx_hashes);
```
cryptoAtwill, line 61: will this block other concurrent operations if there are quite some txn in a block

fendermint/eth/api/src/cache.rs
```diff
@@ -3,27 +3,73 @@
...
-use lru_time_cache::LruCache;
-use tendermint_rpc::Client;
+
+use fendermint_rpc::client::FendermintClient;
+use fendermint_rpc::query::QueryClient;
+use fendermint_vm_message::query::FvmQueryHeight;
+
+use crate::state::ActorType;
+
+// The `LruCache` is wrapped in `Mutex` beause even reading requiers mutation.
```
fridrik01, line 22:
```suggestion
// The `LruCache` is wrapped in `Mutex` beause even reading requires mutation.
```

fridrik01 (approved):
lgtm

fendermint/eth/api/src/cache.rs
aakoshh: No, it's optional.

fendermint/eth/api/src/mpool.rs
aakoshh: It will, yes, even reads. I thought it might block it less than acquiring the lock hundreds of times, to acquire it once and quickly do all the operations.

PR #844, opened 2024-03-28
ENG-731: Ethereum API transaction buffer (Part 2)

aakoshh:
Depends on https://github.com/consensus-shipyard/ipc/pull/842

Buffer out-of-order transactions sent to the Ethereum API facade until they can be included. 

- [x] Test to show that out-of-order transactions are rejected
- [x] Create a buffer indexing unsent transactions by sender and nonce
- [x] Detect nonce rejection in the facade
- [x] Add rejected transactions to the buffer and return the hash
- [x] Decode the nonce of transactions included in blocks in the `NewBlocks` subscription
- [x] Check if nonces included in the blocks can trigger a chain of new submissions
- [x] Submit transactions when enabled
- [x] Remove transactions included in the blocks from the buffer
- [x] Add a configuration for how much the nonce can be in the future; 0 means buffer is disabled
- [x] Parse the error message for expected and observed nonce
- [x] If the nonce is too far ahead, return the error immediately

### Testing

The tests use the `standalone.yaml` layout (a single node) to deploy transactions with an opposite order of nonces.

```bash
cargo test -p fendermint_materializer --test docker -- out_of_order --nocapture
```


fendermint/eth/api/src/mpool.rs
```diff
@@ -25,22 +27,81 @@ const RETRY_SLEEP_SECS: u64 = 5;
...
+    tx_cache: TransactionCache,
+    tx_buffer: TransactionBuffer,
+) where
+    C: Client + SubscriptionClient + Send + Sync,
+{
     loop {
         let query = Query::from(EventType::NewBlock);
 
-        match client.underlying().subscribe(query).await {
+        match client.subscribe(query).await {
```
cryptoAtwill, line 104: If the Client dies, will there be some kind of restart mechanism or resubscription? Then the pruning might be stoped and out of order txns are not submitted.

fendermint/eth/api/src/mpool.rs
```diff
@@ -25,22 +27,81 @@ const RETRY_SLEEP_SECS: u64 = 5;
...
+                    buffer.remove(&nonce);
+                }
+            }
+        })
+    }
+
+    /// Gather any messages that have been enabled by transactions added to a block.
+    ///
+    /// These are removed from the cache, submission is only attempted once.
+    fn remove_enabled<'a, I>(&self, txs: I) -> Vec<(Address, Nonce, ChainMessage)>
```
cryptoAtwill, line 61: do you think it's a good idea to limit the size of items removed?

fendermint/eth/api/src/mpool.rs
```diff
@@ -25,22 +27,81 @@ const RETRY_SLEEP_SECS: u64 = 5;
...
+    where
+        I: Iterator<Item = (&'a Address, Nonce)>,
+    {
+        self.0.with(|c| {
+            let mut msgs = Vec::new();
+            for (sender, mut nonce) in txs {
+                if let Some(buffer) = c.get_mut(sender) {
+                    nonce += 1;
+                    while let Some(msg) = buffer.remove(&nonce) {
```
cryptoAtwill, line 70: this is more like a depth first removal, not sure if this is fair to other addresses that might have submitted the txn earlier. Actually is fairness a concern at all here.

fendermint/eth/api/src/mpool.rs
```diff
@@ -57,8 +118,22 @@ async fn tx_cache_clearing_loop(
...
+                                }
+
+                                let tx_hashes = txs.iter().map(|(h, _, _)| h);
+                                let tx_nonces = || txs.iter().map(|(_, s, n)| (s, *n));
+
+                                tx_cache.remove_many(tx_hashes);
+                                // First remove all transactions which have been in the block (could be multiple from the same sender).
+                                tx_buffer.remove_many(tx_nonces());
+                                // Then collect whatever is enabled on top of those, ie. anything that hasn't been included, but now can.
+                                let enabled = tx_buffer.remove_enabled(tx_nonces());
```
fridrik01, line 134: `enabled` seems like a boolean, maybe use sth like `enabled_txs`

fridrik01 (commented):
this looks reasonable to me, though @cryptoAtwill has some comments

fendermint/eth/api/src/mpool.rs
aakoshh: Yes if `.next()` returns an `Err` then we come back here and subscribe again. The subscription will cause a new websocket client to be created.

fendermint/eth/api/src/mpool.rs
aakoshh: You are right it is depth first. This was the simplest code I could come up with, which I believe is fine for a first PR. If you have a more elaborate strategy then we can create a ticket for it. 

I think ultimately fairness should be built into the proposal methods, as you won't be able to control how transactions are submitted to the mempool, but there you can order them to your liking.

fendermint/eth/api/src/mpool.rs
aakoshh: Renaming to `unblocked_msgs`

fridrik01 (approved):

PR #845, opened 2024-03-28
CONFIG: Add setting for external addresses for the resolver

aakoshh:
I was under the impression that the `libp2p::Swarm` somehow gets the external [address reported](https://github.com/libp2p/rust-libp2p/blob/1a9cf4f7760724032b729c43165716c7ecd842ad/swarm/src/lib.rs#L1193) to it and it passes it on [to Kademlia](https://github.com/libp2p/rust-libp2p/blob/695404d56bf51cd86d983e3c699d71bdbd9e87d1/protocols/kad/src/behaviour.rs#L2621) but it doesn't seem to work with more involved network setups. 

This PR adds a `FM_RESOLVER__CONNECTION__EXTERNAL_ADDRESSES` setting which accepts a list of `Multiaddr` types.

### DIFF
--- a/fendermint/app/config/default.toml
+++ b/fendermint/app/config/default.toml
@@ -183,6 +183,11 @@ max_provider_age = 300
 # Leaving it empty disables the IPLD Resolver.
 listen_addr = ""
 
+# A list of known external addresses this node is reachable on.
+# If left empty we rely on the `libp2p::Swarm` and the `Identity` protocol to discover it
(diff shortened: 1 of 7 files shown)


folex (approved):
LGTM! We do the same in [Nox](https://github.com/fluencelabs/nox/blob/master/nox/src/node.rs#L512).

fridrik01 (approved):
LGTM

PR #848, opened 2024-03-31
Add a few prometheus metrics for IPC relayer daemon and export them via local endpoint

mb1896:


ipc/provider/src/checkpoint.rs
```diff
@@ -105,6 +138,11 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
     pub async fn run(self, submitter: Address, submission_interval: Duration) {
         log::info!("launching {self} for {submitter}");
 
+        const PROMETHEUS_EXPORTER_URL: &str = "0.0.0.0:9185";
```
raulk, line 141: * Note that this is not a URL, it's a bind address.
* Let's make this a CLI argument to the relayer command (`--prometheus-bind`) with a default value.

ipc/provider/src/checkpoint.rs
```diff
@@ -52,6 +63,28 @@ impl<T: BottomUpCheckpointRelayer> BottomUpCheckpointManager<T> {
...
+                    "checkpoint_submit_succeed_count",
+                    "Number of successful checkpoint submission"
+                )
+                .unwrap(),
+                checkpoint_submit_fail_count: register_counter!(
+                    "checkpoint_submit_fail_count",
+                    "Number of failed checkpoint submission"
+                )
+                .unwrap(),
+            },
```
raulk, line 87: Let's use the same design as in main:

https://github.com/consensus-shipyard/ipc/blob/147180ce386c9d950d04dcbad2a6ca19e3a1ff60/fendermint/app/src/cmd/run.rs#L312

Let's move the `metrics!` macro somewhere common, e.g. inside a new `ipc-metrics` crate under `ipc/metrics`.

Then invoke the macro from within the `checkpoint` module to define the lazy statics.

Then invoke `ipc_provider::checkpoint::register_metrics()` from the relayer command handler.

We may refactor all of this going forward, but it'll be easier if all metrics producers use the same pattern.

ipc/provider/src/checkpoint.rs
```diff
@@ -176,13 +218,20 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                            "submitted bottom up checkpoint({}) in parent at height {}",
+                            event.height,
+                            epoch
+                        );
+                    }
+                    Err(err) => {
+                        self.metrics.checkpoint_submit_fail_count.inc();
+                        return Err(err);
+                    }
+                }
```
raulk, line 234: We'll want to use the best practice for binary outcomes (a total count and an error count): https://promlabs.com/blog/2023/09/19/errors-successes-totals-which-metrics-should-i-expose-to-prometheus/#recommended-for-binary-outcomes-exposing-errors-and-totals

raulk (changes requested):

ipc/provider/src/checkpoint.rs
mb1896: Yes that's exactly I was thinking. Currently the `event` => `metric` binding is specific to fendermint app. If we want to re-use it, we need to refactor existing code to make those macro a shared lib. If that's a more preferred approach I can definitely change it into that.

ipc/provider/src/checkpoint.rs
mb1896: Fixed as you mentioned.

ipc/provider/src/checkpoint.rs
mb1896: Refactored as you mentioned.

ipc/provider/src/checkpoint.rs
mb1896: Thanks. Fixed.

common/metrics/src/prometheus.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+//! Prometheus metrics
+
+#[macro_export]
+macro_rules! metrics {
```
aakoshh, line 6: If we export these then they should be supported with docs and examples. Previously it was just a local macro, so the code itself was there to make it clear how it was working, but without that context it's always helpful to explain how it's supposed to be used.

common/metrics/src/tracing.rs
```diff
@@ -0,0 +1,215 @@
...
+#[macro_export]
+macro_rules! event_mapping {
+    ($op:ident, $event:ident, $event_ty:ident :: $field:ident, $metric:expr) => {
+        $op!($event, $event_ty::$field, $metric);
+    };
+}
+
+/// Match the event name to event DTO types and within the map fields to metrics.
+#[macro_export]
+macro_rules! event_match {
```
aakoshh, line 104: Out of the macros here at least this one needs to be supported by examples in its docstrings. It was highlighted as difficult to understand even when the the usage was right next to it.

common/metrics/src/tracing.rs
```diff
@@ -0,0 +1,215 @@
...
+    ($event:ident { $( $event_ty:ident { $( $field:ident => $op:ident ! $metric:expr  ),* $(,)? } ),* } ) => {
+        println!("jiejie: Got a event with name {}", $event.metadata().name());
+        match $event.metadata().name() {
+            $(
+                event_name!($event_ty) => {
+                    println!("jiejie: Matched event!");
```
aakoshh, line 110: This can be removed.

ipc/event/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+pub type BlockHeight = u64;
+
+#[derive(Debug, Default)]
+pub struct GetLatestAcceptedCheckpoint {
```
aakoshh, line 7: This could have some docs as I have no clue what this event might represent.

ipc/event/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub block_height: BlockHeight,
+}
+
+#[derive(Debug, Default)]
+pub struct SubmitBottomUpCheckpoint {
+    pub block_height: BlockHeight,
+    pub checkpoint_count: u64,
```
aakoshh, line 14: What is a `checkpoint_count`?

ipc/event/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+#[derive(Debug, Default)]
+pub struct SubmitBottomUpCheckpointFail {
```
aakoshh, line 18: Maybe this should carry the reason of failure?

ipc/metrics/src/prometheus.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+//! Prometheus metrics
+
+
+/// Metrics emitted by ipc.
+pub mod app {
```
aakoshh, line 7: Why do we need an `app` module here?

ipc/provider/src/checkpoint.rs
```diff
@@ -232,13 +244,22 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+                    "submitted bottom up checkpoint({}) in parent at height {}",
+                    event.height,
+                    epoch
+                );
+                Ok(())
+            }
+            Err(err) => {
+                emit!(SubmitBottomUpCheckpointFail {
+                    block_height: event.height as BlockHeight,
+                    checkpoint_count: 1,
```
aakoshh, line 259: Why do we need a field that's always 1?

ipc/event/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "ipc_event"
```
aakoshh, line 2: I'm not sure this has to be its own crate. These events could be moved under the `ipc_provider` which emits them.

ipc/metrics/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "ipc_metrics"
```
aakoshh, line 2: I don't think this has to be a crate. IMO it should live under `ipc_cli` which hosts them.

ipc/metrics/src/tracing.rs
```diff
@@ -0,0 +1,54 @@
...
+use metrics_utils::{check_field, event_mapping, event_match, event_name, inc_counter, set_gauge};
+
+/// Create a layer that handles events by incrementing metrics.
+pub fn layer<S>() -> impl Layer<S>
+where
+    S: Subscriber,
+    for<'a> S: LookupSpan<'a>,
+{
+    MetricsLayer::new().with_filter(filter::filter_fn(|md| {
+        md.level() == &Level::INFO && md.name().starts_with("event::")
```
aakoshh, line 24: This logic could be moved to the common crate.

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -68,6 +70,25 @@ impl CommandLineHandler for BottomUpRelayer {
...
+            let tracing_subscriber_registry = tracing_subscriber::registry().with(metrics_layer);
+            tracing::subscriber::set_global_default(tracing_subscriber_registry)
+                .expect("Unable to set a global collector");
+
+            let prometheus_registry = prometheus::Registry::new();
+            ipc_metrics::register_app_metrics(&prometheus_registry)
+                .context("failed to register metrics")?;
+
+            // Start metrics export HTTP server.
+            let addr: SocketAddr = arguments.metric_export_bind_address.parse().unwrap();
```
aakoshh, line 86:
```suggestion
            let addr: SocketAddr = arguments.metric_export_bind_address.parse().expect("failed to parse metrics bind address");
```

Almost never a good idea to just panic without any context.

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -95,4 +116,16 @@ pub(crate) struct BottomUpRelayerArgs {
...
+    #[arg(
+        long,
+        default_value = "true",
+        help = "Flag to turn on/off Prometheus metrics reporting."
+    )]
+    pub enable_metrics: bool,
+    #[arg(
+        long,
+        default_value = "0.0.0.0:9185",
+        help = "Local port for exporting Prometheus metrics."
```
aakoshh, line 128: It's not just a "port". You could either have a host and a port separately, or call it perhaps `metrics_listen_addr` instead.

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -95,4 +116,16 @@ pub(crate) struct BottomUpRelayerArgs {
...
+    )]
+    pub metric_export_bind_address: String,
```
aakoshh, line 130:
```suggestion
    pub metrics_listen_addr: String,
```

I call these `listen_addr` in fendermint. Also note that you called it `metric` here and `metrics` in the other setting.

ipc/cli/src/commands/checkpoint/relayer.rs
```diff
@@ -95,4 +116,16 @@ pub(crate) struct BottomUpRelayerArgs {
         help = "The max parallelism for submitting checkpoints"
     )]
     pub max_parallelism: usize,
+    #[arg(
+        long,
+        default_value = "true",
+        help = "Flag to turn on/off Prometheus metrics reporting."
+    )]
+    pub enable_metrics: bool,
```
aakoshh, line 124:
```suggestion
    pub metrics_enabled: bool,
```

I reckon `--metrics-enabled` and `--metrics-listen-addr` rhymes better. Plus we use `enabled` as a nested property in the fendermint configs.

fendermint/app/Cargo.toml
```diff
@@ -63,6 +62,9 @@ fendermint_vm_resolver = { path = "../vm/resolver" }
 fendermint_vm_snapshot = { path = "../vm/snapshot" }
 fendermint_vm_topdown = { path = "../vm/topdown" }
 
+metrics_utils = { path = "../../common/metrics" }
+tracing_utils = { path = "../../common/tracing" }
+
```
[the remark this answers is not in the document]
aakoshh: It also needs to be added to the CI pipeline for testing.

maciejwitowski:
Closing this PR and we'll followup in the set of smaller, more focused PRs.

PR #838, opened 2024-03-22
ENG-629: Prometheus HTTP

aakoshh:
Serve Prometheus metrics over HTTP in `fendermint run`. 

The default port is 9184. 

The list of metrics is in [prometheus.rs](https://github.com/consensus-shipyard/ipc/blob/7c00889732997cc444cfcd65b95353bf5a9202bc/fendermint/app/src/metrics/prometheus.rs)

TODO: 
- [x] Check that the materializer correctly publishes the port

In future PRs:
* Add more events and metrics to the `fendermint run` part, e.g. snapshots (thought about bottom-up checkpoint submissions but they are opaque)
* Add events and metrics to the `fendermint eth` part, e.g. requests served labeled by the endpoint
* Export metrics from `fendermint eth` (has to be mapped to a different port on the host).

## Where to put the Ethereum API metrics settings?

Because there are two executables, maybe there should be different metrics socket configs be under the `eth` config sections, so the default commands don't clash :thinking: 

I tried moving the `FM_METRICS__*` variables to `FM_ABCI__METRICS__*` but I'm not sure I like that, they are not strictly about ABCI, and currently all other top-level settings are for the ABCI Application, and the `ethapi` only gets the `FM_ETH__*` subsetting. 

So I thought we can just have another `MetricsSetting` under `EthSettings`; it's just might be confusing for people that there are two different `FM_METRICS_..._ENABLE` flags. On the other hand, it's unlikely they'd actually want to touch the defaults. 

Yet another option is that in the `ethapi` we already have a HTTP server, so we can just add a `/metrics` endpoint and serve it from there, rather than `prometheus_exporter`, which is what the project recommends as well. It [wouldn't be a lot of code](https://github.com/AlexanderThaller/prometheus_exporter/blob/9d4120fcef803ec9ccb4dc5a347f4117e85dd8f0/src/lib.rs#L700-L721). For some reason the [CometBFT metrics](https://docs.cometbft.com/v0.37/core/metrics) are on a different port, though.

## Testing

First materialize a subnet so we have some more interesting metrics:

```bash
cargo run -q -p fendermint_app -- \
        materializer \
          --data-dir $PWD/testing/materializer/tests/docker-materializer-data/ \
          setup \
            --manifest-file ./testing/materializer/tests/manifests/layer2.yaml
```

Then we can query the metrics on one of the subnet nodes:

```
❯ curl http://localhost:30384/metrics
# HELP abci_committed_block_height Highest committed block
# TYPE abci_committed_block_height gauge
[...]

PR #849, opened 2024-03-31
[HOLD] Use cargo make to run Prometheus locally

mb1896:
Closes https://github.com/consensus-shipyard/ipc/issues/945

We added `cargo-make` scripts to be able to run Prometheus locally to help Prometheus metrics related development.

Run `cargo make prometheus-run` to start a local Prometheus instance. Then you can visit `http://localhost:9090` for Prometheus UI.

You can see the metrics endpoints that's being scraped by Prometheus.
image

You can also query metric by name and make a graph.

image

This is good enough for testing metrics reporting locally during development. For prod monitoring it's recommended to use Grafana which we will probably integrate with Prometheus later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/849)

### DIFF
--- a/infra/prometheus/Makefile.toml
+++ b/infra/prometheus/Makefile.toml
@@ -1,15 +1,8 @@
-extend = [
-  { path = "scripts/docker.toml" },
-]
-
-[config]
-default_to_workspace = false
-
 [tasks.prometheus-pull]
 script = "docker pull prom/prometheus"
 
 [tasks.prometheus-run]
-
+script = "docker run -p 9090:9090 -v ./prometheus.yml:/etc/prometheus/prometheus.yml prom/prometheus"
 
 [tasks.default]
 clear = true
--- a/infra/prometheus/Makefile.toml
+++ b/infra/prometheus/Makefile.toml
@@ -1,0 +1,160 @@
+extend = [
+  { path = "scripts/docker.toml" },
+  { path = "scripts/cometbft.toml" },
+  { path = "scripts/fendermint.toml" },
+  { path = "scripts/ethapi.toml" },
+  { path = "scripts/genesis.toml" },
+  { path = "scripts/node.toml" },
+  { path = "scripts/promtail.toml" },
+  { path = "scripts/testnet.toml" },
+  { path = "scripts/testnode.toml" },
+  { path = "scripts/subnet.toml" },
+]
+
+[config]
+default_to_workspace = false
+
+[env]
+# General network-specific parameters
+SUBNET_ID = { value = "/r0", condition = { env_not_set = ["SUBNET_ID"] } }
+# The network name is derived from the SUBNET_ID, replacing slashes with dashes, and dropping the first dash if any.
+NETWORK_NAME = { script = ["echo $SUBNET_ID | sed -e 's|/|-|g' -e 's|^-||1'"] }
+# External P2P address advertised by CometBFT to other peers.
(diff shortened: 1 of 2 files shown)


infra/prometheus/Makefile.toml
```diff
@@ -0,0 +1,29 @@
+extend = [
+  { path = "../fendermint/scripts/docker.toml" },
+]
+
+[tasks.prometheus-pull]
+script = "docker pull prom/prometheus"
+
+[tasks.prometheus-start]
+env = { "CONTAINER_NAME" = "Prometheus" }
```
raulk, line 9: Let's call this `ipc-prometheus`.

infra/prometheus/Makefile.toml
```diff
@@ -0,0 +1,29 @@
+extend = [
+  { path = "../fendermint/scripts/docker.toml" },
+]
+
+[tasks.prometheus-pull]
+script = "docker pull prom/prometheus"
```
raulk, line 6: Let's just run this as a dependency of `prometheus-start`. (Although I think docker run will anyway pull from the Docker Registry automatically, so this might not even be needed at all?)

infra/prometheus/prometheus.yml
```diff
@@ -0,0 +1,41 @@
...
+alerting:
+  alertmanagers:
+    - follow_redirects: true
+      enable_http2: true
+      scheme: http
+      timeout: 10s
+      api_version: v2
+      static_configs:
+        - targets: [ ]
+scrape_configs:
```
raulk, line 18: Because this is designed to be used with cargo-make, and our cargo-make scripts use Docker containers, let's use [Docker Prometheus discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#docker_sd_config) instead of static configurations.

Can you please make it work with Fendermint as well, since it's also exporting metrics now? (see https://github.com/consensus-shipyard/ipc/pull/838)

raulk (changes requested):

infra/prometheus/Makefile.toml
mb1896: Removed this task.

infra/prometheus/prometheus.yml
mb1896: We need to make sure all of them live in the same network to make discovery possible. I highly suggest we merge this PR as it is right now to unblock development. What you said can be done as a future task. Do you want me to write a Linear task for that?

raulk: We discussed this in Slack. The goal of ENG-728 is to set up the observability stack for all processes started by cargo-make. That includes Fendermint and CometBFT, there is no reason to single out the relayer.

In fact, cargo-make does not even have a target to spin up a relayer, so I'm not sure what development this would unblock? This PR will not work as it hardcodes assumptions about the port (9185).

Let's do it the right way since the beginning, @mb1896: with Docker service discovery. What exactly are you concerned about re: Docker network? AFAIK, Docker service discovery scans all networks (and adds the network as a label).

raulk (changes requested):
Unaddressed comments.

fridrik01:
Assigned this to me and put into draft

karlem:
@consensus-shipyard/ipc-core-devs I suggest closing this PR and instead using Docker Compose to run Prometheus, as we ideally want to move away from the Cargo script in favor of Docker Compose and Materializer.

karlem:
Closing in favour favour of Docker Compose scripts.

PR #850, opened 2024-04-01
Supporting prefund with token

cryptoAtwill:
This PR attempts to add `prefundWithToken` feature. Some of the key changes are:
1. Adding `prefundWIthToken` contract method
2. Gateway `register` queries the subnet supply source and `lock` the genesisCirculation and collateral. Previously it's using `msg.value`.
3. Propagating the same changes to the gateway: `addStake`, `leave` and `kill`.
4. Adding `join` with supply source

With the above 4 key changes, the tests are passing. But there are other methods to be updated, in SubnetActor: `stake`, `unstake`, `leave` and `kill`. This will be updated in a follow up PR.


cryptoAtwill:
no longer needed as we are refactoring the contracts.

PR #851, opened 2024-04-02
fundToken Inflation / Deflation / Tax on transfer Fix

snissn:
closes ENG-713

- [x] In `SupplySourceHelper#lock`, take a reading of our token balance before calling safeTransferFrom. Take another reading after. 
- [x] Calculate the difference between before and after, and return it.
- [x] Use that in GatewayManagerFacet#fundWithToken as the value to mint in the subnet.
- [x] In the gateway, make sure to run a 0 check, as the effectively transferred amount may be 0. In this case, ~skip sending the top down message.~ revert 

NOTE the suggested behavior on the previous item was to skip the top down message, but instead this PR executes a revert. This seemed to be the more appropriate behavior from my perspective on best practices. LMK if i am missing something and will remove the revert and skip the top down message in the gateway

- [x] IMPORTANT: Make sure to use safe arithmetic operations, as an underflow here would be fatal!

See [this thread](https://filecoinproject.slack.com/archives/C065VJDDSJH/p1711370060434969) for a doubt I am checking with Zokyo.

Add tests for these scenarios:

- [x] transfer is inflationary
- [x] transfer is deflationary
- [x] transfer has nil effect (effectively transferred value is 0)

### DIFF
--- a/contracts/src/gateway/GatewayManagerFacet.sol
+++ b/contracts/src/gateway/GatewayManagerFacet.sol
@@ -170,8 +170,9 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         SupplySource memory supplySource = SubnetActorGetterFacet(subnetId.getActor()).supplySource();
         supplySource.expect(SupplyKind.ERC20);
 
-        // Lock the specified amount into custody.
-        // Account for tokens with Transfer Fees
+        // Locks a specified amount into custody, adjusting for tokens with transfer fees. This operation
+        // accommodates inflationary tokens, potentially reflecting a higher effective locked amount.
+        // Operation reverts if the effective transferred amount is zero.
         uint256 transferAmount = supplySource.lock({value: amount});
 
         // Create the top-down message to mint the supply in the subnet.
--- a/contracts/src/gateway/GatewayManagerFacet.sol
+++ b/contracts/src/gateway/GatewayManagerFacet.sol
@@ -171,14 +171,15 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         supplySource.expect(SupplyKind.ERC20);
 
         // Lock the specified amount into custody.
-        supplySource.lock({value: amount});
+        // Account for tokens with Transfer Fees
+        uint256 transferAmount = supplySource.lock({value: amount});
 
         // Create the top-down message to mint the supply in the subnet.
         IpcEnvelope memory crossMsg = CrossMsgHelper.createFundMsg({
             subnet: subnetId,
             signer: msg.sender,
             to: to,
-            value: amount
+            value: transferAmount
         });
 
         // Commit top-down message.
--- a/contracts/src/lib/SupplySourceHelper.sol
+++ b/contracts/src/lib/SupplySourceHelper.sol
@@ -13,6 +13,7 @@ library SupplySourceHelper {
     using SafeERC20 for IERC20;
 
     error InvalidERC20Address();
+    error NoBalanceIncrease();
     error UnexpectedSupplySource();
     error UnknownSupplySource();
 
@@ -46,12 +47,19 @@ library SupplySourceHelper {
     }
 
     /// @notice Locks the specified amount sent by the msg.sender into custody.
-    function lock(SupplySource memory supplySource, uint256 value) internal {
+    function lock(SupplySource memory supplySource, uint256 value) internal returns (uint256) {
         if (supplySource.kind == SupplyKind.ERC20) {
             IERC20 token = IERC20(supplySource.tokenAddress);
+            uint256 initialBalance = token.balanceOf(address(this));
             token.safeTransferFrom({from: msg.sender, to: address(this), value: value});
+            uint256 finalBalance = token.balanceOf(address(this));
+            if (!(finalBalance > initialBalance)) {
+                revert NoBalanceIncrease();
+            }
+            return finalBalance - initialBalance;
         }
         // Do nothing for native.
+        return value;
     }
 
     /// @notice Transfers the specified amount out of our treasury to the recipient address.
--- a/contracts/src/lib/SupplySourceHelper.sol
+++ b/contracts/src/lib/SupplySourceHelper.sol
@@ -46,7 +46,9 @@ library SupplySourceHelper {
         }
     }
 
-    /// @notice Locks the specified amount sent by the msg.sender into custody.
+    /// @notice Locks the specified amount from msg.sender into custody.
+    ///         Reverts with NoBalanceIncrease if the token balance does not increase.
+    ///         May return more than requested for inflationary tokens due to balance rise.
     function lock(SupplySource memory supplySource, uint256 value) internal returns (uint256) {
         if (supplySource.kind == SupplyKind.ERC20) {
             IERC20 token = IERC20(supplySource.tokenAddress);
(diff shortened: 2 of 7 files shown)


contracts/src/lib/SupplySourceHelper.sol
```diff
@@ -46,12 +47,19 @@ library SupplySourceHelper {
...
 
     /// @notice Locks the specified amount sent by the msg.sender into custody.
-    function lock(SupplySource memory supplySource, uint256 value) internal {
+    function lock(SupplySource memory supplySource, uint256 value) internal returns (uint256) {
         if (supplySource.kind == SupplyKind.ERC20) {
             IERC20 token = IERC20(supplySource.tokenAddress);
+            uint256 initialBalance = token.balanceOf(address(this));
             token.safeTransferFrom({from: msg.sender, to: address(this), value: value});
+            uint256 finalBalance = token.balanceOf(address(this));
+            if (!(finalBalance > initialBalance)) {
```
raulk, line 56: Removing the negation makes it easier to read:

```suggestion
            if (finalBalance <= initialBalance) {
```

contracts/src/lib/SupplySourceHelper.sol
```diff
@@ -46,12 +47,19 @@ library SupplySourceHelper {
...
+                revert NoBalanceIncrease();
+            }
+            return finalBalance - initialBalance;
```
raulk, line 59: Can you leave a comment here that no safe arithmetic is necessary because no underflow is possible due to the check above?

contracts/src/gateway/GatewayManagerFacet.sol
```diff
@@ -171,14 +171,15 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         supplySource.expect(SupplyKind.ERC20);
 
         // Lock the specified amount into custody.
-        supplySource.lock({value: amount});
+        // Account for tokens with Transfer Fees
+        uint256 transferAmount = supplySource.lock({value: amount});
```
raulk, line 175: Let's document that if the transfer amount is zero, this will have reverted.

contracts/src/lib/SupplySourceHelper.sol
```diff
@@ -46,12 +47,19 @@ library SupplySourceHelper {
     }
 
     /// @notice Locks the specified amount sent by the msg.sender into custody.
```
raulk, line 49: Let's document the conditions under which this can revert (`NoBalanceIncrease`).

contracts/src/gateway/GatewayManagerFacet.sol
raulk: Let's also document that this can return a higher effectively locked amount if the token is inflationary.

contracts/src/lib/SupplySourceHelper.sol
raulk: Also mention that this can return a higher amount that requested if the token is inflationary.

contracts/src/gateway/GatewayManagerFacet.sol
raulk, line 175: According to Zokyo, the `transferAmount` should be clamped by the originally requested amount. So if the transfer happens to lock more tokens than desired, we would still mint and transfer the exact amount of tokens requested by the user.

However, this will lead to locked funds:

* The user deposits 100.
* The inflationary/rewarding token ends up transferring 150 to us.
* We only recognise 100 and mint that amount to the user in the subnet.
* We increment the circulating supply by 100, not 150.
* Those extra 50 will not be accounted for in the protocol and they will end up locked forever in treasury.

I'm checking with Zokyo again, because I don't think their recommendation was valid for our protocol.

contracts/test/integration/MultiSubnet.t.sol
```diff
@@ -49,12 +54,16 @@ contract MultiSubnetTest is Test, IntegrationTestBase {
...
     TestSubnetDefinition public nativeSubnet;
     TestSubnetDefinition public tokenSubnet;
+    TestSubnetDefinition public deflationaryTokenSubnet;
+    TestSubnetDefinition public inflationaryTokenSubnet;
+    TestSubnetDefinition public nilTokenSubnet;
 
     IERC20 public token;
+    IERC20 public deflationaryToken;
+    IERC20 public inflationaryToken;
+    IERC20 public nilToken;
```
raulk, line 64: Do we need these as state variables? We only use each in its own test, once, so I think each one of these could be local variables in their corresponding test case.

contracts/test/integration/MultiSubnet.t.sol
```diff
@@ -308,6 +334,110 @@ contract MultiSubnetTest is Test, IntegrationTestBase {
...
+        vm.prank(caller);
+        vm.expectRevert(); // half of tokens burned on transfer
+        rootSubnet.gateway.manager().fundWithToken(
+            deflationaryTokenSubnet.id,
+            FvmAddressHelper.from(address(caller)),
+            amount
+        );
+
+        vm.prank(caller);
+        rootSubnet.gateway.manager().fundWithToken(
```
raulk, line 361: Shouldn't we assert that: (a) the circ supply matches the deflated amount, and (b) the gateway's token balance matches the deflated amount. Same for the inflationary and nil test cases.

contracts/src/lib/SupplySourceHelper.sol
snissn: added comment here 320cadd9

contracts/src/gateway/GatewayManagerFacet.sol
snissn: 521cb344

contracts/src/lib/SupplySourceHelper.sol
snissn: 8f397bcd

contracts/src/gateway/GatewayManagerFacet.sol
snissn: My current understanding based on the latest communications from Zokyo that we are following their recommended protocol by taking the difference of token balances before and after the transfer. Summarized 

```
So the actual amount transferred into the protocol will be actual = after - before
Which should account for both inflationary and deflationary tokens on transfer
```

Does that sound right?

raulk: Agreed. Sounds right.

contracts/test/integration/MultiSubnet.t.sol
raulk: @snissn ^^

contracts/test/integration/MultiSubnet.t.sol
raulk: @snissn ^^

contracts/test/integration/MultiSubnet.t.sol
snissn: I think that we should put these variables in the setUp in case we want to re-create more tests around these three tokens. Also the setUp requires a few variables that aren't saved as state, so we would need to place rootGateway and rootSubnetName into state at least in order to set up token subnets for them calling `createTokenSubnet`.

contracts/test/integration/MultiSubnet.t.sol
snissn: Thanks, that makes sense, adding

raulk (approved):

raulk:
Feel free to merge. The test failure is fixed on main.

PR #854, opened 2024-04-02
Update "Deploy Linked Token" GH action to support Proxy upgrade pattern

snissn:
This PR adds support for proxy upgrades to the linked_token_deployment branch which is now in Main.

Currently the check makes it to `make deploy-replica-proxy` and then fails with the error:
```
err: (code: 2, message: expected sequence 11, got 12, data: Some(String("")))
```

next step to investigate that is to see if the account sending transactions related to the linked token is separate from an ipc validator and to make it its own funded account if it is not

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/854)


maciejwitowski:
@snissn what is the plan for this PR? I assume it also supersedes https://github.com/consensus-shipyard/ipc/pull/789 which could be closed, right?

snissn:
@maciejwitowski that assumption is correct. I was optimistically working on this and it's also turning into a large time sync. The code has a dependency on a server that I don't know the details about, takes a long time to run so it's hard to test, and seemed to have quirks that I haven't gotten deep into where it would be flaky - I think due to multiple runs of tests happening a time. For now I had just left it but do not have a clear plan because of the cost benefit analysis on how much time to fix vs how much its worth fixing.

karlem:
@consensus-shipyard/ipc-core-devs I would suggest to close the issue as it is not something critical and it does not work.

PR #855, opened 2024-04-02
extras/linked-token: add prettier + make fmt.

raulk:
This uses the same prettier config as for `contracts`.


snissn (approved):
thanks! Prettier consistency is so helpful

PR #860, opened 2024-04-03
Top-down syncer: Adding more logs to help debug

cryptoAtwill:
Adding more logs to help debug.

### DIFF
--- a/fendermint/vm/interpreter/src/chain.rs
+++ b/fendermint/vm/interpreter/src/chain.rs
@@ -173,12 +173,16 @@ where
                     // 2) if a validator proposes something invalid, we can make them pay during execution.
                     let is_resolved =
                         atomically(|| match env.checkpoint_pool.get_status(&item)? {
-                            None => Ok(false),
+                            None => {
+                                tracing::debug!("no status found");
+                                Ok(false)
+                            },
                             Some(status) => status.is_resolved(),
                         })
                         .await;
 
                     if !is_resolved {
+                        tracing::debug!("proposal rejected by checkpoint pool");
                         return Ok(false);
                     }
                 }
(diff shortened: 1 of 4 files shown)


raulk (commented):
I agree in principle with adding more logs. But now that we resolved the issue, I suggest we spend a bit more time adding more thoughtful messages, events, and metrics.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -173,12 +173,16 @@ where
                     // 2) if a validator proposes something invalid, we can make them pay during execution.
                     let is_resolved =
                         atomically(|| match env.checkpoint_pool.get_status(&item)? {
-                            None => Ok(false),
+                            None => {
+                                tracing::debug!("no status found");
```
aakoshh, line 177: As always it would be great to do logging outside `atomically` as it can be restarted. Maybe by tweaking the returned value. The message could also be a bit more descriptive, as so many things can have a status.
We aren't using these bottom-up checkpoints right now anyway.

aakoshh: But I see debug logs added to all the other STM methods, so it's probably moot.

fendermint/vm/topdown/src/voting.rs
```diff
@@ -291,11 +291,17 @@ where
     pub fn set_finalized(&self, block_height: BlockHeight, block_hash: V) -> Stm<()> {
         self.chain.update(|chain| {
             let (_, mut chain) = chain.split(&block_height);
-            chain.insert(block_height, Some(block_hash));
+            chain.insert(block_height, Some(block_hash.clone()));
+            tracing::debug!(
+                block_height,
+                block_hash = hex::encode(block_hash),
+                "inserted into vote tally chain"
```
aakoshh, line 298:
```suggestion
                "finalized block inserted into vote tally chain"
```

aakoshh (approved):
I see voting has been thoroughly sprinkled with logs at this point. I hope it helped, and if it did then perhaps this will, too.

jsoares (approved):

PR #861, opened 2024-04-03
Launching abci server only when parent syncer has loaded enough blocks in cache

cryptoAtwill:
For some special cases, there might be a race condition for proposal processing and parent syncer data ingestion. 

One such scenario would be cometbft WAL replay. If somehow fendermint and cometbft crashes,  when restart, cometbft might replay the voting proposals with parent finality height. At this moment, fendermint has no time to sync with the parent node and proposals will be rejected. This PRs enforces that during initialization, only when the parent syncer has synced `max_proposal_range` number of blocks, should abci be available.

Closes ENG-841.

### DIFF
--- a/fendermint/vm/topdown/src/sync/mod.rs
+++ b/fendermint/vm/topdown/src/sync/mod.rs
@@ -111,6 +111,7 @@ pub async fn launch_polling_syncer<T, C, P>(
     vote_tally: VoteTally,
     parent_client: Arc<P>,
     tendermint_client: C,
+    tx: tokio::sync::oneshot::Sender<bool>,
 ) -> anyhow::Result<()>
 where
     T: ParentFinalityStateQuery + Send + Sync + 'static,
@@ -154,6 +155,7 @@ where
         parent_client,
         query,
         tendermint_client,
+        tx,
     );
 
     Ok(())
@@ -167,6 +169,7 @@ fn start_syncing<T, C, P>(
     parent_proxy: Arc<P>,
     query: Arc<T>,
     tendermint_client: C,
+    tx: tokio::sync::oneshot::Sender<bool>,
 ) where
     T: ParentFinalityStateQuery + Send + Sync + 'static,
     C: tendermint_rpc::Client + Send + Sync + 'static,
@@ -176,11 +179,19 @@ fn start_syncing<T, C, P>(
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
 
     tokio::spawn(async move {
+        let max_proposal_range = config.max_proposal_range();
+
         let lotus_syncer =
             LotusParentSyncer::new(config, parent_proxy, view_provider, vote_tally, query)
                 .expect("");
 
         let mut tendermint_syncer = TendermintAwareSyncer::new(lotus_syncer, tendermint_client);
+
+        if let Err(e) = tendermint_syncer.init(max_proposal_range).await {
+            tracing::error!("cannot init tendermint syncer: {e}");
+            let _ = tx.send(false);
+            return;
+        }
 
         loop {
             interval.tick().await;
--- a/fendermint/vm/topdown/src/sync/syncer.rs
+++ b/fendermint/vm/topdown/src/sync/syncer.rs
@@ -13,6 +13,7 @@ use anyhow::anyhow;
 use async_stm::{atomically, atomically_or_err, StmError};
 use ethers::utils::hex;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 use tracing::instrument;
 
 use fendermint_tracing::emit;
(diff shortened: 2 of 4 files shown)


fendermint/vm/topdown/src/sync/mod.rs
```diff
@@ -176,12 +179,20 @@ fn start_syncing<T, C, P>(
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
 
     tokio::spawn(async move {
+        let max_proposal_range = config.max_proposal_range();
+
         let lotus_syncer =
             LotusParentSyncer::new(config, parent_proxy, view_provider, vote_tally, query)
                 .expect("");
```
aakoshh, line 186: This `expect` should have a non-empty message.

fendermint/vm/topdown/src/sync/mod.rs
```diff
@@ -176,12 +179,20 @@ fn start_syncing<T, C, P>(
...
 
         let mut tendermint_syncer = TendermintAwareSyncer::new(lotus_syncer, tendermint_client);
 
+        if let Err(e) = tendermint_syncer.init(max_proposal_range).await {
+            tracing::error!("cannot init tendermint syncer: {e}");
+            let _ = tx.send(false);
+            return;
+        }
```
aakoshh, line 194: Maybe I'm blind, but I can't see where `true` is sent.

fendermint/vm/topdown/src/sync/mod.rs
```diff
@@ -176,12 +179,20 @@ fn start_syncing<T, C, P>(
...
     tokio::spawn(async move {
+        let max_proposal_range = config.max_proposal_range();
+
         let lotus_syncer =
             LotusParentSyncer::new(config, parent_proxy, view_provider, vote_tally, query)
                 .expect("");
 
         let mut tendermint_syncer = TendermintAwareSyncer::new(lotus_syncer, tendermint_client);
 
+        if let Err(e) = tendermint_syncer.init(max_proposal_range).await {
```
aakoshh, line 190: Can you add commments as to why we need `max_proposal_range` blocks to be pre-filled in the cache?

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -58,6 +59,51 @@ where
...
+            // obtain the finalized chain head first
+            let chain_head = loop {
+                if let Some(h) = self.finalized_chain_head().await? {
+                    break h;
+                }
+
+                tracing::debug!(
+                    "syncer not initialized yet as chain head not obtained, sleep 5 sec"
+                );
+                tokio::time::sleep(Duration::from_secs(5)).await;
```
aakoshh, line 85: Move to a top level const? I don't think we need to mention in the comments how long it's going to sleep, it's bound to get out of sync this way.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -58,6 +59,51 @@ where
...
+        let (mut latest_height, mut non_null_parent_hash) = self.latest_cached_data().await;
+        let target_height = cached_blocks + latest_height + 1;
+        tracing::debug!(
+            target_height,
+            latest_height,
+            "syncer init with last committed finality"
+        );
+
+        loop {
+            // obtain the finalized chain head first
```
aakoshh, line 76: It would be helpful to leave a comment to whether this is going to the parent API or the local ledger.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -58,6 +59,51 @@ where
...
+
+        let (mut latest_height, mut non_null_parent_hash) = self.latest_cached_data().await;
+        let target_height = cached_blocks + latest_height + 1;
+        tracing::debug!(
+            target_height,
+            latest_height,
+            "syncer init with last committed finality"
+        );
+
+        loop {
```
aakoshh, line 75: What does this `loop` achieve?

raulk:
@aakoshh when CometBFT restarts, it replays the [WAL](https://docs.cometbft.com/v0.38/spec/consensus/wal) during its early start-up routine by calling Fendermint's ABCI API. If the WAL contains a block proposal with top-down finality (TopDownExec), we end up here:

https://github.com/consensus-shipyard/ipc/blob/4091b89d00d3735646f8672cc0e274e0b226db18/fendermint/vm/interpreter/src/chain.rs#L194

Which in turn ends up here:

https://github.com/consensus-shipyard/ipc/blob/2d3c76d130d780a3d81f0bfa62822a528933877b/fendermint/vm/topdown/src/finality/null.rs#L322

And ultimately results in u s rejecting the block because Fendermint was still not ready. This edge case uncovered a deeper race condition. We were able to uncover it thanks to the premature execution resulting from the early WAL replay, but which can equally happen if CometBFT starts syncing faster than Fendermint initializes.

The idea here is to suspend exposing the ABCI API until the system is actually ready to accept calls.

cryptoAtwill:
No longer needed with topdown refactoring.

PR #867, opened 2024-04-05
Add a configuration number getter

cryptoAtwill:
Add a configuration numbers getter in the top down facet. These numbers track the configuration number applied in the child gateway.


snissn (approved):
This change makes sense to me! The addition of external view functions looks correct and valid. It would be nice to see a test for the additional getters functions. If we are currently pressed for time, could we at least document what specific tests you would like to see in the future as we merge this PR? Thanks! Looks good to me

contracts/src/gateway/GatewayGetterFacet.sol
```diff
@@ -20,6 +21,12 @@ contract GatewayGetterFacet {
     using EnumerableSet for EnumerableSet.AddressSet;
     using EnumerableSet for EnumerableSet.Bytes32Set;
 
+    /// @notice Get the configuration numbers in the validator changes. The configuration numbers
```
jsoares, line 24:
```suggestion
    /// @notice Returns the configuration numbers in the validator changes. The configuration numbers
```

Nit, just for consistency.

contracts/src/gateway/router/TopDownFinalityFacet.sol
```diff
@@ -35,6 +35,15 @@ contract TopDownFinalityFacet is GatewayActorModifiers {
         s.validatorsTracker.batchStoreChange(changeRequests);
     }
 
+    /// @notice Get the configuration numbers in the validatorTracker struct that shadows the configuration changes
```
jsoares, line 38: Nit: the docs leak internal structure and don't actually state/explain the return values

jsoares (approved):

contracts/src/gateway/GatewayGetterFacet.sol
```diff
@@ -20,6 +21,12 @@ contract GatewayGetterFacet {
...
+    /// obtained are from changes made in the parent.
```
jsoares, line 25:
```suggestion
    /// @notice Returns the next and start configuration numbers in the validator changes. 
    /// The configuration numbers are from changes made in the parent.
```

Nit: consistency and clarity

contracts/src/gateway/router/TopDownFinalityFacet.sol
```diff
@@ -35,6 +35,15 @@ contract TopDownFinalityFacet is GatewayActorModifiers {
...
+    /// from the parent in the child gateway
```
jsoares, line 39:
```suggestion
    /// @notice Get the next and start configuration numbers in the tracker of changes
    /// from the parent in the child gateway
```

Nit: clarity (explain 2 return values), abstraction (avoid leaking internal structure). It's still not super clear but probably makes sense in context.

PR #870, opened 2024-04-09
ENG-842: Fetch data at height concurrently

aakoshh:
A small change in `fetch_data` to query the two event logs concurrently, which might bring down the ~2.5s query time to something closer to 1s.


aakoshh:
Merged as part of https://github.com/consensus-shipyard/ipc/pull/871

PR #871, opened 2024-04-09
ENG-842: Export topdown events

aakoshh:
Adds a CLI command to fetch topdown events of a subnet and export them to JSON. These could be read into an upgrade and applied on the child subnet.

As an alternative, one can use the Filfox API: https://filfox.info/api/v1/address/f410fqpww4v74jydq25jncdbletyqd42oxyeoapzaz4a/events?page=0&pageSize=5

### Usage

```console
❯ cargo run -q -p fendermint_app -- debug ipc export-top-down-events --help
Fetch topdown events from the parent and export them to JSON.

This can be used to construct an upgrade to impute missing events.

Usage: fendermint debug ipc export-top-down-events [OPTIONS] --subnet-id <SUBNET_ID> --parent-endpoint <PARENT_ENDPOINT> --start-block-height <START_BLOCK_HEIGHT> --end-block-height <END_BLOCK_HEIGHT> --events-file <EVENTS_FILE>

Options:
  -s, --subnet-id <SUBNET_ID>
          Child subnet for with the events will be fetched

  -p, --parent-endpoint <PARENT_ENDPOINT>
          Endpoint to the RPC of the child subnet's parent

      --parent-auth-token <PARENT_AUTH_TOKEN>
          HTTP basic authentication token

      --parent-gateway <PARENT_GATEWAY>
          IPC gateway of the parent; 20 byte Ethereum address in 0x prefixed hex format
          
          [default: 0xff00000000000000000000000000000000000064]

      --parent-registry <PARENT_REGISTRY>
          IPC registry of the parent; 20 byte Ethereum address in 0x prefixed hex format
          
          [default: 0xff00000000000000000000000000000000000065]

      --start-block-height <START_BLOCK_HEIGHT>
          The first block to query for events

      --end-block-height <END_BLOCK_HEIGHT>
          The last block to query for events

      --events-file <EVENTS_FILE>
          Location of the JSON file to write events to

  -h, --help
          Print help (see a summary with '-h')

```

### Rationale

I added this under `debug ipc` because although it's very similar to `genesis ipc from-parent`, it doesn't have anything to do with the genesis per-se, and all `genesis` commands require a `--genesis-file` argument. The expectation is that we won't need this; the only reason for it to exist is to reduce the chance of manual errors in replaying events as part of an upgrade, but that's something that we ideally don't have to do. For this reason it's under a new `debug` command family, rather than an `ipc` one, to show that it's not a normal operation.

I didn't want to put it into the `ipc-cli` because I wanted to reuse the `fetch_data` in the `fendermint` project, and because the only reason for this to exist is to support a (one-off :crossed_fingers:) upgrade which happens in fendermint.

### Example

Trying to find events in the range of the following query:

```console
⟩ curl -X POST --location "https://gerovit.filmine.dev/rpc/v1" \
          -H "Content-Type: application/json" \
          -d '{"id": 1, "jsonrpc": "2.0", "method": "eth_getLogs", "params": [{"fromBlock": "0x39a454", "toBlock": "0x39a459"}]}'
{"jsonrpc":"2.0","result":[],"id":1}
```

```shell
cargo run  -p fendermint_app -- debug ipc export-top-down-events \
            --subnet-id /r314/f410fqpww4v74jydq25jncdbletyqd42oxyeoapzaz4a \
            --parent-endpoint <RPC_ENDPOINT> \
            --parent-auth-token <RPC_AUTH_TOKEN> \
            --parent-gateway 0x5cFA791d5c1b162216309fba6fc2A68758A3b233 \
            --parent-registry 0xe0B7D47e3f0443e8DC6cA6F84c75F65Cb3a29676 \
            --start-block-height 3777620 \
            --end-block-height 3777625 \
            --events-file topdown.json
```

See which configuration numbers were exported from https://node.glif.io/fvm-archive/lotus/rpc/v1:

```console
❯ cat topdown.json | jq -c ".[] | .[1][1][] | .configuration_number"
55
56
57
58
59
60
61
62
63
64
65
66

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5060,7 +5060,6 @@ dependencies = [
  "ipc-types",
  "ipc-wallet",
  "ipc_actors_abis",
- "log",
  "num-derive 0.3.3",
  "num-traits",
  "reqwest",
@@ -5075,6 +5074,7 @@ dependencies = [
  "tokio",
  "tokio-tungstenite 0.18.0",
  "toml 0.8.10",
+ "tracing",
  "url",
  "zeroize",
 ]
--- a/fendermint/app/options/src/debug.rs
+++ b/fendermint/app/options/src/debug.rs
@@ -1,0 +1,67 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::path::PathBuf;
+
+use crate::parse::parse_eth_address;
+use clap::{Args, Subcommand};
+use fvm_shared::address::Address;
+use ipc_api::subnet_id::SubnetID;
+
+#[derive(Args, Debug)]
+pub struct DebugArgs {
+    #[command(subcommand)]
+    pub command: DebugCommands,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum DebugCommands {
+    /// IPC commands.
+    Ipc {
+        #[command(subcommand)]
+        command: DebugIpcCommands,
+    },
+}
+
+#[derive(Subcommand, Debug, Clone)]
+pub enum DebugIpcCommands {
+    /// Fetch topdown events from the parent and export them to JSON.
+    ///
+    /// This can be used to construct an upgrade to impute missing events.
+    ExportTopDownEvents(Box<DebugExportTopDownEventsArgs>),
+}
+
+#[derive(Args, Debug, Clone)]
+pub struct DebugExportTopDownEventsArgs {
+    /// Child subnet for with the events will be fetched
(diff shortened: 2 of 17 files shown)


cryptoAtwill (approved):
LGTM

jsoares (approved):
tested, works fine

aakoshh:
It would be nice to make the format human readable. @cryptoAtwill I thought I saw you copy the `IsHumanReadable` stuff from fendermint to somewhere, but can't find it anywhere. Maybe it's in a pending PR somewhere. If this gets more use, it would be worth adding.

PR #873, opened 2024-04-10
Topdown consistency checks: Validate number of top down messages

cryptoAtwill:
This is the start of a series of updates to enhance topdown finality robustness and accuracy. 

This PR adds a simple check to ensure the number of top down messages received is actually correct. The purpose of this is that after running a few tests, some RPC nodes, especially when querying historical data, might not return events completely. Adding this check will alert this from happening and let throws an error/stops parent syncer for node operator to perform checks.

As the gateway already exposes the applied top down nonce by subnet, one just needs to query the consecutive historical state to deduce the number of top down messages.

Draft: updating unit tests.

### DIFF
--- a/fendermint/vm/topdown/src/finality/fetch.rs
+++ b/fendermint/vm/topdown/src/finality/fetch.rs
@@ -334,6 +334,13 @@ mod tests {
                 block_hash: r.0,
             })
         }
+
+        async fn get_top_down_nonce(
+            &self,
+            _block_hash: &[u8],
+        ) -> anyhow::Result<TopDownQueryPayload<u64>> {
+            todo!()
+        }
     }
 
     fn new_provider(
--- a/fendermint/vm/topdown/src/finality/mod.rs
+++ b/fendermint/vm/topdown/src/finality/mod.rs
@@ -88,6 +88,13 @@ mod tests {
                 block_hash: vec![],
             })
         }
+
+        async fn get_top_down_nonce(
+            &self,
+            _block_hash: &[u8],
+        ) -> anyhow::Result<TopDownQueryPayload<u64>> {
+            todo!()
+        }
     }
 
     fn mocked_agent_proxy() -> Arc<MockedParentQuery> {
--- a/fendermint/vm/topdown/src/sync/syncer.rs
+++ b/fendermint/vm/topdown/src/sync/syncer.rs
@@ -234,6 +234,8 @@ where
         }
 
         let data = self.fetch_data(height, block_hash_res.block_hash).await?;
+        self.check_data(height, &data, &block_hash_res.parent_block_hash)
+            .await?;
 
         tracing::debug!(
             height,
@@ -264,6 +266,44 @@ where
         });
 
         Ok(data.0)
+    }
+
+    #[instrument(skip(self))]
+    async fn check_data(
+        &self,
+        height: BlockHeight,
+        current_view: &ParentViewPayload,
+        parent_block_hash: &[u8],
+    ) -> Result<(), Error> {
+        let prev_nonce = self.top_down_nonce(parent_block_hash).await?;
+        let curr_nonce = self.top_down_nonce(&current_view.0).await?;
+
+        let num_msgs = curr_nonce - prev_nonce;
+        let msgs = &current_view.2;
+        if num_msgs != msgs.len() as u64 {
+            return Err(Error::TopDownMsgsLengthIncorrect(
+                height,
+                num_msgs,
+                current_view.2.len() as u64,
+            ));
+        }
+
+        Ok(())
+    }
+
+    #[inline]
+    async fn top_down_nonce(&self, block_hash: &[u8]) -> Result<u64, Error> {
+        Ok(self
+            .parent_proxy
+            .get_top_down_nonce(block_hash)
+            .await
+            .map_err(|e| {
+                Error::CannotQueryParentHash(
+                    format!("top down nonce: {e}"),
+                    hex::encode(block_hash),
+                )
+            })?
+            .value)
     }
 
     #[instrument(skip(self))]
@@ -428,6 +468,13 @@ mod tests {
                 value: vec![],
                 block_hash: self.blocks.get_value(height).cloned().unwrap().unwrap(),
             })
+        }
+
+        async fn get_top_down_nonce(
+            &self,
+            _block_hash: &[u8],
+        ) -> anyhow::Result<TopDownQueryPayload<u64>> {
+            todo!()
         }
     }
 
--- a/ipc/cli/src/commands/crossmsg/mod.rs
+++ b/ipc/cli/src/commands/crossmsg/mod.rs
@@ -13,6 +13,7 @@ use fund::FundArgs;
 use propagate::PropagateArgs;
 use release::ReleaseArgs;
 
+use crate::commands::crossmsg::topdown_cross::{GetTopDownMsgNonce, GetTopDownMsgNonceArgs};
 use clap::{Args, Subcommand};
 
 pub mod fund;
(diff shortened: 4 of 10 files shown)


fendermint/vm/topdown/src/finality/fetch.rs
```diff
@@ -334,6 +334,13 @@ mod tests {
                 block_hash: r.0,
             })
         }
+
+        async fn get_top_down_nonce(
```
cryptoAtwill, line 338: Unit test, will update them accordingly

fendermint/vm/topdown/src/finality/fetch.rs
```diff
@@ -334,6 +334,13 @@ mod tests {
...
+            &self,
+            _block_hash: &[u8],
+        ) -> anyhow::Result<TopDownQueryPayload<u64>> {
+            todo!()
```
cryptoAtwill, line 342: Unit test, will update them accordingly

fendermint/vm/topdown/src/finality/mod.rs
```diff
@@ -88,6 +88,13 @@ mod tests {
                 block_hash: vec![],
             })
         }
+
+        async fn get_top_down_nonce(
+            &self,
+            _block_hash: &[u8],
+        ) -> anyhow::Result<TopDownQueryPayload<u64>> {
```
cryptoAtwill, line 95: Unit test, will update them accordingly

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -266,6 +268,44 @@ where
         Ok(data.0)
     }
 
+    #[instrument(skip(self))]
+    async fn check_data(
+        &self,
+        height: BlockHeight,
+        current_view: &ParentViewPayload,
+        parent_block_hash: &[u8],
```
aakoshh, line 276: Isn't there an alias for block hash at least?

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -266,6 +268,44 @@ where
...
+    ) -> Result<(), Error> {
+        let prev_nonce = self.top_down_nonce(parent_block_hash).await?;
```
aakoshh, line 278: Could this be preserved from the previous iteration?

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -429,6 +469,13 @@ mod tests {
                 block_hash: self.blocks.get_value(height).cloned().unwrap().unwrap(),
             })
         }
+
+        async fn get_top_down_nonce(
+            &self,
+            _block_hash: &[u8],
+        ) -> anyhow::Result<TopDownQueryPayload<u64>> {
+            todo!()
```
cryptoAtwill, line 477: Unit test, will update them accordingly

aakoshh (commented):
This makes sense to me :+1: 

I was wondering about two things:
1. Whether it's enough to only do this check, or do we have to check the configuration number as well. The answer is that because messages can be empty where the validator changes are not, we have to check both if we want to catch RPCs that don't serve data as soon as possible. 
2. Whether it's enough to do this check in the syncer, or should it also happen during catching up with others, ie. when the block is executed. The answer is that as long as others did the checking before voting on finality, if during execution we do not do the check, but have an RPC without history, we'll get a consensus failure on the _next_ block, when the application hash difference is revealed. Not sure how easy it is to recover from that. 

Random thoughts: 
1. Perhaps if someone is confident in the RPC, they can make this check optional in the settings and save some time by not executing extra queries. 
2. Perhaps the nonces could be added to the IPC finality the votes are about, but that would require major refactoring in the voting.

fendermint/vm/topdown/src/sync/syncer.rs
cryptoAtwill: updated

cryptoAtwill:
@aakoshh For:
```
1. Perhaps if someone is confident in the RPC, they can make this check optional in the settings and save some time by not executing extra queries.
```
With the past few days, I would say we just check this regardless, I have fewer trust in RPC nodes now. Even if the RPC is not missing any events, it could joined a partitioned network.

```
2.Perhaps the nonces could be added to the IPC finality the votes are about, but that would require major refactoring in the voting.
```
Yeah, I think this came up a couple of times, but mostly in bottom up checkpoint. But we still haven't seen a very strong case for the refactoring.

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -215,7 +221,7 @@ where
 
                     // Null block received, no block hash for the current height being polled.
                     // Return the previous parent hash as the non-null block hash.
-                    return Ok(parent_block_hash);
+                    return Ok((parent_block_hash, None));
```
aakoshh, line 224: Why not return `parent_block_nonce` to not lose the previous context?

ipc/cli/src/commands/crossmsg/mod.rs
```diff
@@ -52,5 +54,6 @@ pub(crate) enum Commands {
     PreRelease(PreReleaseArgs),
     Propagate(PropagateArgs),
     ListTopdownMsgs(ListTopdownMsgsArgs),
+    ListTopdownNonce(GetTopDownMsgNonceArgs),
```
aakoshh, line 57: Should this be `GetTopdownNonce`?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -845,6 +846,36 @@ impl SubnetManager for EthSubnetManager {
...
+        let evm_subnet_id = gateway_getter_facet::SubnetID::try_from(subnet)?;
+        let (exists, nonce) = gateway_contract
+            .get_top_down_nonce(evm_subnet_id)
+            .block(BlockId::Hash(H256::from_slice(block_hash)))
+            .call()
+            .await?;
+
+        let nonce = if !exists {
+            log::warn!("subnet does not exists at block hash, return 0 nonce");
+            0
```
aakoshh, line 869: Is there a particular reason why you opt to return 0 with a warning in the logs, instead of returning `None` or an `Err`? I looked at other cases in the file, and it looks in most other cases an error is returned: 
* [genesis_epoch](https://github.com/consensus-shipyard/ipc/blob/4110c41e575cb33bddeadf606049c5b2f5bf4f1a/ipc/provider/src/manager/evm/manager.rs#L99-L100) 
* [get_applied_topdown_nonce](https://github.com/consensus-shipyard/ipc/blob/4110c41e575cb33bddeadf606049c5b2f5bf4f1a/ipc/provider/src/manager/evm/manager.rs#L931-L932), 
* [prev_bottom_up_checkpoint_hash](https://github.com/consensus-shipyard/ipc/blob/4110c41e575cb33bddeadf606049c5b2f5bf4f1a/ipc/provider/src/manager/evm/manager.rs#L966)
* [get_block_hash](https://github.com/consensus-shipyard/ipc/blob/4110c41e575cb33bddeadf606049c5b2f5bf4f1a/ipc/provider/src/manager/evm/manager.rs#L170)

Because `get_block_hash` returns an `Err`, it means `get_validator_changeset` and `get_topdown_msgs` do so as well. You can argue that by the time you call this it will most certainly not return an error since it would have already failed on the previous two, but still, for consistency I think it should behave the same way.

raulk:
Recording here that @cryptoAtwill encountered an unexpected roadblock because of this: https://github.com/filecoin-project/lotus/issues/11205. It turns out that `eth_call` in Lotus behaves non-deterministically depending on how the node operator configured this env variable: `LOTUS_SKIP_APPLY_TS_MESSAGE_CALL_WITH_GAS`.

On Glif nodes, it's configured to skip the execution of requested block prior to applying the call, resulting in the call being executed on the **prestate** of the requested height, instead of the **poststate** (which is the general expectation).

This PR attempts to solve this: https://github.com/filecoin-project/lotus/pull/11905, but it seems to be in some form of limbo.

aarshkshah1992:
@raulk That PR needs a review from someone who knows the ETH RPC stack really well but Steb is currently fully focused on F3. Would you or Mikers or Fridrick have the time to review it ?

PR #874, opened 2024-04-10
Several improvements to deploy subnet script

fridrik01:
- Fix relayer not starting due to missing argument
- Fix bug when default-key was first in the json array
- Streamline local deployment by introducing a prepare_local.sh script
- Improve documentation for local deployment and document more known issues
- Fix Mac support

Closes ENG-849
Closes ENG-857

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/874)


scripts/deploy_subnet_under_calibration_net/prepare_local.sh
```diff
@@ -0,0 +1,50 @@
...
+set -eo pipefail
+
+SCRIPT_DIR=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+echo "SCRIPT_DIR: $SCRIPT_DIR"
+
+echo "Installing dependencies..."
+cargo install toml-cli
+
+echo "Removing existing .ipc folder"
+rm -rf ~/.ipc
```
aakoshh, line 22: This seems a bit brutal. Should it perhaps make a backup, if it doesn't exist yet?

scripts/deploy_subnet_under_calibration_net/prepare_local.sh
```diff
@@ -0,0 +1,50 @@
...
+
+echo "Copying new .ipc folder"
+cp -r $SCRIPT_DIR/.ipc ~/.ipc
+
+echo "Fetching info about newest current deployment contracts"
+hash=$(curl -s https://raw.githubusercontent.com/consensus-shipyard/ipc/cd/contracts/deployments/r314159.json)
+echo $hash | jq
```
aakoshh, line 29: Why do you call this a `hash`? Is it supposed to be a "HashTable"? I associate it with an actual hash. How about "deployment"?

scripts/deploy_subnet_under_calibration_net/prepare_local.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Parsing gateway and registry address..."
+gateway_addr=$(echo $hash | jq -r '.gateway_addr')
+echo "- Gateway address: $gateway_addr"
+registry_addr=$(echo $hash | jq -r '.registry_addr')
+echo "- Registry address: $registry_addr"
+
+echo "Updating config with new gateway and registry address"
+toml set ~/.ipc/config.toml subnets[0].config.gateway_addr $gateway_addr > ~/.ipc/config.toml.tmp
+toml set ~/.ipc/config.toml.tmp subnets[0].config.registry_addr $registry_addr > ~/.ipc/config.toml.tmp2
+cp ~/.ipc/config.toml.tmp2 ~/.ipc/config.toml
```
aakoshh, line 40: Can it not do an in-place update, or overwrite itself? It seems error prone to have to create a new file in each step. In any case, do they have to stick around?

aakoshh (commented):
I am not familiar with these deployment scripts, I have to defer to someone else here.

scripts/deploy_subnet_under_calibration_net/prepare_local.sh
aakoshh: Coming from Ruby?

fridrik01: haha, yeah coming from Ruby, I can rename it to something else

scripts/deploy_subnet_under_calibration_net/prepare_local.sh
fridrik01: > This seems a bit brutal. Should it perhaps make a backup, if it doesn't exist yet?

Maybe call it `reset_local_.sh` and/or add a prompt to confirm to user that it will reset your ipc folder?

scripts/deploy_subnet_under_calibration_net/prepare_local.sh
fridrik01: I believe toml-cli only outputs the modified toml version, see https://github.com/gnprice/toml-cli?tab=readme-ov-file#toml-set. But yeah can cleanup the temp files.

scripts/deploy_subnet_under_calibration_net/prepare_local.sh
raulk: The upsert trick from @aakoshh would work for the ipc-cli, but this folder is also storing validator config and keys. @fridrik01 alternatively create a timestamped folder?

raulk (changes requested):
Pinged @fridrik01 for action here.

How are we planning to move this PR forward? I think we've established that rm'ing the `.ipc` folder is undesirable. Can we rotate it instead? (I liked @aakoshh’s suggestion, but it only solves part of the problem.)

fridrik01:
@raulk I updated the PR so it renames the ipc folder (to include a timestamp) when the script runs so nothing gets deleted

scripts/deploy_subnet_under_calibration_net/README.md
```diff
@@ -4,16 +4,16 @@
...
-3. Run `bash deploy.sh local` to deploy IPC locally.
+1. Run `prepare_local.sh` to generate a local ipc repo
+2. Go to [Calibration faucet](https://faucet.calibration.fildev.network/funds.html) to fund all of your addresses.
+3. Run  `deploy.sh local` to deploy IPC locally.
 
 Please also notice that
-1. The `deploy.sh` is only for running on Linux. If you are using a Mac, you need to disable all `apt` based dependency installation. You may also need to install bash (version >= 5) to run this script since the script isn't fully compatible with zsh (default shell on Mac).
+1. The `deploy.sh` is only for running on Linux. If you are using a Mac, you need to disable all `apt` based dependency installation. You may also need to install bash (version >= 5) to run this script since the script isn't fully compatible with zsh (default shell on Mac). Also, make sure you habe openssl packages installed (`brew install openssl@3`).
```
jsoares, line 16:
```suggestion
1. The `deploy.sh` is only for running on Linux. If you are using a Mac, you need to disable all `apt` based dependency installation. You may also need to install bash (version >= 5) to run this script since the script isn't fully compatible with zsh (default shell on Mac). Also, make sure you have the openssl packages installed (`brew install openssl@3`).
```

karlem:
This PR is being closed in favor of our endeavor to use Docker Compose to run a local network.

PR #876, opened 2024-04-10
METRICS: Additional events around voting and quroum

aakoshh:
Adds the following events:
* `ParentFinalityVoteAdded`: when a new vote is received and added to the tally
* `ParentFinalityVoteIgnored`: when a new vote is received but isn't added to the tally, which could be because the height is already final
* `ParentFinalityMissingQuorum`: when the proposer knows about a new block which it thinks is final, but can't find the quorum for it in the tally

Based on these the following new metrics are defined:
* `TOPDOWN_FINALITY_VOTE_BLOCK_HEIGHT`: The highest block we received a vote for
* `TOPDOWN_FINALITY_VOTE_ADDED`: The number of votes added
* `TOPDOWN_FINALITY_VOTE_IGNORED`: The number of votes ignored
* `TOPDOWN_FINALITY_MISSING_QUORUM`: The number of occasions we had something to propose but couldn't

Based on these we could have alerts on the rate of the growth in the counters. 

Under normal circumstances we'd expect to receive votes for each parent block from each validator, so the _added+ignored_ should grow steadily, however there are some caveats:
* if there was a Byzantine peer it could cast votes without being a validator, which would skew _ignored_
* as blocks are finalized, the latecomer votes are go _ignored_, so _added_ is expected to grow on a statistically average rate, but not by a predictable number in each height, because a particular node might receive the a block containing the finality sooner than it sees a quorum of votes over gossip

The PR also adds an option to emit events with DEBUG level, so these votes don't spam the INFO level logs.

### DIFF
--- a/fendermint/app/src/metrics/tracing.rs
+++ b/fendermint/app/src/metrics/tracing.rs
@@ -128,7 +128,8 @@ impl<S: Subscriber> Layer<S> for MetricsLayer<S> {
                 block_height              => set_gauge   ! &am::TOPDOWN_FINALIZED_BLOCK_HEIGHT,
             },
             ParentFinalityVoteAdded {
-                block_height              => max_gauge    ! &am::TOPDOWN_FINALITY_VOTE_BLOCK_HEIGHT,
+                block_height              => max_gauge    ! &am::TOPDOWN_FINALITY_VOTE_MAX_BLOCK_HEIGHT,
+                block_height              => set_gauge    ! &am::TOPDOWN_FINALITY_VOTE_LAST_BLOCK_HEIGHT,
                 validator                 => inc1_counter ! &am::TOPDOWN_FINALITY_VOTE_ADDED,
             },
             ParentFinalityVoteIgnored {
--- a/fendermint/app/src/metrics/tracing.rs
+++ b/fendermint/app/src/metrics/tracing.rs
@@ -128,8 +128,12 @@ impl<S: Subscriber> Layer<S> for MetricsLayer<S> {
                 block_height              => set_gauge   ! &am::TOPDOWN_FINALIZED_BLOCK_HEIGHT,
             },
             ParentFinalityVoteAdded {
+                // This one can move up and down randomly as votes come in, but statistically should
+                // be less likely to be affected by Byzantine validators casting nonsense votes.
+                block_height              => set_gauge    ! &am::TOPDOWN_FINALITY_VOTE_BLOCK_HEIGHT,
+                // This one should only move up, showing the highest vote in the tally.
+                // It should be easy to produce this on Grafana as well from the one above.
                 block_height              => max_gauge    ! &am::TOPDOWN_FINALITY_VOTE_MAX_BLOCK_HEIGHT,
-                block_height              => set_gauge    ! &am::TOPDOWN_FINALITY_VOTE_LAST_BLOCK_HEIGHT,
                 validator                 => inc1_counter ! &am::TOPDOWN_FINALITY_VOTE_ADDED,
             },
             ParentFinalityVoteIgnored {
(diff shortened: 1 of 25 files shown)


fendermint/app/src/metrics/tracing.rs
```diff
@@ -57,6 +55,17 @@ macro_rules! set_gauge {
     };
 }
 
+/// Set a gague to the maximum of its value and a field in an event.
+macro_rules! max_gauge {
```
folex, line 59: There's usually no point in filtering data, as it is easily done in Prometheus / Mimir / Grafana.

And it could be problematic as it becomes not just a pure stream of data as is, but a complex filtered thing. To interpret it, one would need to take this Rust impl into account, which complicates things a lot if you're just operating Fendermint without being acknowledged with its metric reporting code base.

For example, if it would be `cmp::min`, then after restart `gauge.get()` would return zero, and it would always be zero. 

Here it is `cmp::max`, so the problem with default value after restart is not so pronounced, but you cannot keep this invariant of "never set values lower than before. Imagine, after the Fendermint restart, the gauge value will become 0, and `max` will return true.

So metrics collector could observe eg the following values `1 2 3 (restart) 2 3 4 5 6`, so this invariant no longer holds, as this is an in-memory local-only storage, it can't know the data in Prometheus.

The usual rule of thumb is to submit values as raw as possible, as to never hide any anomalies, and aggregate or normalise on Prometheus side.

fendermint/app/src/metrics/tracing.rs
```diff
@@ -108,6 +127,16 @@ impl<S: Subscriber> Layer<S> for MetricsLayer<S> {
             ParentFinalityCommitted {
                 block_height              => set_gauge   ! &am::TOPDOWN_FINALIZED_BLOCK_HEIGHT,
             },
+            ParentFinalityVoteAdded {
+                block_height              => max_gauge    ! &am::TOPDOWN_FINALITY_VOTE_BLOCK_HEIGHT,
```
folex, line 131: What if this height goes backwards? I think it would be nice to see that in Prometheus / Grafana. Suggest to lift the `max` filtering, unless there's a reason that I miss.

fridrik01 (approved):
This looks fine to me, but comments from @folex sound reasonable its often typical to do the aggregation in the UI but don't really have a strong opinion either way

fendermint/app/src/metrics/tracing.rs
aakoshh: The reason is that votes arrive unordered through gossip. I can receive a vote from Alice about block 123, then a block from Bob about block 99. There's no point jumping up and down as they come randomly, that would make this rather meaningless IMO. You can pick those raw values up from the log if you want to see them. This gauge is merely an indication of what is the highest block we see votes on, so it should, in general, go up, indicating that the group is making progress through the gossiping of votes.

fendermint/app/src/metrics/tracing.rs
aakoshh: To be clear it was never my intention to promise an invariant that this can never go down, only that this is the highest value received since the node was started.

fendermint/app/src/metrics/tracing.rs
aakoshh: Okay, I added both versions with comments. I know you will say it doesn't make sense to keep the max version because you can produce it in Grafana, but in the past I found it convenient to just use `curl` and see what I'm interested in. Grafana will naturally only take a sample of the observations, while a metric for the highest vote received is durable and won't be masked by a followup vote on a lower value. 

In the end I agree that the `set_gauge` version is useful because at the moment if a single Byzantine validator casts a vote on some stupid high value, it can lock the `max` metric for good. Now that will be easy to see and will indicate foul play. On Grafana it would have been down to luck to query the node at the time where the `set` version shows a high value.

fendermint/app/src/metrics/tracing.rs
aakoshh: And yes, a `min` version wouldn't make sense with a metric such as this, and without an exception made for 0.

PR #877, opened 2024-04-10
Add an early getSubnet() exists check in fundWithToken()

snissn:
closes ENG-851

Uses same check code as in fund()


raulk (approved):

raulk (approved):
I believe I left this out deliberately to save gas, as this check is already done later anyway when committing the top-down message inside `LibGateway#commitTopDownMsg()`. 

However, in hindsight, it's best to be defensive and conduct an explicit check, as the implementation of `LibGateway#commitTopDownMsg()` could evolve to remove the check in the future, and we wouldn't notice here.

PR #880, opened 2024-04-12
ENG-854: Limit consensus concurrency to 1

aakoshh:
Adds a `tower` service layer that limits the concurrent requests going to the `consensus` connection to 1, which should hopefully solve the issue of transactions being applied in a different order than they appear in the block. This is the actual fix for a [problem we had with end_block](https://github.com/consensus-shipyard/ipc/commit/4b2c90b2cf053c85d6ff684e4ae7d6b228c31eb9). 

### Buffering

A buffer is added so that the other channels aren't blocked while `consensus` isn't ready. See [here](https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html#order) for example. 

To be honest I don't understand this - this limit only affects the `Connection`, but a single `Connection` instance effectively only serves one kind of request, because CometBFT opens 4 different sockets to the service. By the looks of it it should wait, and buffering should not be required, yet the end-to-end tests failed when there wasn't a buffer :eyes: 

Ah, one reason for this is is that if there is no `buffer`, then while `Connection` is waiting to submit the next request [here](https://github.com/penumbra-zone/tower-abci/blob/dc2d5efee8cd748b4d816563002c3f569d12b55a/src/v037/server.rs#L240), it cannot process the asynchronously executing responses [here](https://github.com/penumbra-zone/tower-abci/blob/dc2d5efee8cd748b4d816563002c3f569d12b55a/src/v037/server.rs#L274). So the reason the tests might fail is that if CometBFT sends multiple requests in a row (which it does, pipelined) such as `[begin, deliver_tx, end, commit]` for the deployment that failed, then this happens:
1. Receive `deliver_tx`, take a permit, [put it in the response](https://github.com/tower-rs/tower/blob/39adf5c509a1b2141f679654d8317524ca96b58b/tower/src/limit/concurrency/service.rs#L92)
2. Receive `end`, wait for the permit
3. Now, even though `deliver_tx` is ready, it cannot be processed by `Connection::run` because it's blocked waiting for the permit, therefore it won't free the permit inside the `Response`, and we have a deadlock. 

By adding a buffer, we unblock 2, which unblocks 3, and replenishes the permit. 

Thus it is important that the buffer size is at least as much as the maximum number of messages in a block, so the settings have been extended with a default limit of 1000, which is enforced in `prepare_proposal` and `process_proposal`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2913,6 +2913,7 @@ dependencies = [
  "tendermint-proto 0.31.1",
  "tendermint-rpc",
  "tokio",
+ "tower",
  "tower-abci",
  "tracing",
  "tracing-appender",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -214,6 +214,7 @@ frc42_dispatch = "6.0.0"
 
 # Using the same tendermint-rs dependency as tower-abci. From both we are interested in v037 modules.
 tower-abci = { version = "0.7" }
+tower = { version = "0.4" }
 tendermint = { version = "0.31", features = ["secp256k1"] }
 tendermint-config = "0.33.0"
 tendermint-rpc = { version = "0.31", features = [
--- a/fendermint/abci/examples/kvstore.rs
+++ b/fendermint/abci/examples/kvstore.rs
@@ -4,12 +4,17 @@
 
 use async_stm::{atomically, TVar};
 use async_trait::async_trait;
-use fendermint_abci::{AbciResult as Result, Application, ApplicationService};
+use fendermint_abci::{
+    util::take_until_max_size, AbciResult as Result, Application, ApplicationService,
+};
 use structopt::StructOpt;
 use tendermint::abci::{request, response, Event, EventAttributeIndexExt};
 use tower::ServiceBuilder;
 use tower_abci::{split, v037::Server};
 use tracing::{info, Level};
+
+// For the sake of example, sho the relationship between buffering, concurrency and block size.
+const MAX_TXNS: usize = 100;
 
 /// In-memory, hashmap-backed key-value store ABCI application.
 ///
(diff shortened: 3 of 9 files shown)


raulk (commented):
Won't this throttle us in dangerous ways? Mempool tx validation, Eth API calls will all be limited to zero parallelism? 

IUC BeginBlock/DeliverTx/EndBlock is the only critical section that induces mutability on our end with perfect ordering assumptions? If so, I'd consider guarding and linearising those operations only. 

We could do this in application space by tracking and monitoring the Futures we return, and/or activating this middleware on BeginBlock, disabling it on EndBlock, and only preventing state-mutating ABCI ops during the critical section (eg CheckTx and queries would be allowed to run).

Alternatively we could just migrate to CometBFT 0.38, but we'd need provide a clean upgrade path for existing users (and a hotfix branch until they're able to upgrade).

raulk:
Or is that what the split effectively does? 🤔

aakoshh:
@raulk yes, `split` creates 4 different services on top of a single common service to control their individual policies, for example by [rate limiting](https://github.com/penumbra-zone/tower-abci/blob/dc2d5efee8cd748b4d816563002c3f569d12b55a/examples/kvstore_37/main.rs#L168-L183) the mempool and query channels, without affecting the others.

raulk (approved):
These are the ABCI calls covered by the Consensus service:

https://github.com/informalsystems/tendermint-rs/blob/adf235f6251dc32956af10b7656fe7d4ee139166/tendermint/src/v0_34/abci/request.rs#L55-L59

aakoshh:
Opened an issue to raise awareness of this in the library https://github.com/penumbra-zone/tower-abci/issues/44

PR #881, opened 2024-04-15
DEBUG: Print response on JSON-RPC query errors

aakoshh:
Help debug some serde parsing issue with gas estimations.

### DIFF
--- a/fendermint/rpc/Cargo.toml
+++ b/fendermint/rpc/Cargo.toml
@@ -13,6 +13,7 @@ base64 = { workspace = true }
 bytes = { workspace = true }
 prost = { workspace = true }
 serde = { workspace = true }
+serde_json = { workspace = true }
 tendermint = { workspace = true }
 tendermint-rpc = { workspace = true }
 tendermint-proto = { workspace = true }
--- a/fendermint/rpc/src/client.rs
+++ b/fendermint/rpc/src/client.rs
@@ -266,10 +266,97 @@ where
     let height: u64 = height.into();
     let height = Height::try_from(height).context("failed to conver to Height")?;
 
+    // This is how we'd call it, but here we're trying to debug what's going on using
+    // the `perform` method below with a request that prints the response if it fails
+    // to deserialize for any reason.
+    // let res = client
+    //     .abci_query(None, data, Some(height), false)
+    //     .await
+    //     .context("abci query failed")?;
+
+    let req = tendermint_rpc::endpoint::abci_query::Request::new(None, data, Some(height), false);
+
     let res = client
-        .abci_query(None, data, Some(height), false)
+        .perform(debug::DebugRequest(req))
         .await
-        .context("abci query failed")?;
+        .context("abci query failed")?
+        .0
+        .response;
 
     Ok(res)
+}
+
+mod debug {
+    use serde::{Deserialize, Serialize};
+    use tendermint_rpc as trpc;
+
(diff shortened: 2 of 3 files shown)


fridrik01 (approved):
lgtm

aakoshh:
For the record after adding this PR the problem revealed itself in the following way. 

On the client the error was unchanged:
```
message: 'failed to call estimate gas query: estimate gas query failed: abci query failed: serde parse error\n' +
        '\n' +
        'Caused by:\n' +
        '    expected value at line 1 column 1\n' +
        '\n' +
        'Location:\n' +
        '    /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9: serde parse error: expected value at line 1 column 1'
```

This came through an `ethapi` that connected to a load balancer, which routed queries to cometbft nodes.

In the `ethapi` logs we saw the body that failed to parse:
```
2024-04-16T14:42:57.466868Z ERROR fendermint/rpc/src/client.rs:335: failed to parse JSON-RPC response error="serde parse error\n\nCaused by:\n    expected value at line 1 column 1\n\nLocation:\n    /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9" response="upstream connect error or disconnect/reset before headers. reset reason: connection termination"
```

Trying locally the error was slightly different, pointing exactly at the problem:

```
message: 'failed to call estimate gas query: estimate gas query failed: abci query failed: HTTP error\n' +
        '\n' +
        'Caused by:\n' +
        '    connection closed before message completed\n' +
        '\n' +
        'Location:\n' +
        '    /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9: HTTP error: connection closed before message completed'
```

We were given extra context that the call for which they were estimating gas was big, took over 8s to complete, which could have caused a timeout. Trying it locally with an artificial 15s sleep added to any estimation, the same problem occurred in the e2e tests:

```
Error: failed to call eth_estimateGas w/ height: (code: 0, message: failed to call estimate gas query: estimate gas query failed: abci query failed: HTTP error

Caused by:
    connection closed before message completed

Location:
    /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9: HTTP error: connection closed before message completed, data: None)
```

After some digging around in CometBFT code I found [these timeouts](https://github.com/cometbft/cometbft/blob/7d230d7ddbcf0fbbd154f7b973bc3cef8aa818bc/rpc/jsonrpc/server/http_server.go#L44-L45) in the HTTP server, and the [way we can increase them](https://github.com/cometbft/cometbft/blob/cce2e5da5b7f54aed1fe6dd36c14404ad941d714/node/node.go#L1281-L1285) was found:

```
# How long to wait for a tx to be committed during /broadcast_tx_commit.
# WARNING: Using a value larger than 10s will result in increasing the
# global HTTP write timeout, which applies to all connections and endpoints.
# See https://github.com/tendermint/tendermint/issues/3435
timeout_broadcast_tx_commit = "10s"
```

That is, the `CMT_RPC_TIMEOUT_BROADCAST_TX_COMMIT` env var needs to be adjusted upwards if HTTP requests are expected to take longer than 10s.

PR #886, opened 2024-04-18
fix aderyn false positives

snissn:
Fixes ENG-860
Fixes three errors found by Aderyn checker among two types of error: encodePacked and Unprotected initializer

### encodePacked
**Error**:

files : `contracts/src/lib/SubnetIDHelper.sol` and `contracts/test/helpers/FvmAddressHelper.sol`
error message: `Found high vulnerability: abi.encodePacked() should not be used with dynamic types when passing the result to a hash function such as keccak256()` 

**Solution**: 

changed abi.encodePacked to string.concat 
### Unprotected initializer
**Error**:
`src/lib/LibDiamond.sol`
`Unprotected initializer at src/lib/LibDiamond.sol line 204`

**Solution**: 
Modify the bash script contracts/tools/check_aderyn.sh that allow lists specific filename, error name and line number tuples. 
```
IGNORE_SPECIFIC["src/lib/LibDiamond.sol:203:Unprotected initializer"]=1
IGNORE_SPECIFIC["src/lib/LibDiamond.sol:204:Unprotected initializer"]=1
```
**Note:** because of differences in the script running on my machine and the github action, there are two allow list entries for line 203 and line 204. I do not understand the origin of the discrepancy at the moment

### DIFF
--- a/contracts/src/lib/SubnetIDHelper.sol
+++ b/contracts/src/lib/SubnetIDHelper.sol
@@ -42,7 +42,7 @@ library SubnetIDHelper {
     }
 
     function toString(SubnetID calldata subnet) public pure returns (string memory) {
-        string memory route = string(abi.encodePacked("/r", Strings.toString(subnet.root)));
+        string memory route = string.concat("/r",Strings.toString(subnet.root));
         uint256 subnetLength = subnet.route.length;
         for (uint256 i; i < subnetLength; ) {
             route = string.concat(route, "/");
--- a/contracts/src/lib/SubnetIDHelper.sol
+++ b/contracts/src/lib/SubnetIDHelper.sol
@@ -42,7 +42,7 @@ library SubnetIDHelper {
     }
 
     function toString(SubnetID calldata subnet) public pure returns (string memory) {
-        string memory route = string.concat("/r",Strings.toString(subnet.root));
+        string memory route = string.concat("/r", Strings.toString(subnet.root));
         uint256 subnetLength = subnet.route.length;
         for (uint256 i; i < subnetLength; ) {
             route = string.concat(route, "/");
--- a/contracts/tools/check_aderyn.sh
+++ b/contracts/tools/check_aderyn.sh
@@ -11,6 +11,10 @@ SEVERITIES=(critical high medium)
 # List of vulnerability titles to ignore
 IGNORE_TITLES=("Centralization Risk for trusted owners")
 
+# Specific vulnerabilities to ignore with path and line number
+declare -A IGNORE_SPECIFIC
+IGNORE_SPECIFIC["src/lib/LibDiamond.sol:204:Unprotected initializer"]=1
+
 containsElement() {
   local e match="$1"
   shift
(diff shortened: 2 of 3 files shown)


contracts/src/lib/SubnetIDHelper.sol
```diff
@@ -42,7 +42,7 @@ library SubnetIDHelper {
     }
 
     function toString(SubnetID calldata subnet) public pure returns (string memory) {
-        string memory route = string(abi.encodePacked("/r", Strings.toString(subnet.root)));
+        string memory route = string.concat("/r",Strings.toString(subnet.root));
```
raulk, line 45:
```suggestion
        string memory route = string.concat("/r", Strings.toString(subnet.root));
```

raulk: Was Aderyn complaining about this, or is it just an opportunistic optimization?

contracts/tools/check_aderyn.sh
```diff
@@ -11,6 +11,10 @@ SEVERITIES=(critical high medium)
 # List of vulnerability titles to ignore
 IGNORE_TITLES=("Centralization Risk for trusted owners")
 
+# Specific vulnerabilities to ignore with path and line number
+declare -A IGNORE_SPECIFIC
+IGNORE_SPECIFIC["src/lib/LibDiamond.sol:204:Unprotected initializer"]=1
```
raulk, line 16: This line number is certainly going to change :-/ Does the Aderyn report not emit a symbol that we can match against?

raulk (commented):
Could you please document this PR with a list of concrete errors we're trying to fix?

Also note that the build fails.

snissn:
> Could you please document this PR with a list of concrete errors we're trying to fix? Also note that the build fails.
Yes! Thank you and apologies - I'm confused I had definitely written up an analysis I'm not sure what happened to it. I will clean this up

contracts/tools/check_aderyn.sh
snissn: I don't see any convenient way for allow listing a specific failure. Aderyn provides a way to ignore a specific file, but that feels like a way to set ourselves up for problems down the line

snissn: > This line number is certainly going to change

I'm curious what you mean by that because I'm seeing a really confusing pattern at the moment. On my desktop the error claims to be for line 204 but the github action claims the error is on line 203. 

also because this specific error that i'm filtering is in the file `src/lib/LibDiamond.sol` I think that is something we will touch relatively infrequently

contracts/src/lib/SubnetIDHelper.sol
snissn: Aderyn was in fact complaining about this - here is the latest run https://github.com/consensus-shipyard/ipc/actions/runs/8782737964/job/24097651471 - when this error is fixed, the test file that I changed also shows the same error, and when that second error is fixed there is a 3rd reported error related to the diamond initialize function

cryptoAtwill:
@snissn is the `Unprotected initializer` related to https://forum.openzeppelin.com/t/unsecured-initializer-uups/29080? If yes, I think add the:
```
/// @custom:oz-upgrades-unsafe-allow constructor
constructor() {
    _disableInitializers();
}
``` 
would be a better approach.

snissn:
@cryptoAtwill I think that aderyn doesn't support the inline comments similar to `/// @custom:oz-upgrades-unsafe-allow constructor` I tried to add that line to my local copy and running aderyn still gives the error

cryptoAtwill (approved):

raulk (approved):
The specific line references are rather unfortunate, but I don't see another way out at the moment.

PR #891, opened 2024-05-03
bugfix: ipc-cli: Remove invalid fvm keystore logging 

snissn:
Remove debug logs when the fvm keystore is not encountered. The logging was creating confusion because it was appearing in the common use cases of ipc-cli when the FVM wallet is not expected to exist.

closes ENG-848


raulk (commented):
This isn't the right way to solve this, as it trades off cases where the log is relevant for cases where it's not relevant.

PR #892, opened 2024-05-03
Fix cicd lint warning

cryptoAtwill:
Fix the cicd lint warnings.


ipc/wallet/src/evm/mod.rs
```diff
@@ -7,11 +7,15 @@ mod memory;
...
 
 use anyhow::Result;
-use std::{hash::Hash, str::FromStr};
+use std::fmt::{Display, Formatter};
+use std::hash::Hash;
 use zeroize::Zeroize;
 
 pub use crate::evm::persistent::{PersistentKeyInfo, PersistentKeyStore};
 
+#[cfg(feature = "with-ethers")]
```
raulk, line 16: What does this do exactly?

fendermint/app/settings/src/lib.rs
```diff
@@ -111,13 +113,16 @@ pub enum DbCompaction {
...
-            .as_str()
-            .expect("compaction is a string")
-            .to_string()
+impl Display for DbCompaction {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            serde_json::to_value(self)
+                .expect("compaction serializes to JSON")
```
raulk, line 122: Let's return an error instead of panicking, now that we can.

raulk (approved):

PR #894, opened 2024-05-07
docs(fendermint): explain config in README.

aakoshh:
Adds sections to the fendermint README about configuration.


raulk (approved):
LGTM for now. We should remove the nested README.md and move this info to the operator/users/dev manual.

fendermint/README.md
```diff
@@ -61,3 +61,47 @@ The image can be pulled with the following command:
...
+`fendermint` can be configured using either a configuration file, or using environmnet variables.
+
+### Config files
+
+The default configuration is in [default.toml](./app/config/default.toml). This file is copied into the `fendermint` docker image and should not be edited, so that any further releases can provide any new keys with default values necessary for the application to start; instead the operator can provide further partial configuration files to override the defaults.
+
+The [Settings::config](./app/settings/src/lib.rs) method expects the following files in a configuration directory:
+* `default.toml` with the settings for the keys that have meaningful default values
+* `<mode>.toml` is an optional file correponding to the `--mode` CLI option (by default `dev`); these are files that could be checked into Github e.g. to have values for production or test environments
+* `local.toml` is also optional with potentia overrides specific to the developer's machine
```
raulk, line 78:
```suggestion
* `local.toml` is also optional with potential overrides specific to the developer's machine
```

raulk, line 78: I don't find these naming choices usable. `default` and `local` are too fungible with `<mode>`. In reality, they are "special" files and should be signalled as such.

In lieu of introducing a formal order/hierarchy-based cascading system, I'd rather have:
- special files prefixed with `_`
- `local` replaced with `_overrides`, which is a more intentional name (what does "local" mean? it could reasonably be a mode name)

All in all, this conveys more intuitive info than what we have now:

* `_default.toml`
* `<mode>.toml`
* `_overrides.toml`

raulk (approved):

fendermint/README.md
raulk: https://github.com/consensus-shipyard/ipc/issues/1028

Issue #211, opened 2023-09-07
IPC: Checkpoint construction in the child subnet


aakoshh:
This is the other half of consensus-shipyard/ipc-monorepo#231 dealing with the construction and relaying of a checkpoint from the child subnet to the parent. 

Unlike consensus-shipyard/ipc-monorepo#231 which is only relevant if Fendermint itself has child subnets, this one is needed for Fendermint acting as a subnet of the Lotus rootnet. 

```[tasklist]
### Tasks
- [ ] https://github.com/consensus-shipyard/fendermint/issues/244
- [ ] https://github.com/consensus-shipyard/fendermint/issues/245
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/208
- [ ] https://github.com/consensus-shipyard/fendermint/issues/247
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/206
- [ ] https://github.com/consensus-shipyard/fendermint/issues/249
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/204
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/306
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/305
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/304
- [ ] https://github.com/consensus-shipyard/fendermint/issues/254
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/302
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/301
- [ ] https://github.com/consensus-shipyard/fendermint/issues/257
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/299
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/298
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/203
- [ ] https://github.com/consensus-shipyard/fendermint/issues/261
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/284
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/283
- [ ] https://github.com/consensus-shipyard/fendermint/issues/283
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/281
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/279
- [ ] https://github.com/consensus-shipyard/ipc-monorepo/issues/278
[...]

PR #899, opened 2024-05-08
Spec: Bottom up interaction

cryptoAtwill:
The bottom up interaction spec.

### DIFF
--- a/spec/bottom up interraction.md
+++ b/spec/bottom up interraction.md
@@ -1,0 +1,119 @@
+# Bottom Up Interactions
+This document takes a closer look in the IPC mechanics involved in information flowing from the child to the parent subnet, a.k.a. bottom-up.
+
+# Interactions
+
+There are two user initiated interactions in IPC that result in bottom-up messages being sent:
+
+- `release` sends tokens from a user account on the child subnet to another on the parent subnet
+- general cross-net messages sent to, or through, the parent
+
+The mechanism for propagating information from the child to the parent is through *checkpoints*.
+
+# Checkpoints
+
+The epic for implementing checkpointing is [here](https://github.com/consensus-shipyard/ipc/issues/211). The end-to-end workflow can be followed on this [diagram](https://github.com/consensus-shipyard/ipc/blob/main/docs/fendermint/checkpointing.md). The IPLD Resolver docs also present a [use case](https://github.com/consensus-shipyard/ipc/blob/specs/ipld/resolver/docs/README.md#checkpointing) for checkpointing.
+
+## Contents
+
+The original idea for a checkpoint was to contain the following information:
+
+- `subnet_id`: to identify to the parent subnet which child the checkpoint is for, and to prevent any replay attacks across subnets run by the same validators
+- `block_height` : the height of the child subnet blockchain at the time of checkpoint creation
+- `block_hash`: the hash of the block whose execution results in a checkpoint being added to the ledger, to prevent long range attacks on the subnet by anchoring it to the parent chain
+- `next_configuration_number`: this is the identifier of the validator set which is going to sign the *next* checkpoint; the current checkpoint is always going to be signed by the *current* validator set
+
+To these would be added fields to carry the contents of the checkpoint, which would be either:
+
+- `messages`, which would be a list of bottom-up cross-net messages, or
+- `messages_cid`, which would be the CID of the messages, but not the payload itself, which would be procured by the [IPC Spec - IPLD Resolver](https://www.notion.so/IPC-Spec-IPLD-Resolver-7b4290a0d60c40cdba98cd6d3e66648b?pvs=21)
+
+The CID based approach would only work with Fendermint, not Lotus running on rootnet, but in general a commitment based approach can work with Lotus too.
+
+## Triggers
+
+Originally checkpoints were supposed to be submitted at regular intervals, which was governed by the parameters with which the child subnet contract was created on the parent subnet, and made part of the `IpcParams` in [`genesis`](https://github.com/consensus-shipyard/ipc/blob/7af25c4c860f5ab828e8177927a0f8b6b7a7cc74/fendermint/vm/genesis/src/lib.rs#L227). However, this presented a problem with the number of messages that could be included in a checkpoint, which is why later the triggers for checkpoint creation were amended to be any of the following conditions:
+
+- a fixed period in terms of block height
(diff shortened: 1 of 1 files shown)


spec/bottom up interraction.md
```diff
@@ -0,0 +1,119 @@
...
+
+Ideally we would like some redundancy in the number of validators, so there is no single point of failure.
+
+Relayers generally should be rewarded for their service. There are numerous opinions on how to implement rewards, each with their drawback:
+
+- *Only the first submitter gets a reward.* To achieve redundancy the reward would have to be multiple of the cost. Depending on how likely it is to beat the fastest relayer (a parent validator might insert their own transaction to steal the rewards), it might make it unprofitable for multiple validators to operate.
+- *The first N submitters get rewards.* It is easy for any relayer to submit N transactions in a Sybil attack to reap all rewards, hampering redundancy.
+- *All submitters in a fixed time period get equal share of a fixed reward.* This takes out the competition aspect, and discourages Sybil attacks because the reward doesn’t grow. It should lead to a dynamic equilibrium of the number of relayers. However if the fixed time window is too wide, it encourages freeloaders who just repeat the first submission, which would make it look like there is redundancy where there isn’t.
```
sanderpick, line 86: I like this option because it implies that the goal should be that all honest validators share the burden of the cost of submitting a txn. ie, at equilibrium, `shared_reward / num_validators ~= txn_cost`.
 
 > it encourages freeloaders who just repeat the first submission
 
 Commit-reveal pattern could be helpful here?

spec/bottom up interraction.md
```diff
@@ -0,0 +1,119 @@
...
+
+Once the quorum has been verified, the `next_configuration_id` informs the parent how far ahead it can apply the pending validator updates on the active validator set, and thus know who to expect the *next* checkpoint to be signed by.
+
+## Execution
+
+When the checkpoint is submitted to the Lotus rootnet, it is currently expected to either contain all bottom-up messages or that they accompany the checkpoint in a different way, and only a commitment is in the checkpoint. However in both cases the bottom-up messages would be executed and their gas cost paid for by the relayer.
+
+With a Fendermint parent network, the same thing works if that’s how the smart contracts are implemented. There was another way laid it in the epic above, which involved the IPLD Resolver procuring the checkpoint payload from the subnet based on a CID, and executing the messages implicitly when the validators decide that they all have the data available. In this case the relayer would have only paid for the *validation* of the checkpoint, not its *execution*.
```
sanderpick, line 107: > There was another way laid it in the epic above

Typo?

spec/bottom up interraction.md
```diff
@@ -0,0 +1,119 @@
...
+- In the case where the validators execute messages implicitly, they can choose whether to include it in a block or not, but to do so they would need to estimate the cost at some point, and again it might exceed the limits. Implicit execution also makes it difficult to deal with errors, in particular there is no room for retries.
+
+To overcome this issue, ideally cross messages would *not* be executed in the block where the checkpoint is included. Instead either just a commitment would be stored, or messages would be parked in inboxes (e.g. organised by sender account). The senders could come later and initiate their own transactions to kick off the execution of the messages delivered as part of the IPC consensus mechanism, at which point they can pay for the gas and retry as many times as they see fit if they run out of gas.
```
sanderpick, line 114: :+1:

sanderpick (commented):
Great to see this. Note, the epic mentioned (https://github.com/consensus-shipyard/ipc/issues/211) has unchecked issues that were all "closed as not planned". Is that correct? There is at least one [todo in the codebase](https://github.com/consensus-shipyard/ipc/blob/main/fendermint/vm/interpreter/src/chain.rs#L288) that is closed.

aakoshh:
> the epic mentioned (https://github.com/consensus-shipyard/ipc/issues/211) has unchecked issues that were all "closed as not planned". Is that correct? 

I think the numerous unticked URLs are a remnant of where this issue was migrated from the `fendermint` to the `ipc` monorepo. 

>There is at least one [todo in the codebase](https://github.com/consensus-shipyard/ipc/blob/main/fendermint/vm/interpreter/src/chain.rs#L288) that is closed.

I don't know why that was closed, I think @jsoares did some cleanup and might have closed it. It wasn't implemented because in the end they implemented the submission in pure Solidity where this CID based approach would have been a lot of extra work. If the actors were in Rust, we could have just pulled stuff from their state using CIDs, but with Ethereum actors this isn't possible due to how Solidity stores the data. Or at least I don't think it's as straight forward. So it's right to be closed as nobody is working on this or planning to finish it right now, but the code was left in there because it provides some talking points, should it be revived.

spec/bottom up interraction.md
aakoshh: Yes, that's what I thought, at equilibrium the reward covers the cost. Any more joiners and somebody leaves; too many leave and it's profitable to join, so you can target a 3x redundancy by giving out 3x transaction cost. 

I haven't thought of a commit reveal pattern. It sounds promising; maybe the payload can be turned into a Merkle tree with a sender specific proof to show they have it, and then anyone can reveal the content a few blocks later?

spec/bottom up interraction.md
aakoshh, line 107:
```suggestion
With a Fendermint parent network, the same thing works if that’s how the smart contracts are implemented. There was another way laid out it in the epic above, which involved the IPLD Resolver procuring the checkpoint payload from the subnet based on a CID, and executing the messages implicitly when the validators decide that they all have the data available. In this case the relayer would have only paid for the *validation* of the checkpoint, not its *execution*.
```

jsoares:
Indeed, don't read anything into my closing it -- it was merely the auto-close stale issues policy in action (in line with the [linear method](https://linear.app/changelog/2020-08-19-auto-close-and-auto-archive)). The logic is that if it's relevant, someone will reopen it, and we see that at play here :) 

That said, also note that the line is pre-monorepo, so the issue wouldn't match current numbering either.

spec/bottom up interraction.md
sanderpick: Another option: 
- Senders update a hash with the Payload and with a sender-chosen one-time-use "secret". 
- Sender posts a txn with this final hash (the "commitment"). 
- In the next time window, senders post the actual payload plus their one-time secret (the "reveal").
- The contract validates this second txn by trying to recreate the commitment (hash the sent payload w/ secret and comparing to content of first txn).

The downside is needing some intermediate state (one `bytes32` per sender) and multiple txns, but the first one is relatively inexpensive.

sanderpick: I suppose you could make the first transaction also take the "reveal" from the prior time window. So, just one txn per window needed.

raulk (approved):

PR #904, opened 2024-05-08
Spec: Upgrades

fridrik01:
Added documentation for IPC Upgrades

### DIFF
--- a/specs/upgrades.md
+++ b/specs/upgrades.md
@@ -22,8 +22,8 @@ Once you have [deployed your own subnet](https://docs.ipc.space/quickstarts/depl
 
 The code for the subnet actor is located in `contracts/src/subnet` which you can directly edit to make your changes. Once ready, you run the following steps to upgrade your already deployed subnet actor:
 
-1. First, you must know your subnet ID which was returned when you created a child subnet (using `ipc-cli subnet create`). Lets say when you created your subnet that your subnet ID was: `/r314159/t410fkp4r67rks3ok4bvbn2rjojhkbig2rwvp4nmor5q`
-2. We must convert the subnet ID to an ETH address. To do this, visit the [Beryx address converter](https://beryx.zondax.ch/address_converter) and input the `t410` address in the Filecoin address input, in this example this would be `t410fkp4r67rks3ok4bvbn2rjojhkbig2rwvp4nmor5q`. Click `Convert to ETH` and it will compute the Ethereum address as `0x53f91f7e2a96dcae06a16ea29724ea0a0da8daaf`.
+1. First, you must know your subnet ID which was returned when you created a child subnet (using `ipc-cli subnet create`). Lets say when you created your subnet that your subnet ID was: `/r314159/f410fkp4r67rks3ok4bvbn2rjojhkbig2rwvp4nmor5q`
+2. We must convert the subnet ID to an ETH address. To do this, visit the [Beryx address converter](https://beryx.zondax.ch/address_converter) and input the `f410` address in the Filecoin address input, in this example this would be `f410fkp4r67rks3ok4bvbn2rjojhkbig2rwvp4nmor5q`. Click `Convert to ETH` and it will compute the Ethereum address as `0x53f91f7e2a96dcae06a16ea29724ea0a0da8daaf`.
 3. You must set the `RPC_URL` and `PRIVATE_KEY` environmental variables to point to your network provider and the private key of the address you want to use for the deployment, respectively
 
 Now you should be able to upgrade the subnet actor by running the following command (change NETWORK if you have deployed your subnet on another network)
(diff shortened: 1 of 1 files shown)


specs/upgrades.md
```diff
@@ -0,0 +1,387 @@
...
+
+The IPC actors are implemented using the diamond pattern to make it easier to upgrade them. The following sections describes how to upgrade each of these actors.
+
+## Upgrading the Subnet actor
+
+Once you have [deployed your own subnet](https://docs.ipc.space/quickstarts/deploy-a-subnet) you might want to make changes to the subnet actor, for example make custom logic changes, add new functionality, fix bugs, etc.
+
+The code for the subnet actor is located in `contracts/src/subnet` which you can directly edit to make your changes. Once ready, you run the following steps to upgrade your already deployed subnet actor:
+
+1. First, you must know your subnet ID which was returned when you created a child subnet (using `ipc-cli subnet create`). Lets say when you created your subnet that your subnet ID was: `/r314159/t410fkp4r67rks3ok4bvbn2rjojhkbig2rwvp4nmor5q`
```
maciejwitowski, line 25: I think the spec should use `f410` in all examples, instead of `t410` in order to describe the situation of the mainnet by default, not the testnet.

specs/upgrades.md
```diff
@@ -0,0 +1,387 @@
...
+2. We must convert the subnet ID to an ETH address. To do this, visit the [Beryx address converter](https://beryx.zondax.ch/address_converter) and input the `t410` address in the Filecoin address input, in this example this would be `t410fkp4r67rks3ok4bvbn2rjojhkbig2rwvp4nmor5q`. Click `Convert to ETH` and it will compute the Ethereum address as `0x53f91f7e2a96dcae06a16ea29724ea0a0da8daaf`.
+3. You must set the `RPC_URL` and `PRIVATE_KEY` environmental variables to point to your network provider and the private key of the address you want to use for the deployment, respectively
```
maciejwitowski, line 27: I think it's worth clarifying that "the address you want to use for the deployment" must be the one who deployed the subnet in the first place (that's my understanding)

specs/upgrades.md
```diff
@@ -0,0 +1,387 @@
...
+-----------------------------------
+Diamond Address: 0x53f91f7e2a96dcae06a16ea29724ea0a0da8daaf
+Replacement Facet Name: SubnetActorCheckpointingFacet
+
+Deployment Status:
+-------------------------
+New replacement facet (SubnetActorCheckpointingFacet) deployed.
+```
+
+If you run the `make upgrade-sa-diamond` command again, you should see no output which would also confirm that the upgrade has succeeded.
```
maciejwitowski, line 75: Does it mean that during the upgrade, the bytecode of deployed vs new contract is compared? Maybe worth clarifying here that "you should see no output" until you update the Subnet Actor code again.

specs/upgrades.md
```diff
@@ -0,0 +1,387 @@
...
+
+The on-chain logic for IPC is implemented in two contracts, the IPC gateway and the subnet actor, as well as an auxiliary contract used for UX purposes, the registry.
+
+The gateway is a singleton actor that lives in every IPC subnet and implements the common logic for IPC. It is responsible for managing the collateral of subnet, enforcing the firewall requirement, and the cross-net interactions.
+
+The subnet actor is a user-defined actor that implements the specific logic of a subnet. This contract is deployed in the parent from which the child subnet wants to be deployed. There is one subnet-actor for each child subnet in the parent.
+
+The subnet registry behaves as a subnet actor factory that offers users a convenient way of deploying instances of the reference implementation of the subnet actor in a network.
+
+The IPC actors are implemented using the diamond pattern to make it easier to upgrade them. The following sections describes how to upgrade each of these actors.
```
maciejwitowski, line 17: It says "how to upgrade each of these actors." however there is only a section on Subnet Actor. What is the status of upgrading the gateway and registry?

specs/upgrades.md
```diff
@@ -0,0 +1,387 @@
...
+
+If you run the `make upgrade-sa-diamond` command again, you should see no output which would also confirm that the upgrade has succeeded.
+
+# Fendermint upgrades
+
+Fendermint incorporates a builtin `UpgradeScheduler`, enabling the execution of hardcoded `Upgrade` migrations at predetermined block heights to advance the on-chain state. Fendermint also supports functionality to halt at predetermined `halt_height` in order to switch binary versions.
+
+## Upgrade scheduler
+
+At the start of processing each block, Fendermint checks the `UpgradeScheduler` whether it contains an `Upgrade` for that block height. Only a single `Upgrade` can be scheduled for each block height. If an upgrade is found to be scheduled, it will execute the `migration` function that is associated with that `Upgrade`. If the upgrade migration returns an error it will crash the node.
```
maciejwitowski, line 83: Can you add the link to code where "Fendermint checks the `UpgradeScheduler`" happens?

specs/upgrades.md
```diff
@@ -0,0 +1,387 @@
...
+```bash
+> cat .fendermint/config/default.toml | grep halt_height
+halt_height = 0
+```
+
+**5. Start the new Fendermint version**
+
+We can now start the new version of Fendermint which contains the upgrade we need.
+
+Could you add in the section for upgrading gateway/registry and any other info you think may be missing there for IPC actors
```
maciejwitowski, line 387: Is this a comment copy-pasted by mistake? :)

specs/upgrades.md
fridrik01: I will update the PR with an empty section for upgrading the gateway/registry which @snissn was going to fill out ([slack](https://filecoinproject.slack.com/archives/C06KWC57DRA/p1714494536816469?thread_ts=1714456806.300739&cid=C06KWC57DRA))

specs/upgrades.md
fridrik01: Is that not clear from step 1?

raulk (approved):

PR #905, opened 2024-05-08
Add IPC spec - Cross-net Messaging

snissn:
IPC spec for Cross-net Messaging

### DIFF
--- a/specs/cross-net-messaging.md
+++ b/specs/cross-net-messaging.md
@@ -1,0 +1,279 @@
+# IPC Cross-net Messaging
+
+## Cross-net messages
+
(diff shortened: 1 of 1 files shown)


specs/cross-net-messaging.md
```diff
@@ -0,0 +1,279 @@
...
+        IPCAddress to;
+    		bool success;   // we can optionally use an int flag instead of bool
+        bytes32 id;     // id of the message being ack'ed. This id is the keccak of IpcMsg
+    	  bytes reason;   // byte encoded failure reason (as in try-catch)
+    }
+    ```
+    
+- The `id` of an `Receipt` message is determined through the `keccak(IpcEnvelope)` of the message being acknowledged.
+    - `sendXnetMessage` should be extended to return the keccak of the message, for the contract to retain and correlate the result with later
+- Only the execution of `Call` messages trigger `Receipt`, for plain `Transfer` messages, i.e. `fund` and `release` , no `Ack` are needed (at least for now that we are subsidizing these).
```
cryptoAtwill, line 53: I think if the `Transfer` fails, an `Ack` is also sent?

snissn: I agree. That line looks out of date to me. I shall update it. Thanks for the comment

raulk (commented):
There is useful info here, but not in a mergeable state.

PR #906, opened 2024-05-08
Add IPC spec - Ipc Actors

snissn:
Adds Ipc Actors Spec

### DIFF
--- a/specs/ipc-actors.md
+++ b/specs/ipc-actors.md
@@ -1,0 +1,114 @@
+# Ipc Actors
+
+## Architecture Design
+
+- An IPC instance is always rooted from an EVM-compatible network (in our case Filecoin).
+- Users are allowed to deploy new subnets from any of the subnets in the hierarchy (as long as the corresponding subnet has the feature enabled).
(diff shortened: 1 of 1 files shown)


cryptoAtwill (approved):

cryptoAtwill:
This seems more like a IPC tech overview than IPC Actor, is this the goal of the doc?

snissn:
> This seems more like a IPC tech overview than IPC Actor, is this the goal of the doc?

I agree there may be something a little unpolished about this spec, but overall it is helpful. It is a combination of this notion doc https://www.notion.so/pl-strflt/IPC-Actors-3ca52db23a114eb193573e88f901ba98?pm=c and notes that I wrote about the ipc actors. I also noticed that the images from the notion doc were missing and I added them to this PR. I think the images give context to make this document more palatable and worth including as is.

raulk (commented):
There is useful information here, but not in a clean/structured enough way. Keeping this PR open so we can potentially use this content as reference if needed while we engage in a major docs overhaul.

PR #907, opened 2024-05-08
Spec: Deployment

snissn:
Adds deployment ipc spec


raulk:
This is duplicative with existing docs. The hack about using the scripts to initialize a subnet within a Docker deployment, to later copy its state/config out of the containers for faster development cycles is useful. But it'd be better to package it in a script, than to embed the script in a doc.

PR #909, opened 2024-05-08
Spec: Axelar token

snissn:
adds spec for axelar token

### DIFF
--- a/specs/extras/axelar-token.md
+++ b/specs/extras/axelar-token.md
@@ -1,0 +1,75 @@
+# IPC Axelar Token Example
+
+> Cross L1 chain atomic deposit of ERC20 tokens into token-supply subnets, via the Axelar network.
+> 
+
+**⚠️ Disclaimer: The contracts and implementation herein have not been security-reviewed or audited.
+They are considered bleeding-edge in terms of development.
+IPC developers do not make any guarantees.
+USE AT YOUR OWN RISK.**
+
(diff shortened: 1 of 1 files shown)


specs/extras/axelar-token.md
```diff
@@ -0,0 +1,75 @@
...
+But once `TOK` arrives to Filecoin in the form of `filTOK`, users would have to sign yet another transaction (this time on Filecoin) to deposit the newly arrived tokens into the IPC L2 subnet.
+
+In summary: sign a tx on the foreign L1, wait, switch network to the root L1, sign another tx.
+This is too much work for users.
+
+This repo facilitates atomic cross L1 transfers and deposits of tokens from a foreign L1, straight into IPC L2 subnets.
+
+## Solution
+
+![./architecture.png](./architecture.png)
```
maciejwitowski, line 36: Broken link

maciejwitowski:
Oh, so this is just a copy of [the axelar token's readme](https://github.com/consensus-shipyard/ipc/blob/main/extras/axelar-token/README.md), right? I don't think we need it then.

snissn:
> Oh, so this is just a copy of [the axelar token's readme](https://github.com/consensus-shipyard/ipc/blob/main/extras/axelar-token/README.md), right? I don't think we need it then. Wdyt?

That works for me too. This would only be redundant. Closing this PR

PR #916, opened 2024-05-10
docs(specs): Executions

aakoshh:
Add IPC execution docs.

### DIFF
--- a/specs/executions.md
+++ b/specs/executions.md
@@ -1,0 +1,292 @@
+# IPS Spec - Executions
+
+This document is a top-down view of the Fendermint application, concentrating on its aspects as a transition function.
+
+# ABCI
+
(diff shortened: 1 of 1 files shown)


raulk (approved):
These implementation notes will get outdated quickly as we refactor the codebase. Ideally, we would've not narrated how the code works (that belongs in Rustdocs), but we should've captured the design assumptions, expected behaviour, constraints, limitations, future considerations, etc. of the Fendermint "app".

For example, a more appropriate description of `ProcessProposal` would've been something like:

> This method is called by the consensus engine to quickly validate an incoming proposal from a block producer. It is not supposed to do any heavy-lifting, but simply validate the ability for the author to produce the block, the overall structural and syntactical correctness of it constituents, and our ability to fully execute the block if/when we're asked to (e.g. that data availability constraints are satisfied).
>
> Fendermint validates the following:
> ...
>
> Fendermint only reads these app state elements, and does not perform any internal state mutations:
> ...
> ...

That said, we'll merge this regardless to clear the queue, and then we'll refactor docs, specs, and Rustdocs.

PR #919, opened 2024-05-10
STORAGE: KVRead::iterate

aakoshh:
Adds an `iterate` method to the `KVRead` abstraction if the `KVStore::Repr` type implements `Ord`. 

It could be extended to take a parameter like the RocksDB implementation to say whether we want to iterate from the start, the end, from a specific key, or in between two keys; perhaps `iterate` could take parameters and a `KVCollection::iter` could just do the default in-order one. Here I just wanted to see how it can be implemented and where to add it.

This is a support PR for https://github.com/consensus-shipyard/ipc/pull/914 to implement initialisation from the storage.

### DIFF
--- a/fendermint/storage/src/im.rs
+++ b/fendermint/storage/src/im.rs
@@ -160,6 +160,8 @@ where
     where
         S: Decode<K> + Decode<V>,
         <S as KVStore>::Repr: Ord + 'static,
+        K: 'static,
+        V: 'static,
     {
         if let Some(m) = self.data.get(ns) {
             let mut items = m.iter().map(|(k, v)| (k, v.as_ref())).collect::<Vec<_>>();
--- a/fendermint/storage/src/im.rs
+++ b/fendermint/storage/src/im.rs
@@ -156,7 +156,7 @@ where
         Ok(None)
     }
 
-    fn iter<K, V>(&self, ns: &S::Namespace) -> impl Iterator<Item = KVResult<(K, V)>>
+    fn iterate<K, V>(&self, ns: &S::Namespace) -> impl Iterator<Item = KVResult<(K, V)>>
     where
         S: Decode<K> + Decode<V>,
         <S as KVStore>::Repr: Ord + 'static,
--- a/fendermint/storage/src/im.rs
+++ b/fendermint/storage/src/im.rs
@@ -2,13 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 use std::{
     hash::Hash,
+    marker::PhantomData,
     mem,
     sync::{Arc, Mutex, MutexGuard},
     thread,
 };
 
 use crate::{
-    Decode, Encode, KVRead, KVReadable, KVResult, KVStore, KVTransaction, KVWritable, KVWrite,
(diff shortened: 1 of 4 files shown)


fendermint/storage/src/im.rs
```diff
@@ -152,6 +155,23 @@ where
         }
         Ok(None)
     }
+
+    fn iterate<K, V>(&self, ns: &S::Namespace) -> impl Iterator<Item = KVResult<(K, V)>>
+    where
+        S: Decode<K> + Decode<V>,
+        <S as KVStore>::Repr: Ord + 'static,
+        K: 'static,
+        V: 'static,
```
fridrik01, line 164: This needs to be static so K/V is guaranteed to outlive the iterator?

fendermint/storage/src/im.rs
```diff
@@ -183,6 +203,48 @@ where
     }
 }
 
+struct KVIter<'a, S: KVStore, K, V> {
+    items: Vec<(&'a S::Repr, &'a S::Repr)>,
+    next: usize,
```
fridrik01, line 208: maybe `curr_size` ?

fendermint/storage/src/im.rs
```diff
@@ -183,6 +203,48 @@ where
...
+impl<'a, S, K, V> Iterator for KVIter<'a, S, K, V>
+where
+    S: KVStore + Decode<K> + Decode<V>,
+{
+    type Item = Result<(K, V), KVError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some((k, v)) = self.items.get(self.next) {
+            self.next += 1;
+            let kv = S::from_repr(k).and_then(|k| S::from_repr(v).map(|v| (k, v)));
```
fridrik01, line 240: :+1:

fendermint/storage/src/lib.rs
```diff
@@ -156,4 +166,18 @@ where
...
+    pub fn iterate<'a, 'b>(
+        &'a self,
+        kv: &'b impl KVRead<S>,
+    ) -> impl Iterator<Item = KVResult<(K, V)>> + 'b
+    where
+        S::Repr: Ord + 'static,
+        S: Decode<K>,
+        K: 'static,
+        V: 'static,
+        'a: 'b,
```
fridrik01, line 179: I recall seeing this before, but its one of these lifetimes I never remember how is written

fridrik01 (approved):
Pretty good stuff!

fendermint/storage/src/im.rs
aakoshh: I'm not sure I can explain, but the compiler insisted that the iterator can outlive these types :shrug:

fendermint/storage/src/im.rs
aakoshh: What I wanted to express here is the next index that will be returned from the `Vec`. The vec isn't being consumed, its size stays the same.

fendermint/storage/src/lib.rs
aakoshh: This is also from the compiler. I didn't think adding this method on `KVCollection` would involve more constraints. It complained that the `impl Interator` captures the lifetimes of both reference inputs, so it needs to outlive them, and this was the only way to express that they are the same, or compatible.

PR #1007, opened 2024-05-24
Fixes E2E tests on Apple Silicon

ec2:
Fixes the issues that I've run into trying to run the e2e tests on my new Macbook. 
I was getting errors running `make e2e` under the `fendermint` directory. 

[] Specify `amd/64` platform to Docker when pulling the go-ipfs and graph-node images
[] Unix `sleep 30` instead of GNU `sleep 30s` (without specifying a unit, it will default to seconds anyways)


raulk (approved):

raulk:
Thanks for the fix, @ec2! ❤️

Issue #1002, opened 2024-05-23
Deterministic genesis (+ import/export commands)


cryptoAtwill:
# Introduction
When a follower node want to join a network, one could run into `app hash not equal` at the genesis block when starting fendermint. 

This is because:
- A wrong genesis file
- Wrong versions for fvm actors or ipc contracts

For 1, it is most likely human error, the correct genesis configuration file should fix it. 

The more problematic one is the second scenario. It happens when, say, a 3 node cluster is setup, but then there are some `ipc-contracts` update and when a 4th node tries to join the network, the `ipc-contracts` byte code is not the same as when the original 3 nodes were setup, leading to a wrong `app_hash`. 

The root cause of scenario 2 is due to some configuration parameters are not in the genesis file. These parameters are:
- builtin_actors_bundle
- custom_actors_bundle
- ipc_contracts
specifically, the above three parameters are paths, the content at those paths can change.

To fix this, one must make sure the data loaded into fendermint at genesis is the same. There are a few approaches:

### 1.Export/Import genesis
When the `init_chain` is called from `cometbft`, once the genesis state is created, we force a `genesis` snapshot to be taken and write this snapshot to disk. This snapshot can be uploaded to some public space so others could access it. Then when fendermint is launched, one can just pass the snapshot either throw path or url to indicate it's loaded from genesis snapshot. Ofc, one needs to ensure the correct genesis snapshot is used.

### 2.Include everything to genesis
Instead of passing in `builtin_actors_bundle` and all those paths in `fendermint` configuration file, we can just put the actual bytes in the genesis file, this would result in something like:
```
{
    "builtin-actor-bundle": "0x....",
    "custom-actor_bundle": {
        "eam": "0x...",
        ...
    },
    "ipc-contracts": {
        "gateway": "0x...."
         ... 
    }
}
```
We could include content hashes to the above genesis file and update genesis generation command to include these parameters. Ofc, the downside is bloating the genesis file size, but perhaps this is a small price for consistency.

### 3. Include hashes and links
Instead of putting raw bytes into the genesis, maybe we can just add the url and sha256 hashes to the path contents. In a way, the genesis file does not care where one fetches the content, as long as the hashes match or when not matching, one can quitely tell which parameter is causing the inconsistency, it's good enough. But ofc, the deployer needs to know where to get the historical content because most likely, what the deployer has is really the latest code.

raulk:
## Context

[...]

PR #1009, opened 2024-05-27
Including builtin and custom actor bundles in genesis

cryptoAtwill:
As part one of #1002, the builtin and custom actor bundles are added into the genesis as precompile actor code.

Updating unit tests, but implementation should be fixed.


cryptoAtwill:
Closing in favour of #1016.

PR #1016, opened 2024-05-30
feat(node): consistent genesis

cryptoAtwill:
This PR follows this [discussion](https://github.com/consensus-shipyard/ipc/pull/1015#pullrequestreview-2085968059) to address #1002.

The flow is simpler now with:
- Addition of `SealGenesis` command that runs the previous genesis creation logic:
    * Load builtin/custom actor and deploy ipc contracts using memory store. Implementation not changed, just shifted location.
    * Write memory store to Car file.
- `InitChain` just reads the car file bytes into the `StateStore`

This PR is just an implementation of the flow, it's a quick way to show if the flow actually works. Pending items to make it ready for review:
- [x] Integration testing. The current state of the PR might have missing steps (unlikely, but possible), will actually perform integration testing to ensure it actually works. 
- [x] Unit tests.
- [x] Some further code clean up and simplification.
- [x] Fix broken CI due to this change.

To actually try it out, perform the following:
```sh
# generate the fendermint genesis first, then seal the genesis state
cargo run -p fendermint_app --release -- genesis --genesis-file test-network/genesis.json \
  ipc \
    seal-state \
    --builtin-actors-path <FULL PATH TO YOUR BUILT_IN BUNDLE> \
    --custom-actors-path  <FULL PATH TO YOUR CUSTOM BUNDLE> \
    --artifacts-path  <FULL PATH TO YOUR CONTRACT OUT FOLDER> \
    --output-path <FULL PATH TO YOUR SEALED STATE OUTPUT PATH>

# convert to cometbft genesis
cargo run -p fendermint_app --release -- \
  genesis --genesis-file test-network/genesis.json \
  into-tendermint --out ~/.cometbft/config/genesis.json --sealed  <FULL PATH TO YOUR SEALED STATE OUTPUT PATH>
```


fendermint/app/options/src/genesis.rs
```diff
@@ -147,6 +147,28 @@ pub enum GenesisIpcCommands {
...
+    pub builtin_actors_path: PathBuf,
+
+    /// The custom actors bundle path
+    #[arg(long, short)]
+    pub custom_actors_path: PathBuf,
+
+    /// The ipc artifacts output path. If you are using ipc-monorepo, it should be the `out` folder
+    /// of `make build`
+    #[arg(long, short)]
+    pub ipc_artifacts_path: PathBuf,
```
raulk, line 167: I wonder if we can embed these artifacts in the build, and load them from the target directory instead?

fendermint/app/options/src/genesis.rs
```diff
@@ -147,6 +147,28 @@ pub enum GenesisIpcCommands {
...
+
+    /// The sealed, i.e. finalized genesis state CAR file dump path
+    #[arg(long, short)]
+    pub sealed_genesis_path: PathBuf,
```
raulk, line 171: Let's simply call this output_path.

fendermint/vm/genesis/src/lib.rs
```diff
@@ -7,6 +7,7 @@ use anyhow::anyhow;
 use fvm_shared::bigint::{BigInt, Integer};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
+use std::path::PathBuf;
```
raulk, line 10: Needed?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,785 @@
...
+    } else {
+        return Err(anyhow!("invalid genesis car, metadata not found"));
+    };
+
+    Ok((metadata.validators, metadata.state_params))
+}
+
+/// The output of genesis creation
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FvmGenesisOutput {
```
raulk, line 91: Can drop Fvm from the name.

raulk (commented):
Will do a deeper review tomorrow, but this is headed in the right direction!

In the meantime, can you work on the command to extract the state tree from Fluence's genesis? 

@folex can give you an endpoint if you need one.

fendermint/vm/genesis/src/lib.rs
cryptoAtwill: removed alr

fendermint/app/src/cmd/genesis.rs
```diff
@@ -94,6 +95,8 @@ cmd! {
             set_ipc_gateway(&genesis_file, args),
         GenesisIpcCommands::FromParent(args) =>
             new_genesis_from_parent(&genesis_file, args).await,
+        GenesisIpcCommands::SealState(args) =>
```
karlem, line 98: I wonder if it makes sense to have this under IPC?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+        Self {
+            hardhat,
+            top_contracts,
+            lib_addrs: Default::default(),
+            phantom_db: PhantomData,
+        }
+    }
+
+    /// Deploy a library contract with a dynamic ID and no constructor.
+    pub fn deploy_library(
```
karlem, line 518: No need for it to be pub?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+        // Maybe the masked ID is quicker because it doesn't need to be resolved.
+        self.lib_addrs.insert(fqn, id_addr);
+
+        *next_id += 1;
+
+        Ok(())
+    }
+
+    /// Construct the bytecode of a top-level contract and deploy it with some constructor parameters.
+    pub fn deploy_contract<T>(
```
karlem, line 557: No need for it to be pub?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+            contract_name,
+            "deployed Ethereum contract"
+        );
+
+        // The Ethereum address is more usable inside the EVM than the ID address.
+        Ok(eth_addr)
+    }
+
+    /// Collect Facet Cuts for the diamond pattern, where the facet address comes from already deployed library facets.
+    pub fn facets(&self, contract_name: &str) -> anyhow::Result<Vec<FacetCut>> {
```
karlem, line 595: No need for it to be pub?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+}
+
+/// Sum of balances in the genesis accounts.
+fn circ_supply(g: &Genesis) -> TokenAmount {
+    g.accounts
+        .iter()
+        .fold(TokenAmount::zero(), |s, a| s + a.balance.clone())
+}
+//
+// #[cfg(test)]
```
karlem, line 642: We probably want to either re-write the test so that they work with the new implementation or just delete them at least?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+
+pub struct GenesisCreator {
+    /// Hardhat like util to deploy ipc contracts
+    hardhat: Hardhat,
+    /// The built in actors bundle path
+    builtin_actors_path: PathBuf,
+    /// The custom actors bundle path
+    custom_actors_path: PathBuf,
+    /// The CAR path to flush the sealed genesis state
+    sealed_out_path: PathBuf,
```
karlem, line 111: We don't really mention that it creates sealed genesis anywhere so I guess out_path would do?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+
+        tracing::info!("written sealed genesis state to file");
+
+        Ok(())
+    }
+
+    async fn init_state(&self) -> anyhow::Result<FvmGenesisState<MemoryBlockstore>> {
+        let bundle = std::fs::read(&self.builtin_actors_path).with_context(|| {
+            format!(
+                "failed to read bundle: {}",
```
karlem, line 175:
```suggestion
                "failed to read builtin actors bundle: {}",
```

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+                self.builtin_actors_path.to_string_lossy()
+            )
+        })?;
+
+        let custom_actors_bundle = std::fs::read(&self.custom_actors_path).with_context(|| {
+            format!(
+                "failed to read custom actors_bundle: {}",
```
karlem, line 182:
```suggestion
                "failed to read custom actors bundle: {}",
```

fendermint/vm/interpreter/src/fvm/state/genesis.rs
```diff
@@ -177,6 +177,17 @@ where
         }
     }
 
+    /// Flush the data to the block store. Returns the state root cid and the underlying state store.
+    pub fn finalize(self) -> anyhow::Result<(Cid, DB)> {
```
karlem, line 181: It might be worth to either explain the difference between this and finalise in comment or remove one a keep the other? Now it's a bit ambiguous.

fendermint/app/options/src/genesis.rs
```diff
@@ -133,6 +133,9 @@ pub struct GenesisAddValidatorArgs {
 
 #[derive(Args, Debug)]
 pub struct GenesisIntoTendermintArgs {
+    /// Sealed genesis file that is the initial app bytes for cometbft
+    #[arg(long, short)]
+    pub sealed: PathBuf,
```
karlem, line 138: I think calling it `sealed_state` might be more descriptive?

fendermint/app/options/src/genesis.rs
```diff
@@ -147,6 +150,28 @@ pub enum GenesisIpcCommands {
     Gateway(GenesisIpcGatewayArgs),
     /// Fetch the genesis parameters of a subnet from the parent.
     FromParent(Box<GenesisFromParentArgs>),
+    /// Seal the genesis state from the genesis parameter file
+    SealState(SealGenesisArgs),
+}
+
+#[derive(Args, Debug, Clone)]
+pub struct SealGenesisArgs {
```
karlem, line 158:
```suggestion
pub struct SealGenesisStateArgs {
```

maybe?

fendermint/app/options/src/genesis.rs
```diff
@@ -147,6 +150,28 @@ pub enum GenesisIpcCommands {
...
+    pub builtin_actors_path: PathBuf,
+
+    /// The custom actors bundle path
+    #[arg(long, short)]
+    pub custom_actors_path: PathBuf,
+
+    /// The solidity artifacts output path. If you are using ipc-monorepo, it should be the `out` folder
+    /// of `make build`
+    #[arg(long, short)]
+    pub artifacts_path: PathBuf,
```
karlem, line 170: I wonder if we want this one to be optional? In case we do not want to deploy IPC related functionality? If we always want to then ignore this message :)

fendermint/app/src/cmd/genesis.rs
```diff
@@ -213,7 +216,7 @@ fn set_eam_permissions(
 
 fn into_tendermint(genesis_file: &PathBuf, args: &GenesisIntoTendermintArgs) -> anyhow::Result<()> {
     let genesis = read_genesis(genesis_file)?;
-    let genesis_json = serde_json::to_value(&genesis)?;
+    let app_state = hex::encode(std::fs::read(&args.sealed)?);
```
karlem, line 219: Given the sealed state is extremely large - we might benefit from using `base64` instead of hex as that might be more compact. WDYT?

fendermint/app/src/cmd/genesis.rs
```diff
@@ -248,7 +251,7 @@ fn into_tendermint(genesis_file: &PathBuf, args: &GenesisIntoTendermintArgs) ->
         // Hopefully leaving this empty will skip validation,
         // otherwise we have to run the genesis in memory here and now.
         app_hash: tendermint::AppHash::default(),
-        app_state: genesis_json,
+        app_state: serde_json::Value::String(app_state),
```
karlem, line 254: I suggest renaming the parse_genesis_app_bytes function to deserialise_genesis_app_bytes for better clarity and consistency. Additionally, it would be beneficial to add a corresponding serialise_genesis_app_bytes function. To keep related functionality together, consider moving both functions to a new module, perhaps named serde, within the genesis package. This

fendermint/app/src/cmd/genesis.rs
```diff
@@ -280,6 +283,19 @@ fn set_ipc_gateway(genesis_file: &PathBuf, args: &GenesisIpcGatewayArgs) -> anyh
     })
 }
 
+async fn seal_state(genesis_file: &PathBuf, args: &SealGenesisArgs) -> anyhow::Result<()> {
+    let genesis = read_genesis(genesis_file)?;
+
+    let genesis_creator = GenesisCreator::new(
```
karlem, line 289: Maybe `GenesisStateCreator`?

fendermint/testing/materializer/tests/docker.rs
```diff
@@ -166,7 +166,7 @@ async fn wait_for_startup(testnet: &DockerTestnet) -> anyhow::Result<bool> {
             }
 
             if let Some(client) = dnode.ethapi_http_provider()? {
-                if let Err(e) = client.get_chainid().await {
+                if let Err(e) = client.get_block(1).await {
                     eprintln!("EthAPI on {name} still fails: {e}");
```
karlem, line 170: It's no longer name?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+    } else {
+        return Err(anyhow!("invalid genesis car, metadata not found"));
+    };
+
+    Ok((metadata.validators, metadata.state_params))
+}
+
+/// The output of genesis creation
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct GenesisOutput {
```
karlem, line 93: It seems like there is exactly same struct in in fvm package called `FvmGenesisOutput`. Maybe you can re-use?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
+// Copyright 2022-2024 Protocol Labs
```
karlem, line 1: Is it still strictly necessary to keep this inside the interpreter? Would it be acceptable to move it to `fendermint_vm_genesis` crate?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+        let (metadata_cid, metadata_bytes) = derive_cid(&metadata)?;
+        tracing::info!("generated genesis metadata header cid: {}", metadata_cid);
+
+        // create the target car header with the metadata cid as the only root
+        let car = CarHeader::new(vec![metadata_cid], 1);
+
+        // create the stream to stream all the data into the car file
+        let mut streamer = tokio_stream::iter(vec![(metadata_cid, metadata_bytes)]).merge(streamer);
+
+        let write_task = tokio::spawn(async move {
```
karlem, line 159: Does it need to be wrapped in another async task?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+                        .context("failed to create multisig actor")?;
+                    next_id += 1;
+                }
+            }
+        }
+
+        // STAGE 3: Initialize the FVM and create built-in FEVM actors.
+
+        state
+            .init_exec_state(
```
karlem, line 411: I know this is not something you have implemented in this PR but the fact the state is using the stage enum is really confusing.

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,795 @@
...
+
+        Ok(out)
+    }
+}
+
+fn contract_src(name: &str) -> PathBuf {
+    PathBuf::from(format!("{name}.sol"))
+}
+
+struct ContractDeployer<'a, DB> {
```
karlem, line 496: Would it be suitable to move it elsewhere? As it is not strictly related only to genesis.

fendermint/vm/interpreter/src/genesis.rs
cryptoAtwill: I will remove them first, maybe rewrite them later

fendermint/vm/interpreter/src/fvm/state/genesis.rs
cryptoAtwill: I will remove `commit` method once this PR is merged, currently the old implementation is still there.

fendermint/vm/interpreter/src/genesis.rs
cryptoAtwill: I think that one is in a different mod. I'm planning to remove the old implementation after this one in favour of current implementation.

fendermint/vm/interpreter/src/genesis.rs
cryptoAtwill: but currently only genesis is using it, maybe we can move it if others are using it.

fendermint/app/src/cmd/genesis.rs
cryptoAtwill: maybe a compression before also works. Updated to base64.

cryptoAtwill: Because the bytes were all using `hex` through out the codebase, to be consistent. Maybe a compression before encoding will make it smaller?

fendermint/app/src/cmd/genesis.rs
cryptoAtwill: I think should be fine with `Genesis`, I renamed everything else to align with `genesis` and removed `state`.

fendermint/app/src/cmd/genesis.rs
cryptoAtwill: I think this might need a more in depth refactor, because the serialization is in `app` but deserialization in interpreter.

raulk:
TODO:
- [ ] Add a version number to the sealed state schema.
- [ ] Compress the state tree CAR with Snappy + encode in base64 instead of hex.
- [ ] Add a debugging tool to inspect the state tree. Iterate over the state tree printing actor IDs, code CIDs, actor names (resolved from the Code CID), balances.
- [ ] Update quickstart + user/dev documentation.

fendermint/app/src/cmd/genesis.rs
cryptoAtwill: Currently the separation btw ipc and non-ipc is blurred. I would actually argue we track this in another issue, because it's the downstream implementation that assumes an `artifacts` folder which makes it difficult to split upstream.

fendermint/app/options/src/genesis.rs
cryptoAtwill: see the comment above, it's difficult to refactor, quite some code downstream to change.

cryptoAtwill:
@karlem I have updated based on the review feedback, please take another look. I will remove old implementation in a follow up PR.

.github/workflows/fevm-contract-tests.yaml
```diff
@@ -71,8 +71,7 @@ jobs:
         id: docker-build
         working-directory: ipc/fendermint
         run: |
-          export PATH="$PATH:/home/runner/.config/.foundry/bin"
-          docker pull ghcr.io/consensus-shipyard/fendermint:latest
+          export PATH="$PATH:/home/runner/.config/.foundry/bin" && make docker-build
```
raulk, line 74: Isn't this gonna crazy expensive? IIRC we build a Docker image in another job, can't we reuse that Docker image?

fendermint/app/src/cmd/genesis.rs
```diff
@@ -280,6 +289,19 @@ fn set_ipc_gateway(genesis_file: &PathBuf, args: &GenesisIpcGatewayArgs) -> anyh
...
+        args.builtin_actors_path.clone(),
+        args.custom_actors_path.clone(),
+        args.artifacts_path.clone(),
+        args.output_path.clone(),
+    );
+
+    genesis_creator.create(genesis).await
```
raulk, line 302: These APIs are weird. We have a Genesis and the GenesisCreator. You would imagine the GenesisCreator creates a Genesis, but in reality the Genesis already exists and the GenesisCreator wants it as an input.

I think we need to adjust this a little.
- We should have a `GenesisBuilder`.
- The `genesis_file` is poorly named. These are genesis parameters, so let's name it `genesis_params_path`.
- You can then have a `GenesisBuilder::from_params` to construct a new `GenesisBuilder` from the parameters file.
- We then have a method for each artifact we need to take: `with_builtin_actors`, `with_custom_actors`, `with_system_contracts`.
- We then have a `write_into<W: io::Write>(w: &mut W)` method that writes the genesis file to a writer.
- The file is managed by `seal_genesis`.

In practice:

```rust
let mut output = // create the file

let gen = GenesisBuilder::from_params(genesis_params_path)
    .with_builtin_actors(...)
    .with_custom_actors(...)
    .with_system_contracts(...)
    .write_into(&output)
```

fendermint/testing/materializer/src/docker/node.rs
```diff
@@ -205,13 +205,29 @@ impl DockerNode {
...
+            .context("failed to seal genesis state")?;
+
         // Convert fendermint genesis to cometbft.
         fendermint_runner
             .run_cmd(
                 "genesis \
                     --genesis-file /fendermint/genesis.json \
                     into-tendermint \
                     --out /cometbft/config/genesis.json \
+                    --app-state /cometbft/config/sealed.json \
```
raulk, line 230: Why both an out and an `app-state`?

fendermint/vm/interpreter/src/genesis.rs
raulk: Good question.

raulk (commented):
Initial comments.

fendermint/app/options/src/genesis.rs
raulk: I think this is fine as it is.

fendermint/app/options/src/genesis.rs
raulk: @cryptoAtWill, any thoughts here?

fendermint/testing/materializer/src/docker/node.rs
cryptoAtwill: The `app-state` is practically the genesis state, but the terminology cometbft is using is `app-state`, just took that term here.

fendermint/app/options/src/genesis.rs
cryptoAtwill: oops, missed this. I think we have discussed this before, if I recall correctly, it's because `ipc_contracts` can be updated and compile it will result in a different artifacts build. Using a path gives more flexibility on which artifacts should be imported.

fendermint/app/src/cmd/genesis.rs
cryptoAtwill: This works for me. Just a side note that probably 
```
.with_builtin_actors(...)
.with_custom_actors(...)
.with_system_contracts(...)
```
should be added to constructor as they are required parameters, otherwise it feels it's not really required.

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -0,0 +1,765 @@
...
+
+impl GenesisAppState {
+    pub fn v1(bytes: Vec<u8>) -> Self {
+        Self::V1(bytes)
+    }
+
+    pub fn compress_and_encode(&self) -> anyhow::Result<String> {
+        let bytes = match self {
+            GenesisAppState::V1(ref bytes) => {
+                let mut wtr = snap::write::FrameEncoder::new(vec![0]);
```
raulk, line 88: @cryptoAtwill this looks like a bug? We write the version discriminator (0) under snappy compression, but we expect to find it uncompressed below?

PR #1017, opened 2024-05-31
chore(contracts): prepare for publishing v1.0.0-alpha.0.

raulk:
For now we'll do a manual publish to unblock folks, but we'll want to produce automated releases down the line.

### DIFF
--- a/contracts/.gitignore
+++ b/contracts/.gitignore
@@ -7,6 +7,7 @@ crytic-export/
 broadcast/
 out/
 binding/src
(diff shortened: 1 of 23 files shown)


raulk:
Our Foundry tests have been failing since last week due to this regression: https://github.com/foundry-rs/foundry/issues/8015. It was fixed today and will be released shortly in nightly, which is what we use in our CI build, so the error will disappear automatically soon.

PR #1018, opened 2024-05-31
fix(contracts/sdk): `IpcUpgradableOwnable`: pass `msg.sender` to `Ownable` `init` for compatibility with OZ v5

lordshashank:
[__Ownable_init()](https://github.com/consensus-shipyard/ipc/blob/b6d472d41601af6327df9c98dd0c486f57c77eaf/contracts/sdk/IpcContractUpgradeable.sol#L32) function requires `msg.sender` param in [@openzeppelin/contracts-upgradeable v5](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/f960f47267044822613be18e149c2e0ee1a3bf6e/contracts/access/OwnableUpgradeable.sol#L51)

### DIFF
--- a/.github/workflows/contracts-prettier.yaml
+++ b/.github/workflows/contracts-prettier.yaml
@@ -15,11 +15,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.head_ref }}
+          ref: ${{ github.ref }}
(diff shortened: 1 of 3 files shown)


raulk (approved):

lordshashank:
@raulk , I am not able to get why prettier CI failed.

PR #1021, opened 2024-06-04
feat(ethapi): add eth cors settings

sanderpick:
Second PR requested in a convo with @raulk.

This one adds eth CORS settings to control allowed origins, methods, and headers. It's modeled after the cometbft config, which includes the following settings under `[rpc]`:

```
# A list of origins a cross-domain request can be executed from
# Default value '[]' disables cors support
# Use '["*"]' to allow any origin
cors_allowed_origins = []

# A list of methods the client is allowed to use with cross-domain requests
cors_allowed_methods = ["HEAD", "GET", "POST", ]

# A list of non simple headers the client is allowed to use with cross-domain requests
cors_allowed_headers = ["Origin", "Accept", "Content-Type", "X-Requested-With", "X-Server-Time", ]
```


fendermint/app/config/default.toml
```diff
@@ -138,6 +138,17 @@ host = "127.0.0.1"
...
 port = 8545
 
+[eth.cors]
+# A list of origins a cross-domain request can be executed from
+# Default value '[]' disables cors support
+# Use '["*"]' to allow any origin
+allowed_origins = []
+# A list of methods the client is allowed to use with cross-domain requests
+# Suggested methods if allowing origins: "GET", "OPTIONS", "HEAD", "POST"
+allowed_methods = []
```
sanderpick, line 148: I left these empty by default because there is no way to override them to be _empty_ using an env var (empty strings [are ignored](https://github.com/consensus-shipyard/ipc/pull/1021/files#diff-e8a390f79ef4d76cff941961d69b6cf5c6bc1c5a08a0f0f763c2eaab3029d953R326) by the config list parser).

fendermint/app/settings/src/lib.rs
```diff
@@ -52,15 +52,15 @@ impl Display for SocketAddress {
...
+impl ToSocketAddrs for SocketAddress {
+    type Iter = <String as ToSocketAddrs>::Iter;
 
     fn to_socket_addrs(&self) -> std::io::Result<Self::Iter> {
         self.to_string().to_socket_addrs()
     }
 }
 
-impl TryInto<std::net::SocketAddr> for SocketAddress {
+impl TryInto<SocketAddr> for SocketAddress {
```
sanderpick, line 63: Just some cleanups... can revert if annoying.

fendermint/app/settings/src/lib.rs
```diff
@@ -422,21 +425,41 @@ mod tests {
...
                 ("FM_RESOLVER__NETWORK__NETWORK_NAME", "test"),
             ], || try_parse_config("")).unwrap();
 
-        assert_eq!(settings.resolver.discovery.static_addresses.len(), 2);
         assert_eq!(settings.resolver.connection.external_addresses.len(), 2);
+        assert_eq!(settings.resolver.discovery.static_addresses.len(), 2);
+        assert_eq!(settings.resolver.membership.static_subnets.len(), 2);
+        assert_eq!(
+            format!("{:?}", settings.eth.cors.allowed_origins),
+            "List([\"https://example.com\", \"https://www.example.org\"])"
```
sanderpick, line 441: Ugly, but no other way of testing this... there isn't a public API on `AllowedOrigin` to list its members.

ipc/provider/Cargo.toml
sanderpick, line 1: Just alphabetizing the dependencies... can revert if annoying. Some crates have ordered dependencies and other don't. I thought may as well clean up the ones I'm touching.

ipc/provider/src/config/deserialize.rs
```diff
@@ -106,3 +110,108 @@ fn eth_addr_str_to_address(s: &str) -> anyhow::Result<Address> {
...
+            formatter.write_str("a sequence of strings")
+        }
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: SeqAccess<'de>,
+        {
+            let mut origins = Vec::new();
+            while let Some(v) = seq.next_element::<String>()? {
+                if v == "*" {
```
cryptoAtwill, line 134: A quick question, sometimes, there are situation such as wildcard subdomains, like `*.examples.com`, it seems it's not handled by the parsing here? Just curious will it work if it's only raw strings parsed here and the json rpc server can handle wildcard subdomain?

sanderpick: Technically speaking wildcard subdomains aren't a proper [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin). Given a request, the server can respond with one of...

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Origin: <origin>
Access-Control-Allow-Origin: null
```

ie, "*", a specific domain, or none.

We could enable wildcard subdomains with some additional parsing + a [predicate](https://docs.rs/tower-http/latest/tower_http/cors/struct.AllowOrigin.html#method.predicate):

```
use tower_http::cors::{CorsLayer, AllowOrigin};
use http::{request::Parts as RequestParts, HeaderValue};

let layer = CorsLayer::new().allow_origin(AllowOrigin::predicate(
    |origin: &HeaderValue, _request_parts: &RequestParts| {
        origin.as_bytes().ends_with(b".rust-lang.org")
    },
));
```

This way the server would respond with the full valid request origin. I don't have a strong opinion on how important it is to support this though.

sanderpick:
@cryptoAtwill do we want to allow CORS requests that require authentication? If so, we'll need some additional handling:
- [Set Access-Control-Allow-Credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials)
- For "*" origins, we'd have to mirror the request origin instead of returning "*", otherwise there would be an [error](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials).

ipc/provider/src/config/deserialize.rs
cryptoAtwill: I guess my question is more like current RPC server library does not support `AllowOrigin::from_str` so that we have to parse the strings ourselves?

sanderpick: You can parse a `HeaderValue` from string and then parse `AllowOrigin` from that, but that would only work if the parsing could handle `String` and `Vec<String>` as input:

-  `"*"` -> `HeaderValue` -> `AllowOrigin`
-  `"http://example.com"` -> `HeaderValue` -> `AllowOrigin`
- `["http://example.com", "http://example.org"]` -> `Vec<HeaderValue>` -> `AllowOrigin`
- `["*"]` -> not parse-able without custom handling (`"*"` is not allowed in a list of you want to use the `Vec<HeaderValue>` -> `AllowOrigin` parsing.
- `"*.example.org"` -> not directly parse-able, need to use a predicate / closure

I originally took the `String` and `Vec<String>` as input approach, but it's not possible to provide a single `String` as input from the environment variables (they get parsed as `Vec`). Which means that without custom parsing, there would have been no way to use `"*"`. The same applies to methods and headers. It's also less confusing to only allow a list in the config files (modeled after cometbft config).

raulk (approved):

raulk:
Thanks for the contribution, @sanderpick! Looking forward to many more 💪

PR #1022, opened 2024-06-05
feat(contracts): remove `crypto` feature from `fvm_shared` in binding

dtbuchholz:
## Summary

Removes `crypto` feature from `fvm_shared` in `contracts/binding`

## Details

The `contracts/binding` uses `fvm_shared` with the `crypto` flag enabled. However, this is unneeded as it's not directly used, and it also impacts downstream libraries when trying to compile to wasm. E.g., trying to build to wasm will fail due to a transitive dependency of `positioned-io`, stemming from the enabled `crypto` feature.

### DIFF
--- a/contracts/binding/Cargo.toml
+++ b/contracts/binding/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 ethers = { workspace = true, features = ["abigen", "ws"] }
-fvm_shared = { workspace = true, features = ["crypto"] }
+fvm_shared = { workspace = true }
 anyhow = { workspace = true }
 
 [build-dependencies]


raulk (approved):
LGTM, but would like @cryptoAtwill to confirm, since Git history shows him creating the Cargo.toml with this feature when the crate was created.

dtbuchholz:
@raulk ah, i may have been quick to judge. the wasm-related `positioned-io` build errors were "resolved" (seemingly, no longer blocking things) in our fork, but it looks like building for the default target architecture leads to an issue. so, it appears the `crypto` feature _is_ needed (see test failure [here](https://github.com/consensus-shipyard/ipc/actions/runs/9375952479/job/25815191608?pr=1022#step:7:865)).

however, i do wonder why the `crypto` inclusion would also include the `positioned-io` dep—where the wasm I/O issues stem from. @cryptoAtwill for reference, `positioned-io` uses [`std::io` for files](https://github.com/vasi/positioned-io/blob/5b787fbcd43ce36f1ac6234aaee9917114ffac97/src/raf.rs#L7), which is what causes `wasm-bindgen`/`wasm-pack` to fail because it only considers unix or windows arch. disabling `crypto` confirms that this is where the issue stems from, and (also for reference) [here](https://github.com/filecoin-project/ref-fvm/blob/206a772745978802b431821af481f4523ccf6d5f/shared/Cargo.toml#L51) is what the `crypto` feature requires: 
```
crypto = ["libsecp256k1", "blst", "proofs"]
proofs = ["filecoin-proofs-api"]
```

dtbuchholz:
@raulk gotcha. so, the intention is to compile [our SDK](https://github.com/textileio/basin/tree/main/sdk) to _both_ wasm _and_ to a Rust lib on the machine's native arch. the goal is to generate wasm bindings so we can use the wasm SDK to create a python and JS SDK (vs. developing these independently), ensuring they stay in-sync with the Rust implementation.

fwiw, we don't use any exports from `fvm_shared::crypto` directly in our SDK nor in our [custom actors](https://github.com/textileio/builtin-actors/tree/develop/actors/adm/src).

dtbuchholz:
also, adding in @carsonfarmer and @sanderpick for input

carsonfarmer:
> edit: but, we _do_ use it in our `signer` crate, which the SDK depends on: [here](https://github.com/textileio/basin/blob/main/signer/src/signer.rs). specifically, we need the `crypto::signature::Signature`.

Interestingly, it looks like only _some_ features of Signature require the crypto feature. So I don't think that should be the blocker: https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/crypto/signature.rs#L129

sanderpick:
@dtbuchholz and I were just chatting about this. He's going to try this suggestion here in a fork of `ref-fvm` for wasm architecture: https://github.com/paritytech/libsecp256k1/issues/134. Of [these non-wasm crates](https://github.com/filecoin-project/ref-fvm/blob/master/shared/Cargo.toml#L32), I think `libsecp256k1` is the only one we need to work in the JS environment.

dtbuchholz:
@raulk @sanderpick unfortunately, the fix above didn't work. as part of this attempt, i tried to fork `ref-fvm` and remove the `filecoin-proofs-api` dependency from the crypto feature flag. however, the[ `builtin-actors`](https://github.com/amazingdatamachine/builtin-actors/blob/e1f7b12e605a946d2f905d6e7edd0348abec6aa2/Cargo.toml#L111) need the [helix](https://github.com/helix-onchain/filecoin) dep, which uses `fvm_shared` and pulls in the proofs lib. 

i've opened an issue https://github.com/consensus-shipyard/ipc/issues/1038 and will move the discussion there.

raulk:
@dtbuchholz bummer! I wasn't able to access your fork of builtin-actors. Could you authorize my GitHub user?

raulk:
What exactly is depending on builtin-actors in your dependency graph for the Wasm support? It's rather strange, since the Filecoin builtin-actors were precisely created to be deployed on Wasm.

Issue #925, opened 2024-05-13
Gas completeness (EIP-1559, base fee oscillation, gas premiums, block gas limit)


maciejwitowski:
From @raulk:

TL;DR: The high level goal of this project is to complete the gas system in IPC. We propose to follow the EIP-1559 model by default, but providing customizability for the oscillation function / initial base fee / limits / more. This epic also fixes gas premium handling and burns.

## Status quo in IPC

- Base fee configured at genesis is static; it does not oscillate based on block utilization.
- IPC has no gas block limits yet (but it does have a configurable physical block size limit with a default of 21MiB, see below). Hence it cannot calculate compute utilization because it has no concept of limited capacity.
- When executing a tx, the subnet burns the network fee by transferring it to the burnt funds actor (f099).
- When executing a tx, ref-fvm credits premiums / priority fees to the reward actor (f02), but the client doesn't disburse it to the block proposer like it happens in Filecoin:
	- At the end of the block, Filecoin clients execute an `AwardBlockRewards` implicit transaction to withdraw the cumulative premiums + block reward to the miner.
- We have evidence that `eth_maxPriorityFeePerGas` and/or `eth_feeHistory` are miscalculating something and causing clients to propose extremely spiky priority gas fees.

## Proposed solution

- Investigate and fix `eth_maxPriorityFeePerGas` and/or `eth_feeHistory` to prevent clients from choosing spiky priority fees.
- Send the gas premium to the block proposer.
	- This may require linking a validator pubkey with an account.
	- Evaluate if we maintain or kill the reward actor indirection. This exists in Filecoin to dispense block rewards via inflation subject to locking/vesting, but IPC may choose a simpler path. On the other hand, the abstraction is worth preserving as an extension/customization point.
	- One way or another, I'm pretty sure we'll need to override the default reward actor with our own.
- Introduce a configurable _soft_ block gas limit (EIP-1559 target) and a _hard_ gas limit, as well as a physical maximum block size (ensure we have a maximum transaction size).
	- These can go in the chain metadata actor and be initialized at genesis, then be adjusted later by the subnet owner via on-chain transactions (without requiring an upgrade).
	- Ensure our block production logic only produces valid blocks.
	- Ensure our block validation only accepts valid blocks.
- Introduce a configurable base fee oscillation function, defaulting to EIP-1559.
[...]

PR #1034, opened 2024-06-17
fix: gas priority estimation

cryptoAtwill:
This PR fixes the gas priority estimation error mentioned in this [discussion](https://github.com/consensus-shipyard/ipc/issues/925#issuecomment-2115088510).


fendermint/eth/api/src/gas/mod.rs
```diff
@@ -68,8 +68,27 @@ pub fn median_gas_premium(prices: &mut [(TokenAmount, i64)], block_gas_target: i
 
     if prev2 != TokenAmount::zero() {
         premium += &prev2;
-        premium.div_ceil(BigInt::from(2));
+        premium.div_ceil(BigInt::from(2))
+    } else {
+        premium
     }
```
raulk, line 74: I haven't reviewed this full method and its intended usage in depth (and it looks a bit unidiomatic). But there does indeed seem to be a logical error here. That said, this looks more idiomatic?

```rust
let premium = if !prev2.is_zero() {
    (prev1 + &prev2).div_ceil(BigInt::from(2))
} else {
    prev1
}
premium
```

raulk (approved):
Clippy seems unhappy.

cryptoAtwill:
@raulk The clippy error comes from other parts of the code base, a quick separate [PR](https://github.com/consensus-shipyard/ipc/pull/1035) to fix that.

PR #1047, opened 2024-06-28
consolidate macros for measuring Eth request latency.

raulk:
Normalizes all methods to carry 2 arguments. Methods not requiring parameters accept a Params type with an empty tuple generic type.

---

Yeah, this kinda bites the bullet you didn't wanna bite @karlem. But it results in cleaner code overall.

### DIFF
--- a/fendermint/eth/api/src/apis/eth.rs
+++ b/fendermint/eth/api/src/apis/eth.rs
@@ -54,12 +54,15 @@ use crate::{
 /// Returns a list of addresses owned by client.
 ///
 /// It will always return [] since we don't expect Fendermint to manage private keys.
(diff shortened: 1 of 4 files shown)


karlem (approved):
Hey, @raulk thanks for the adjustment. I agree this looks much better. The only little thing that I would change is probably just name the `with_methods` macro something like `with_methods_and_latency` just to be super obvious, but I can change that in my branch.

raulk:
I would keep the name generic. It's decoupled. In the future we may want to add other middleware (eg tracing, access control, etc.) and naming all such behaviour won't scale well, and will result in overly specific macro names.

PR #1050, opened 2024-07-02
Vote tally commitment

cryptoAtwill:
This is a followup to #1037.

In this PR, it is switching to commitment or CID of the sealed topdown proposal as the voting medium. To be backward compatible, previously it is using block hash, which is 32 bytes. Once switched to CID, it's more than 32 bytes. This PR is using the above trick to determine the vote version. One can simply ignore the backward compatibility.


fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -248,9 +249,12 @@ where
             tracing::debug!(height, "adding data to the cache");
 
             self.provider.new_parent_view(height, Some(data.clone()))?;
-            self.vote_tally
-                .add_block(height, Some(data.0.clone()))
-                .map_err(map_voting_err)?;
+            if let Some(p) = self.provider.sealed_proposal_at_height(height)? {
```
karlem, line 252: Out of interest what does the seal_proposal_at_height do?

cryptoAtwill: It calculates the commitment of the proposal at the target height, kind of sealed it which means no more updates allowed.

PR #1052, opened 2024-07-02
specs: introduce a new observability model

raulk:
**View the [rendered](https://github.com/consensus-shipyard/ipc/blob/raulk/spec-observability/specs/drafts/observability.md) version.**

This PR introduces a spec to improve and evolve our observability features. The spec details:

- Our mental model.
- A list of events we seek to capture, including the traces and metrics they result in.
- A new journal abstraction for trace object logging, inspired by Lotus' (which I happened to design/implement 😆).
- A simplified implementation approach.


cryptoAtwill (approved):

karlem:
Looks good to me!

karlem (approved):

PR #1053, opened 2024-07-04
feat(node): new observability architecture + events

### LINKED ISSUE #1054: Introduce new observibility - Emit Top-Down Traces
#### Description
Design and implement the top-down tracing logic. Ensure all relevant events are captured and emitted correctly. Test that the subscriber works as expected.

#### Tasks
- Design the top-down tracing logic.
- Implement the tracing logic.
- Capture and emit relevant events.
- Test the subscriber functionality.

#### Acceptance Criteria
- Top-down traces are designed and implemented.
- Relevant events are correctly captured and emitted.
- Subscriber functionality is tested and verified.

Part of: #1055

karlem:
Close #1054


fendermint/vm/topdown/src/observe.rs
```diff
@@ -0,0 +1,240 @@
...
+        TOPDOWN_PARENT_FINALITY_VOTING_LATEST_RECEIVED_HEIGHT
+            .with_label_values(&[self.validator])
+            .set(self.block_height as i64);
+    }
+}
+
+#[derive(Debug)]
+pub struct ParentFinalityPeerVoteSent<'a> {
+    pub block_height: BlockHeight,
+    pub block_hash: BlockHashHex<'a>,
```
cryptoAtwill, line 112: why keep these two fields if they are not used?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -288,14 +314,30 @@ where
     /// Call when a new finalized block is added to the ledger, to clear out all preceding blocks.
     ///
     /// After this operation the minimum item in the chain will the new finalized block.
-    pub fn set_finalized(&self, block_height: BlockHeight, block_hash: V) -> Stm<()> {
+    pub fn set_finalized(
+        &self,
+        parent_block_height: BlockHeight,
+        parent_block_hash: V,
+        proposer: Option<&str>,
+        local_block_height: Option<BlockHeight>,
```
cryptoAtwill, line 322: what does this parameter mean? Like current fendermint block height?

fendermint/vm/topdown/src/voting.rs
```diff
@@ -418,6 +460,14 @@ pub async fn publish_vote_loop<V, F>(
                     if let Err(e) = client.publish_vote(vote) {
                         tracing::error!(error = e.to_string(), "failed to publish vote");
                     }
+
+                    let block_hash_hex = hex::encode(&next_hash);
+
+                    emit(ParentFinalityPeerVoteSent {
```
cryptoAtwill, line 466: why not handle hex conversion in the method? It says a lot of `hex::encode` and one does not really have to import hex because of tracing

fendermint/vm/topdown/src/proxy.rs
```diff
@@ -64,23 +68,36 @@ impl IPCProviderProxy {
...
+        .await
     }
 
     /// Getting the block hash at the target height.
     #[instrument(skip(self))]
     async fn get_block_hash(&self, height: BlockHeight) -> anyhow::Result<GetBlockHashResult> {
-        self.ipc_provider
-            .get_block_hash(&self.parent_subnet, height as ChainEpoch)
-            .await
+        emit_event_with_latency(
```
cryptoAtwill, line 91: I kind of feel this is a bit intrusive? It's forcing the app to emit event and install the crates in monitor?

karlem: I am not sure I follow. Could you please elaborate?

fendermint/vm/topdown/src/voting.rs
karlem: Yes exactly. I think it should be the Tendermint block. We can change the name if you find it confusing.

fendermint/vm/topdown/src/voting.rs
karlem: Good point!

fendermint/vm/topdown/src/observe.rs
karlem: They will be stored in the tracing event right.

fendermint/vm/topdown/src/proxy.rs
cryptoAtwill: I mean this changes the implementation of measure latency by default. Now, if I composite the current struct, `emit_event_with_latency` is called, maybe I want to measure the whole process?

 In java or node, we can use decorators. Here, maybe in the caller we use sth like at the outest level:
```rust
emit_event_with_latency(finality.get_genesis_epoch(), ...).await
```
so that we can easily remove it.

karlem: Yeah, I was considering to have another struct (implementing the same trait) that just wraps this one and measures the latency and emits the event. Would you be happier with that?

cryptoAtwill: sounds good, my main concern is the measurement stays with this implementation but cannot easily remove it.

karlem:
@cryptoAtwill Alright, I have resolved all the comments.

cryptoAtwill (approved):

fendermint/app/src/app.rs
```diff
@@ -620,11 +623,42 @@ where
         *guard = Some(state);
 
         let response = match result {
-            Err(e) => invalid_check_tx(AppError::InvalidEncoding, e.description),
+            Err(e) => {
```
raulk, line 626: Oh dear, these recursive results are such a telling symptom of just how unergonomic the nested interpreter design is /tech_debt 😢

fendermint/app/src/app.rs
```diff
@@ -620,11 +623,42 @@ where
...
+            }
             Ok(result) => match result {
-                Err(IllegalMessage) => invalid_check_tx(AppError::IllegalMessage, "".to_owned()),
-                Ok(Err(InvalidSignature(d))) => invalid_check_tx(AppError::InvalidSignature, d),
-                Ok(Ok(ret)) => to_check_tx(ret),
+                Err(IllegalMessage) => {
+                    emit(MpoolReceivedInvalidMessage {
+                        reason: "IllegalMessage",
+                        description: "",
+                    });
```
raulk, line 638: 1. What do you think about merging the `MpoolReceivedInvalidMessage` and `MpoolReceived` events into a single type (`MpoolReceived`), where `accept` determines whether it was accepted or not?
2. I would merge the reason and the description into a single field: `reason`, and [`format!()`](https://doc.rust-lang.org/std/macro.format.html) the original reason and description into it. You can then leverage the derived `Debug` implementation of the `AppError` enum to drop the inline strings (e.g. `"InvalidEncoding"`).
3. You can initialize a mutable struct at the top, update it throughout the nested `match`es, and emit it at the end, so the code is less pedantic.

fendermint/app/src/app.rs
```diff
@@ -650,7 +684,13 @@ where
             .context("failed to prepare proposal")?;
 
         let txs = txs.into_iter().map(bytes::Bytes::from).collect();
-        let txs = take_until_max_size(txs, request.max_tx_bytes.try_into().unwrap());
+        let (txs, size) = take_until_max_size(txs, request.max_tx_bytes.try_into().unwrap());
+
+        emit(BlockProposalSent {
+            height: request.height.value(),
+            tx_count: txs.len(),
```
raulk, line 691: It would be useful to print our own validator address here, which is part of `request::PrepareProposal`.

fendermint/app/src/app.rs
```diff
@@ -666,26 +706,40 @@ where
...
-            is_accepted: accept,
-            block_height: request.height.value(),
-            block_hash: request.hash.to_string().as_str(),
-            num_txs,
-            proposer: request.proposer_address.to_string().as_str()
+        let process_result = self.interpreter.process(self.chain_env.clone(), txs).await;
+
+        emit(BlockProposalReceived {
+            height: request.height.value(),
+            hash: HexEncodableBlockHash(request.hash.into()),
```
raulk, line 716: Looks like `HexEncodableBlockHash` type is repeated in two crates. We could move somewhere more common (e.g. `ipc-observability`), under a `serde` module.

fendermint/app/src/app.rs
```diff
@@ -666,26 +706,40 @@ where
...
+            size: size_txs,
+            tx_count: num_txs,
+            validator: request.proposer_address.to_string().as_str(),
```
raulk, line 719: The `account::Id` type implements `fmt::Display`, so it should be fine to use the type as-is.

fendermint/app/src/app.rs
```diff
@@ -666,26 +706,40 @@ where
...
+                Ok(response::ProcessProposal::Accept)
+            }
+            Err(e) => {
+                emit(BlockProposalRejected {
+                    height: request.height.value(),
+                    size: size_txs,
+                    tx_count: num_txs,
+                    validator: request.proposer_address.to_string().as_str(),
+                    reason: e.to_string().as_str(),
+                });
```
raulk, line 740: Similar to the above, we could merge these types into a single `BlockProposalEvaluated` type.

raulk (commented):
1/3 reviewed.

fendermint/app/src/app.rs
karlem: Yes indeed. I think it might be also worth to parse the transaction directly here - with help of external library.

fendermint/app/src/app.rs
```diff
@@ -620,11 +623,42 @@ where
...
+                Ok(Ok(ret)) => {
+                    emit(MpoolReceived {
+                        from: ret.message.from.to_string().as_str(),
+                        to: ret.message.to.to_string().as_str(),
+                        value: ret.message.value.to_string().as_str(),
+                        param_len: ret.message.params.len(),
+                        gas_limit: ret.message.gas_limit,
+                        fee_cap: ret.message.gas_fee_cap.to_string().as_str(),
+                        premium: ret.message.gas_premium.to_string().as_str(),
+                        accept: ret.exit_code.is_success(),
```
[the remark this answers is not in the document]
karlem: It does indeed. I needed some of the parameters to be strings (for instance, for metrics labels) and wanted to have generic events. But you are right; it would be more efficient to just use the types directly (the ones that implement Display).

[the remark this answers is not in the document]
raulk: Got it, can we attempt to preserve the types during emit, and push the conversion deeper in, to where it's actually needed (emitting the metric with labels)?

fendermint/app/src/app.rs
karlem: I wanted to follow your suggestion, but that would require making a lot of fields in MpoolReceived optional. This is particularly necessary for cases like InvalidEncoding, where the transaction fails to parse the raw bytes. However, I suppose we can live with the optional types.

fendermint/app/src/app.rs
karlem: Yes. I have changed it to be more DRY friendly!

fendermint/app/src/app.rs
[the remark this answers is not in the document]
karlem: Yes, I have already done it :)

fendermint/app/src/app.rs
karlem: Yes, good point. Added it.

fendermint/app/src/app.rs
karlem: Agreed, done.

fendermint/app/src/app.rs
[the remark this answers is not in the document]
karlem: Yeah, that's a very good point! It will also simplify the code give the suggestion about merging the events together.

fendermint/app/src/app.rs
karlem: Alright, I have implemented it using options.

Cargo.toml
```diff
@@ -168,6 +173,7 @@ ipc-provider = { path = "ipc/provider" }
 ipc-wallet = { path = "ipc/wallet", features = ["with-ethers"] }
 ipc_ipld_resolver = { path = "ipld/resolver" }
 ipc-types = { path = "ipc/types" }
+ipc-observability = { path = "ipc/observability" }
 ipc_actors_abis = { path = "contracts/binding" }
```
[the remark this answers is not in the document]
karlem: Created an [issue](https://github.com/consensus-shipyard/ipc/issues/1076)

fendermint/app/options/src/lib.rs
```diff
@@ -102,13 +103,44 @@ pub struct Options {
     #[arg(long, env = "FM_CONFIG_DIR")]
     config_dir: Option<PathBuf>,
 
+    // TODO Karel - move all FM_LOG_FILE* flags to a configuration file instead
```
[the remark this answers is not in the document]
karlem: Alright - there is an [issue](https://github.com/consensus-shipyard/ipc/issues/1071) for it. I can pick it up as long as we finish review of this PR.

fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -350,3 +351,39 @@ fn check_error(e: anyhow::Error) -> (ApplyRet, ActorAddressMap) {
...
+    pub fn execute_explicit(
+        &mut self,
+        msg: Message,
+    ) -> anyhow::Result<(ApplyRet, ActorAddressMap, Duration)> {
+        let start = Instant::now();
+        let result = self.state.execute_explicit(msg)?;
+        let duration = start.elapsed();
+        Ok((result.0, result.1, duration))
+    }
+}
```
raulk, line 389: Introducing a new type seems a bit heavy-handed to me. It's only used in 2 sites, and I think it's acceptable to add a helper function to the observability crate -- this would allow us to time any closure.

```rust
fn measure_time<F, T>(f: F) -> (T, std::time::Duration)
where
    F: FnOnce() -> T,
{
    let start = Instant::now();
    let result = f();
    let duration = start.elapsed();
    (result, duration)
}
```

raulk (commented):
2/3.

fendermint/vm/interpreter/src/bytes.rs
```diff
@@ -125,13 +126,17 @@ where
...
+        state: Self::State,
+        msgs: Vec<Self::Message>,
+    ) -> anyhow::Result<bool, ProcessError> {
         if msgs.len() > self.max_msgs {
             tracing::warn!(
                 block_msgs = msgs.len(),
                 "rejecting block: too many messages"
             );
-            return Ok(false);
+            return Err(ProcessError::TooManyMessages(msgs.len()));
```
raulk, line 139: This changes the return semantics of this method. In fact, this appears to be a breaking change. The caller (`App::process_proposal()`) would previously return a successful rejection to CometBFT, it now propagates the error.

I don't think the motivation (to capture a reason in the event) justifies the risk of a functional regression in this part of the codebase. Let's revert these changes and settle for not capturing the reason for the rejection in the trace.

raulk (changes requested):

raulk (approved):
Overall, great job here! This will bring about a massive improvement in observability and monitoring. Excited to see it operating live. Kudos on your first major epic, @karlem! 👏 

If you're able to get the changes addressed in the next hours, we should be able to merge today and release tomorrow with Axon r02! 🚀

fendermint/app/options/src/lib.rs
```diff
@@ -162,6 +194,17 @@ impl Options {
...
+    pub fn log_file_config(&self) -> FileLayerConfig {
+        FileLayerConfig {
+            enabled: self.log_file_enabled.unwrap_or(false),
+            directory: self.log_dir.clone(),
+            max_log_files: self.max_log_files,
+            rotation: self.log_files_rotation.clone(),
+            domain_filter: self.domains_filter.clone(),
+            events_filter: self.events_filter.clone(),
+        }
+    }
```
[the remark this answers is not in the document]
karlem: [Addressed here](https://github.com/consensus-shipyard/ipc/pull/1078)

fendermint/vm/topdown/src/voting.rs
```diff
@@ -278,6 +294,14 @@ where
             tracing::debug!(weight, quorum_threshold, "showdown");
 
             if weight >= quorum_threshold {
+                emit(ParentFinalityPeerQuorumReached {
+                    block_height: *block_height,
+                    block_hash: HexEncodableBlockHash(block_hash.as_ref().to_vec()),
+                    // TODO Karel - just placeholder - need to use real commitment once implemented
+                    commitment_hash: HexEncodableBlockHash(block_hash.as_ref().to_vec()),
```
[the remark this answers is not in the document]
karlem: Yes, good idea!

fendermint/vm/interpreter/src/fvm/state/exec.rs
karlem: Yeah I agree. This change was a reaction to this [comment here](https://github.com/consensus-shipyard/ipc/pull/1064#discussion_r1679114289)

fendermint/vm/interpreter/src/bytes.rs
karlem: Yes I guess that makes sense. To me though I did not see why would we want to have it succeed in the case. Also I have changed the error handling as well. But I understand why the risk on correct migration is higher then benefit of reason.

karlem:
@raulk Thanks! I have addressed all your comments, and it should be ready to go. However, please bear in mind that there are still two outstanding PRs that require review.

fendermint/app/options/src/lib.rs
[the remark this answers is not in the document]
karlem: https://github.com/consensus-shipyard/ipc/pull/1078 dropping it here to have the configuration PR linked. It's almost done!

fendermint/app/src/observe.rs
```diff
@@ -113,25 +114,48 @@ impl Recordable for BlockCommitted {
...
 }
 
 impl Recordable for MpoolReceived {
     fn record_metrics(&self) {
+        let from = self
+            .message
+            .as_ref()
+            .map(|m| m.from.to_string())
+            .unwrap_or("".to_string());
+
```
raulk, line 156: Let's drop the `from` label from the metric. It's unbounded cardinality, and Prometheus cannot deal with that. Read more here: https://prometheus.io/docs/practices/naming/#labels

fendermint/vm/interpreter/src/fvm/observe.rs
```diff
@@ -18,63 +20,40 @@ register_metrics! {
...
+        match self.purpose {
+            MsgExecPurpose::Check => EXEC_FVM_CHECK_EXECUTION_TIME_SECS.observe(self.duration),
+            MsgExecPurpose::Estimate => {
+                EXEC_FVM_ESTIMATE_EXECUTION_TIME_SECS.observe(self.duration)
+            }
+            MsgExecPurpose::Apply => EXEC_FVM_APPLY_EXECUTION_TIME_SECS.observe(self.duration),
+            MsgExecPurpose::Call => EXEC_FVM_CALL_EXECUTION_TIME_SECS.observe(self.duration),
+        }
     }
 }
```
raulk, line 55: Nice! It's all coming together. This would've been excrutiatingly difficult to do with the previous code.

fendermint/vm/topdown/src/observe.rs
```diff
@@ -214,27 +214,27 @@ mod tests {
...
-            commitment_hash: HexEncodableBlockHash(hash.clone()),
+            commitment_hash: Some(HexEncodableBlockHash(hash.clone())),
             weight: 0,
         });
 
         emit(ParentFinalityCommitted {
             parent_height: 0,
             block_hash: HexEncodableBlockHash(hash.clone()),
             local_height: Some(0),
-            proposer: Some("proposer"),
```
raulk: What's this here? Is this a TODO?

fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -177,11 +176,13 @@ where
 
     /// Execute message implicitly.
     pub fn execute_implicit(&mut self, msg: Message) -> ExecResult {
+        // TODO Karel - measure here
```
raulk, line 179: Do we _really_? I think it's fine not to overload this component with instrumentation, and just leave it up to the user -- just like we're doing now.

raulk (commented):
A few follow-up comments. I'm scrutinizing the ProcessResult and ProcessError changes as we speak, so expect one more final review of that in the next minutes.

fendermint/vm/topdown/src/observe.rs
raulk: Ah, it's just a test. Ignore my comment ;-)

fendermint/app/src/app.rs
```diff
@@ -666,27 +688,43 @@ where
...
         let txs: Vec<_> = request.txs.into_iter().map(|tx| tx.to_vec()).collect();
+        let size_txs = txs.iter().map(|tx| tx.len()).sum::<usize>();
         let num_txs = txs.len();
 
-        let accept = self
+        let process_result = self
             .interpreter
             .process(self.chain_env.clone(), txs)
-            .await
-            .context("failed to process proposal")?;
```
raulk, line 675: Dug into this further:

- Returning an error from the `Application` will induce a panic and cause the Tower Service to crash. 
- This approach is sensible if (a) Tower is configured to restart the service upon crash, AND (b) we're in the middle of some state update (e.g. after begin_block and before end_block). In that case, we want to reset transient state and start over.
- However, (b) is not true for process_proposal -- this is a stateless method. So returning an error here is unnecessary and problematic.

Conclusion:

- Thanks to your diff, we discovered a deeper issue! 🌟 
- But because this is a delicate part of the system, I prefer to extract this diff out to its own changeset and changelog entry.
- Behaviourally speaking, these changes seem reasonable, but I have some hesitations on introducing new types to signal a binary outcome (idiomatic way is to use Result or Option, either flattened or nested).

Next steps:

- Let's back out these changes and forego setting a reason for now.
- Open a separate PR with your changes that also prevents `process_proposal` from propagating an error.

raulk (approved):
Feel free to merge once the above comments have been addressed ;-)

fendermint/app/src/observe.rs
karlem: Good point.

fendermint/vm/interpreter/src/fvm/state/exec.rs
karlem: Not really yeah :) I forgot to remove my comment. It should not be there..

fendermint/vm/interpreter/src/fvm/observe.rs
karlem: Yes, it is also much cleaner this way.

PR #1061, opened 2024-07-09
feat(node): new observability bottom up tracing/metrics

### LINKED ISSUE #1056: Introduce new observibility - Emit Bottom-Up Traces
#### Description
Design and implement the bottom-up tracing logic. Ensure all relevant events are captured and emitted correctly. Test that the subscriber works as expected.

#### Tasks
- Design the bottom-up tracing logic.
- Implement the tracing logic.
- Capture and emit relevant events.
- Test the subscriber functionality.

#### Acceptance Criteria
- Bottom-up traces are designed and implemented.
- Relevant events are correctly captured and emitted.
- Subscriber functionality is tested and verified.

Part of: #1055

karlem:
Close #1056


fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -121,11 +123,16 @@ where
...
-        num_msgs,
-        next_configuration_number,
+    emit(CheckpointCreated {
+        height: height.value(),
+        hash: HexEncodableBlockHash(block_hash.to_vec()),
+        msg_count: num_msgs,
+        config_number: next_configuration_number,
+    });
+
+    emit(CheckpointFinalized {
```
karlem, line 133: @cryptoAtwill You will probably know this. I am not exactly sure we can say it's finalised at this point right? If not where else would be a good place to emit this event? Any idea please?

cryptoAtwill: You can do this in the relayer, once a quorum formed event is detected, it's a good time to emit this event

cryptoAtwill (approved):

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -186,6 +186,8 @@ where
     }
 
     async fn end(&self, mut state: Self::State) -> anyhow::Result<(Self::State, Self::EndOutput)> {
+        checkpoint::emit_trace_if_check_checkpoint_finalized(&self.gateway, &mut state)?;
```
raulk, line 189: If tracing fails, we should degrade gracefully instead of propagating an error through the critical path. After all, tracing is actually dispensable. Options:

1. Skip the trace silently. However, this leads to inconsistency as users might have set up monitoring for it.
2. Log the error as a warning. This gives info to the user to go hunting once they figure out that the finalization is actually happening, and it's only the trace that's not being printed.
3. Add a meta-event as a fallback to capture tracing errors in the journal. This is my preferred solution, as it actively communicates there is an error and keeps the journal consistent.

If you agree, let's go with (3). I'd suggest adding the following event:

```
struct TracingError {
	affected_event: String,
	reason: String,
}
```

And using `.inspect_err()` here to emit that trace on error.

I'd also add Counter metric for this: `tracing_errors[event=$affected_event]`.

WDYT?

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -255,6 +255,12 @@ where
             .await
             .context("failed to broadcast checkpoint signature")?;
 
+            emit(CheckpointSigned {
+                height: height.value(),
+                hash: HexEncodableBlockHash(cp.block_hash.to_vec()),
+                validator: validator_ctx.public_key,
+            });
```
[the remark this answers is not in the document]
karlem: Yes, exactly. I am aware this is not ideal since it won't publish other votes, but I thought having at least its own signature is better than nothing for now. However, your suggestion makes sense. I will also create a ticket to track it as a future update.

fendermint/vm/interpreter/src/fvm/exec.rs
karlem: Great point! I definitely think this should not propagate the error to the upper level. The best solution here, in my opinion, would be the proposed solution 3. The metrics are also a good idea, as they can help flag that the trace is not being saved without the need for proactive debugging.

karlem:
Thanks for the review @raulk. I have address all the issues mentioned above.

raulk (approved):
We're good to merge 🎉

PR #1063, opened 2024-07-11
Introduce New Observability - Configuration

### LINKED ISSUE #1057: Introduce new observibility -  Introduce Subscriber Configuration
#### Description
Implement the configuration for the subscriber. Ensure dynamic reconfiguration is supported. Test the configuration thoroughly.

#### Tasks
- Implement subscriber configuration.
- Support dynamic reconfiguration.
- Test the configuration functionality.

#### Acceptance Criteria
- Subscriber configuration is implemented.
- Dynamic reconfiguration is supported and functional.
- Configuration is thoroughly tested.

Part of #1055

karlem:
Close #1057 

There is a problem with using the configuration file because we want to set up tracing before the configuration is loaded. At least, this is how it currently works. Therefore, some kind of dynamic layer will need to be added to the subscriber layer when we want to use the configuration file. That is why I just used environment variables for now to cut corners and get this out as soon as possible. Once all metrics are implemented, I can fix the configuration reloading.

### DIFF
--- a/fendermint/app/src/main.rs
+++ b/fendermint/app/src/main.rs
@@ -3,9 +3,7 @@
 
 pub use fendermint_app_options as options;
 pub use fendermint_app_settings as settings;
-use ipc_observability::traces::{
-    register_tracing_subscriber, FileLayerConfig, RotationKind, WorkerGuard,
-};
+use ipc_observability::traces::{register_tracing_subscriber, WorkerGuard};
 
 mod cmd;
 
@@ -14,18 +12,9 @@ fn init_tracing(opts: &options::Options) -> Option<WorkerGuard> {
         .log_console_filter()
         .expect("invalid console level filter");
     let file_filter = opts.log_file_filter().expect("invalid file level filter");
+    let file_config = opts.log_file_config();
 
-    // TODO Karel - map the config to journal options
-    let journal_opts = FileLayerConfig {
-        enabled: true,
-        directory: Some("/Users/karlem/work/ipc/test-network/traces"),
-        max_log_files: Some(1),
-        rotation: Some(RotationKind::from("minutely")),
-        domain_filter: None,
-        events_filter: Some(vec!["ParentFinalityPeerVoteReceived"]),
-    };
-
-    register_tracing_subscriber(console_filter, file_filter, journal_opts)
+    register_tracing_subscriber(console_filter, file_filter, file_config)
 }
 
 /// Install a panic handler that prints stuff to the logs, otherwise it only shows up in the console.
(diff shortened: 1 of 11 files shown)


ipc/observability/src/tracing_layers.rs
```diff
@@ -0,0 +1,109 @@
...
+const DOMAIN_FIELD: &str = "domain";
+const EVENT_FIELD: &str = "event";
+
+pub struct DomainEventFilterLayer<L> {
+    domains: Option<Vec<String>>,
+    events: Option<Vec<String>>,
+    inner: L,
+}
+
+impl Deref for DomainEventFilterLayer<Registry> {
```
cryptoAtwill, line 18: I feel a `.register` getter is more suited for this, this is not supposed to be a smart pointer right?

ipc/observability/src/tracing_layers.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl<S, L> Layer<S> for DomainEventFilterLayer<L>
+where
+    S: Subscriber,
+    L: Layer<S>,
+{
+    fn on_event(&self, event: &Event, ctx: Context<S>) {
+        if self.domains.is_none() && self.events.is_none() {
```
cryptoAtwill, line 42: It seems like using `match (self.domain, self.events)` can simplify your downstream processing

fendermint/app/src/main.rs
```diff
@@ -3,30 +3,18 @@
...
-    register_tracing_subscriber, ConsoleLayerOpts, FileLayerOpts, WorkerGuard,
-};
+use ipc_observability::traces::{register_tracing_subscriber, WorkerGuard};
 
 mod cmd;
 
-fn init_tracing(_opts: &options::Options) -> Option<WorkerGuard> {
-    // let console_filter = opts.log_console_filter().expect("invalid filter");
-    // let file_filter = opts.log_file_filter().expect("invalid filter");
+fn init_tracing(opts: &options::Options) -> Option<WorkerGuard> {
```
cryptoAtwill, line 10: why not throw it out in `anyhow` and expect in `main`?

cryptoAtwill (approved):
Just some syntax suggestions that can simplify the code a bit.

fendermint/app/src/main.rs
karlem: Oh, because here we don't really care for an error. But rather the file guard might or might not be there (based on config) an either way is fine.

ipc/observability/src/tracing_layers.rs
karlem: I am not sure, would I not need to then match for all variats (none, none), (some, none), (none, some), (some, some) and essentially repeat the same filtering code all over?

ipc/observability/src/tracing_layers.rs
karlem: yes good point.

PR #1066, opened 2024-07-15
Topdown Enhancement: Adding validator quorum certificate to topdown gossip finality

cryptoAtwill:
In this series of PRs, an attempt is made to:
- [x] Adding validator signature to the gossip payload. This also includes updating the trait interfaces and other side effects (This PR). Previously the ipld-resolver forces any message gossiped to be signed by the a secret key, most likely being the validator key. This PR removes that restriction, any message can be published as long as it's validated by the libp2p peer secret key. Signature requirement is only enforced by application when needed. This makes way for future simplification with basic gossip pubsub from libp2p instead of existing ipld-resolver, which is under-utilised.
- [x] Collect quorum of signatures and forms a quorum certificate (This PR).
- [x] Add quorum signature into the topdown finality proposal, #1068.

This PR also commented off certain parts of the code base so that at least the code compiles. They will be updated in coming PRs.


fendermint/vm/topdown/src/voting/mod.rs
```diff
@@ -148,8 +156,8 @@ where
     pub fn add_block(
         &self,
         block_height: BlockHeight,
-        block_hash: Option<V>,
-    ) -> StmResult<(), Error<K>> {
+        payload: Option<TopdownVote>,
```
cryptoAtwill, line 159: I dont think `add_block` is ever needed, vote tally should directly get the data from `syncer` cache instead of storing its own data. Since `syncer` cache will be persisted, it will be much more deterministic then something in memory. Mark it as `todo` maybe.

PR #1090, opened 2024-07-26
feat(node): separate tracing and logging files

### LINKED ISSUE #1084: Segregate the event journal from general logs
Currently all traces are going to the global subscriber. This merges general logging with the journal. Instead, we want to segregate the journal and treat it differently. All traces emitted via `emit()` should go to a dedicated subscriber, whose formatting we can then configure separately without affecting general logging.

We'd love to do the same for the console, but that's a bit harder since we can only print to stdout and stderr (although, IIRC, we can output to file descriptor 3 if we really do need a third stream, but that's a pain to use). That said, we could prepend something to events printed to console via this dedicated subscriber so we can clearly distinguish them in the sea of other logs.

karlem:
Close #1084

### DIFF
--- a/ipc/observability/src/traces.rs
+++ b/ipc/observability/src/traces.rs
@@ -1,15 +1,16 @@
 // Copyright 2022-2024 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use crate::config::{FileLayerSettings, LogLevel, TracingSettings};
+use crate::tracing_layers::DomainEventFilterLayer;
 use std::num::NonZeroUsize;
 use tracing::Level;
 pub use tracing_appender::non_blocking;
 pub use tracing_appender::non_blocking::WorkerGuard;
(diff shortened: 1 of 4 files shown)


ipc/observability/src/traces.rs
```diff
@@ -22,32 +25,39 @@ pub fn create_temporary_subscriber() -> Subscriber {
...
-        .with_target(false)
-        .with_file(true)
-        .with_line_number(true)
-        .with_filter(console_filter);
-
-    let (file_layer, file_guard) = match &config.file {
+pub fn set_global_tracing_subscriber(
+    config: &TracingSettings,
+) -> Option<(WorkerGuard, WorkerGuard)> {
+    let (logs_file_layer, traces_file_layer, file_guards) = match &config.file {
```
raulk, line 31: The complexity of this method is too large. It's really hard to follow what it does, and we could document it better as well. I'm going to take a stab at improving this.

raulk (approved):

PR #1097, opened 2024-08-05
fix: update crate

cryptoAtwill:
Fix the clippy warning in CI due to crate versioning.


raulk:
Updating Cargo.lock should be enough here. @karlem is on it.

PR #1098, opened 2024-08-05
feat(node): adding gas actor

cryptoAtwill:
As part of the https://github.com/consensus-shipyard/ipc/issues/925, this PR only adds block gas limit.

The following changes are added:
- [x] Implemented `IPCGasActor` that tracks the gas related parameters in a separate actor.
- [x] Introduced a `GasLayer` trait that summaries the key gas functionalities.
- [x] Introduced a default gas implementation that reads the current block limit and deduct gas limit with new incoming txns.
- [ ] Initialise the block gas limit in genesis. This needs the depending genesis PRs to be merged.
- [ ] Adding block gas constraint in `prepare` and `process`.


fendermint/vm/interpreter/src/fvm/gas/default.rs
```diff
@@ -0,0 +1,108 @@
...
+            Ok(())
+        }
+    }
+
+    fn block_gas_limit(&self, state: &Self::State) -> anyhow::Result<Gas> {
+        let s =
+            read_actor_state::<State, DB>(state, fendermint_vm_actor_interface::gas::GAS_ACTOR_ID)?;
+        Ok(s.block_gas_limit())
+    }
+}
```
raulk, line 102: This shouldn't be accessing the state tree directly, but instead invoking a getter method on the actor for better encapsulation.

fendermint/vm/interpreter/src/fvm/gas/mod.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use fendermint_actor_gas::Gas;
+
+pub(crate) mod default;
+
+/// Handles the gas modeling in the current blockchain
+pub trait GasLayer {
```
raulk, line 9: - Let's call this `GasMarket`.
- Let's document that during normal operation, it's supposed to be backed by an actor.
- A more future-proof, pure interface for this:

```rust
pub trait GasMarket<State, Input> {
    /// Fetches the current state of the gas market.
    fn current() -> State;
    /// Updates the state of the gas market from the aggregated gas utilization stats for the block.
    fn update(Input) -> State;
}

struct State {
    block_gas_limit: Gas,
    /// This is the base_fee, but termed in a more abstract way so our terminology is not overfitted to EIP-1559.
    burn_fee_per_gas: TokenAmount,
}
```

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -57,6 +57,8 @@ where
         // Block height (FVM epoch) as sequence is intentional
         let height = state.block_height();
 
+        self.gas.reset_block_gas_quota(&mut state)?;
```
raulk, line 60: Why do we need a reset? We should be loading the current gas market state from the actor before we begin execution?

fendermint/actors/gas/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+        rt.transaction(|st: &mut State, _rt| {
+            st.block_gas_limit = limit;
+            Ok(())
+        })?;
+
+        Ok(())
+    }
+}
+
+impl ActorCode for IPCGasActor {
```
raulk, line 58: I'd call this EIP1559GasMarketActor if we're putting the EIP-1559 logic on chain.
We should define a common interface for gas market actors so that users can implement alternative gas markets and plug them in in their genesis.

fendermint/vm/interpreter/src/fvm/gas/mod.rs
cryptoAtwill: Sounds good, then the actor is really simple, just writing the state and reading the state.

fendermint/vm/interpreter/src/fvm/gas/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+    /// The state of the blockchain
+    type State;
+
+    /// Update the block gas limit. This will only take effect in the next block.
+    #[allow(dead_code)]
+    fn set_block_gas_limit(&self, state: &mut Self::State, limit: Gas) -> anyhow::Result<()>;
```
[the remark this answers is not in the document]
cryptoAtwill: I was thinking of a way to let the admin call this function to update block gas limit. But I think it works without this method as well, since one can just call the actor

fendermint/vm/interpreter/src/fvm/gas/default.rs
```diff
@@ -0,0 +1,108 @@
...
+use fvm_ipld_blockstore::Blockstore;
+use std::marker::PhantomData;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+type AtomicGas = AtomicU64;
+
+pub struct DefaultGas<DB> {
+    /// The total gas available to be used by transactions in the current block
+    /// The executor requires Send + Sync, using an atomic variable instead of u64.
+    block_gas_quota: AtomicGas,
```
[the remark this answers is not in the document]
cryptoAtwill: I didnt use the `limit` here because with each txn, the available block gas deducts the gas used. So that block gas limit just needs to be read once when a new block starts.

fendermint/vm/interpreter/src/fvm/exec.rs
cryptoAtwill: Yep, ideally it's like this. I tried doing this as well, but the current `ABCI` trait is blocking me from doing this because it's `begin(&self, ...)`. If we reset everytime, then we have to do:
```rust
self.gas = GasMarket::try_from(&mut state)?;
```
which requires `begin(&mut self)`.

I tried not to update so many changes and keep the interface consistent.

fendermint/vm/interpreter/src/fvm/gas/default.rs
```diff
@@ -0,0 +1,108 @@
...
+    p: PhantomData<DB>,
```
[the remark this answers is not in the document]
cryptoAtwill: updated!

cryptoAtwill:
@raulk I have updated for most of the review feedbacks. Some key changes:
- The `gas_market` still uses `reset_*` because of trait constraint, see comment: https://github.com/consensus-shipyard/ipc/pull/1098#discussion_r1706342099. It's the same idea, but just to make the compilation pass without too many changes.
- I removed the `GasLayer`/`GasMarket` trait for now as I'm not sure how this trait will look like with more functions added. I will add it back once all functions are clearer.
- I adopted the approach below, but used it in the actor. Since we are reading state in one shot, I just directly fetched from chain state, referencing `chainmetadata` actor.
```rust
pub trait GasMarket<State, Input> {
    /// Updates the state of the gas market from the aggregated gas utilization stats for the block.
    fn update(Input) -> State;
}
```

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
```diff
@@ -0,0 +1,108 @@
...
+use fendermint_vm_actor_interface::gas::GAS_ACTOR_ADDR;
+use fendermint_vm_actor_interface::system;
+use fvm_ipld_blockstore::Blockstore;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+type AtomicGas = AtomicU64;
+
+/// The gas market based on EIP1155
+#[derive(Default)]
+pub struct EIP1559GasMarket {
```
raulk, line 15: The actual gas strategy (EIP1559) should be fully encapsulated in the actor, so this is just an `ActorGasMarket` (a gas market driven by an on-chain actor), i.e. a gas market that resorts to an on-chain actor to keep its state and oscillate the gas parameters.

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// The block gas limit
+    block_gas_limit: AtomicGas,
+    /// The accumulated gas usage so far
+    block_gas_used: AtomicGas,
+}
+
+impl GasMarket for EIP1559GasMarket {
+    type State = EIP1559GasState;
+
+    fn reload_from_chain<DB: Blockstore + Clone + 'static>(
```
raulk, line 25: This should just be `load`.

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
```diff
@@ -0,0 +1,108 @@
...
+        &self,
+        chain_state: &FvmExecState<DB>,
+    ) -> anyhow::Result<()> {
+        let state = get_state(chain_state)?;
```
raulk, line 29: We need a query method in the `GasMarket` actor to fetch the current gas parameters (I called this `current()` in my proposal above), we should not dive into state directly as that is leaky.

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+impl EIP1559GasMarket {
+    fn update_block_gas_used(&self, old_used: Gas, new_used: Gas) -> anyhow::Result<()> {
+        self.block_gas_used
+            .compare_exchange(old_used, new_used, Ordering::SeqCst, Ordering::SeqCst)
+            .map_err(|_| anyhow!("concurrent update in block gas used, should not happen"))?;
+        Ok(())
+    }
```
raulk, line 88: Why do we need this dance between `consume_gas` and `update_block_gas_used`?

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
```diff
@@ -0,0 +1,108 @@
...
+
+type AtomicGas = AtomicU64;
+
+/// The gas market based on EIP1155
+#[derive(Default)]
+pub struct EIP1559GasMarket {
+    /// The block gas limit
+    block_gas_limit: AtomicGas,
+    /// The accumulated gas usage so far
+    block_gas_used: AtomicGas,
```
raulk, line 19: Is `AtomicGas` needed purely due to the ownership issues you pointed out on in our last sync? If so, please add a note here + a TODO, since this component is single-threaded and we should attempt to remove this overhead.

fendermint/vm/interpreter/src/fvm/gas/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use fvm_ipld_blockstore::Blockstore;
+
+pub mod eip1559;
+
+pub type Gas = u64;
+
+/// The gas market for fendermint. This should be backed by an fvm actor.
+pub trait GasMarket {
+    /// The gas market state
+    type State;
```
raulk, line 14: So the more I think about it, I think this trait should be state-type agnostic. The GasMarket should just have its own standard data model with its own struct, and the GasMarket implementation should adapt the concrete gas policy's details to that standard data model / interface.

raulk: Is it possible for the trait implementor to keep a reference to `FvmExecState` instead of having to pass it in? I think you can have a `new(&FvmExecState)` trait method, with the lifetime of the GasMarket conditioned to that of the exec state?

raulk: Or maybe it's not possible without an Rc, not sure.

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
cryptoAtwill: Purely because of the trait bound, I will add the note and todo

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
cryptoAtwill: I added the method to the gas actor and query using the `getter` method. 

But in fendermint I think we dont really have to add the method to `GasMarket` because it limits the trait to have a `State` parameter, which `GasMarket` does not really need to be on chain. I think it's nicer to just keep this as an implementation detail.

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
cryptoAtwill: good point, I was trying to be "correct" when using Atomic, but actually there is no need, a direct store will do. Updated.

fendermint/vm/interpreter/src/fvm/gas/mod.rs
cryptoAtwill: I have removed the `State` in `GasMarket`, it should be state type agnostic, technically no need to know about state and unnecessary.

I tried holding the `&mut FvmExecState` does not seem doable, because `FvmExecState<DB>` is destructed after each `deliver_tx`, then the `GasMarket` needs to be persisted every txn, which seems unnecessary. 

Technically there is no need to be aware of the `ChainState` for gas, but I also think this is part of a `ExecutionContext` or something like that to help determine the gas, implementation could just ignore.

fendermint/vm/interpreter/src/fvm/gas/eip1559.rs
cryptoAtwill: I moved `load` out of the trait. This method is only needed because of the `&self` trait bound, ideally the gas market should be recreated in `begin`. `GasMarket` does not really need to be aware of self loading from state. I shifted to the implementation of `ActorGasMarket`.

fendermint/vm/interpreter/src/fvm/gas/mod.rs
raulk: > I tried holding the `&mut FvmExecState` does not seem doable, because `FvmExecState<DB>` is destructed after each `deliver_tx`, then the `GasMarket` needs to be persisted every txn, which seems unnecessary.

Are you sure? `FvmExecState` seems to be maintained throughout `deliver_tx`, unless there's some nuance I'm unaware of? Did you try to make `gas_market` a member of `FvmExecState` instead of `FvmMessageInterpreter`?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -149,17 +152,22 @@ where
...
     ) -> anyhow::Result<(Self::State, Self::DeliverOutput)> {
         let (apply_ret, emitters, latency) = if msg.from == system::SYSTEM_ACTOR_ADDR {
             let (execution_result, latency) = measure_time(|| state.execute_implicit(msg.clone()));
             let (apply_ret, emitters) = execution_result?;
 
             (apply_ret, emitters, latency)
         } else {
+            // TODO: maybe compare the gas limits is better?
+            msg.gas_limit = msg.gas_limit.min(self.gas.available_block_gas());
+
```
raulk, line 165: This is better modelled as a `pre_exec` and a `post_exec` hook. If the gas limit exceeds the available gas, we should abort, not cap the execution gas. It is the block proposer's mistake, not the user's mistake (if you do this, you'd likely provoke an out of gas despite them having set the gas limit correctly).

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -149,17 +152,22 @@ where
...
+        mut msg: Self::Message,
     ) -> anyhow::Result<(Self::State, Self::DeliverOutput)> {
         let (apply_ret, emitters, latency) = if msg.from == system::SYSTEM_ACTOR_ADDR {
             let (execution_result, latency) = measure_time(|| state.execute_implicit(msg.clone()));
             let (apply_ret, emitters) = execution_result?;
 
             (apply_ret, emitters, latency)
         } else {
+            // TODO: maybe compare the gas limits is better?
+            msg.gas_limit = msg.gas_limit.min(self.gas.available_block_gas());
```
raulk, line 164: This is incorrect here. Instead, we should validate that the sum of all transaction gas limits doesn't exceed the block limit, at block validation time `prepare_proposal`. That guarantees we have enough gas to process all transactions even if they squeeze their gas limits to the max. Then this situation cannot occur.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -149,17 +152,22 @@ where
...
+
             let (execution_result, latency) = measure_time(|| state.execute_explicit(msg.clone()));
             let (apply_ret, emitters) = execution_result?;
 
+            self.gas.consume_gas(apply_ret.msg_receipt.gas_used)?;
```
raulk, line 169:
```suggestion
            self.gas.record_gas_used(apply_ret.msg_receipt.gas_used)?;
```

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -57,6 +58,8 @@ where
         // Block height (FVM epoch) as sequence is intentional
         let height = state.block_height();
 
+        self.gas.load(&mut state)?;
```
raulk, line 61:
```suggestion
        self.gas_market.load(&mut state)?;
```

raulk (commented):
This is on the right path, but needs a bit more work before we can land it. A few things missing in this PR. Some are covered by your TODOs in the PR description.

- Block validation logic (ProcessProposal): validate that the sum of transaction gas limits does not exceed the current block gas limit.
- Message validation logic (CheckTx): check that the gas limit does not exceed the block gas limit.
- Genesis: set up the GasMarketActor on genesis and inform CometBFT of the block gas limit during genesis.
- Block Gas Limit updates: inform CometBFT of block gas limit changes in `end_block`, so that it binpacks transactions to fit within the block gas limit.
- Tests: what do you have in mind here?
- Docs: explain how to configure the default (EIP1559) gas market on genesis, how to update it, and how to replace it with a custom gas market.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn available_block_gas(&self) -> Gas {
+        self.block_gas_limit.load(Ordering::SeqCst) - self.block_gas_used.load(Ordering::SeqCst)
+    }
+
+    fn consume_gas(&self, gas: Gas) -> anyhow::Result<()> {
+        let block_gas_used = self.block_gas_used.load(Ordering::SeqCst);
+
+        if block_gas_used + gas >= self.block_gas_limit.load(Ordering::SeqCst) {
+            anyhow::bail!("out of block gas")
+        }
```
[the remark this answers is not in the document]
cryptoAtwill: I captured this error outside and log it as warning.

fendermint/actors/gas_market/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "fendermint_actor_gas_market"
+description = "Builtin transaction gas tracking actor for IPC"
```
raulk, line 3:
```suggestion
description = "IPC system actor representing a subnet's gas market"
```

fendermint/actors/gas_market/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub base_fee: TokenAmount,
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug, Clone)]
+pub struct BlockGasUtilization {
+    pub block_gas_used: Gas,
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug, Clone)]
+pub struct SetConstants {
```
raulk, line 43:
```suggestion
pub struct Constants {
```

fendermint/actors/gas_market/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+pub enum Method {
+    Constructor = METHOD_CONSTRUCTOR,
+    CurrentGasReading = frc42_dispatch::method_hash!("CurrentGasReading"),
+    SetConstants = frc42_dispatch::method_hash!("SetConstants"),
+    UpdateUtilization = frc42_dispatch::method_hash!("UpdateUtilization"),
+}
+
+impl EIP1559GasMarketActor {
+    /// Creates the actor
+    pub fn constructor(rt: &impl Runtime, st: EIP1559GasState) -> Result<(), ActorError> {
```
raulk, line 60: This should take a ConstructorParams object, otherwise you're leaking the state object externally, which is highly undesirable as it breaks abstraction.

```rust
struct ConstructorParams {
    constants: Constants,
    initial_base_fee: TokenAmunt,
}

File: fendermint/actors/gas_market/src/lib.rs [hunk 22]
```diff
@@ -0,0 +1,209 @@
...
+    fn set_constants(rt: &impl Runtime, constants: SetConstants) -> Result<(), ActorError> {
+        rt.validate_immediate_caller_is(std::iter::once(&SYSTEM_ACTOR_ADDR))?;
+
+        rt.transaction(|st: &mut EIP1559GasState, _rt| {
+            st.block_gas_limit = constants.block_gas_limit;
+            Ok(())
+        })?;
+
+        Ok(())
+    }
```
raulk, line 74: We should also have a `get_constants` that returns the constants, invokable by any.

fendermint/app/options/src/genesis.rs
raulk, line 1: I know this has been open for a while, but it needs a rebase!

PR #1107, opened 2024-08-14
feat(node): block gas limit PR review changes

cryptoAtwill:
This is an update based on review feedbacks: https://github.com/consensus-shipyard/ipc/pull/1098


fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -152,21 +150,28 @@ where
...
             let (execution_result, latency) = measure_time(|| state.execute_implicit(msg.clone()));
             let (apply_ret, emitters) = execution_result?;
 
             (apply_ret, emitters, latency)
         } else {
-            // TODO: maybe compare the gas limits is better?
-            msg.gas_limit = msg.gas_limit.min(self.gas.available_block_gas());
+            if msg.gas_limit > state.gas_market().available_block_gas() {
+                bail!("gas limit exceed available block gas limit")
+            }
```
raulk, line 163: Failing here will cause a panic. Can this even happen? Aren't we checking in `process_proposal` that the sum of gas limits does not exceed the maximum block gas limit?

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,109 @@
...
+        self.block_gas_limit - self.block_gas_used
     }
 
-    fn consume_gas(&self, gas: Gas) -> anyhow::Result<()> {
-        let block_gas_used = self.block_gas_used.load(Ordering::SeqCst);
-
-        if block_gas_used + gas >= self.block_gas_limit.load(Ordering::SeqCst) {
+    fn record_gas_used(&mut self, gas: Gas) -> anyhow::Result<()> {
+        if self.block_gas_used + gas >= self.block_gas_limit {
             anyhow::bail!("out of block gas")
```
raulk, line 28: This will induce a panic. This cannot happen; if it does, we have an internal assertion breakage. We really don't want to threaten the liveness of the chain with a panic. It's probably best to simply to a saturating add up to the block gas limit, and log a warning when we're exceeding it.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,109 @@
...
             gas_limit: u64::MAX,
-            method_num: fendermint_actor_gas_market::Method::GetState as u64,
-            params: fvm_ipld_encoding::RawBytes::serialize(())?,
+            method_num: fendermint_actor_gas_market::Method::UpdateUtilization as u64,
+            params,
             value: Default::default(),
             version: Default::default(),
             gas_fee_cap: Default::default(),
             gas_premium: Default::default(),
         };
```
raulk, line 98: I think this should happen on `end_block`? Why did you choose to have it here?

fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -113,6 +114,8 @@ where
 
     /// Indicate whether the parameters have been updated.
     params_dirty: bool,
+
```
raulk, line 117:
```suggestion
/// Keeps track of block gas usage during execution, and takes care of updating
/// the chosen gas market strategy (by default an on-chain actor delivering EIP-1559 behaviour).
```

raulk (commented):
We're getting very close, @cryptoAtwill!

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: works for me.

fendermint/vm/interpreter/src/fvm/exec.rs
cryptoAtwill: maybe we can put a warning log instead as a sanity check.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
raulk: Ah got it, I was mislead by the name of this method.

raulk (approved):

PR #1108, opened 2024-08-15
feat(contracts): numerous improvements to deploy scripts (+).

raulk:
## Tolerating null rounds

Filecoin chains will occasionally produce null rounds. Ethers' transaction monitoring cannot handle skipped block heights. It will try to request intermediate blocks that don't exist. Lotus returns a JSON-RPC error when calling eth_getBlock* with a height corresponding to a null round. Ethers reacts with an exception, and this aborts the deployment process.

To overcome this situation, we inject a custom provider that catches this specific error, and returns an artificial empty block instead of an error. With this, hardhat-ethers can now cope with null rounds.

## Major refactors and cleanups for easier maintenance

- Adopted the hardhat-deploy plugin, which keeps track of deployments for us.
- Removed hand-rolled deployment tracking and codegen. We no longer codegen scripts; any script can resolve any contract now.
- All the mess under ops/deploy*.sh is now gone.
- Refactored all Hardhat tasks, simplified aggressively, deduplicated code, deleted dead code, and removed tons of cruft.
- Adopted Typescript typing.
- Upgrade scripts are also massively overhauled for simplicity.
- LoC delta: around -1000.

## TODO

- [ ] Update docs.

## Next steps

- [ ] Move to Hardhat Ignition. We should be able to get transaction deployment parallelism, but it requires bumping Hardhat and Ethers versions, which should be much simpler to do now than it was before.


contracts/extensions.ts
raulk, line 1: @cryptoAtwill @gvelez17 @avichalp here's the extension that injects a custom provider to deal with null rounds through the RPC.

contracts/tasks/lib/diamond.ts
raulk, line 1: The diamond stuff is so simple... it's insane how overcomplicated and dirty it was before.

contracts/tasks/deploy-gateway.ts
```diff
@@ -0,0 +1,92 @@
...
+import { HardhatRuntimeEnvironment, TaskArguments } from 'hardhat/types'
+import { Deployments } from './lib'
+
+// Customize your gateway parameters here.
+const gatewayConstructorParams = {
+    bottomUpCheckPeriod: 10,
+    activeValidatorsLimit: 100,
+    majorityPercentage: 66,
+    networkName: {
+        root: undefined, // Will be set later.
```
cryptoAtwill, line 11: As in the program will auto set it for you?

contracts/tasks/lib/diamond.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Computes the added facet cuts.
+     */
+    private computeAdded(current: Selectors, needed: Selectors): FacetCut[] {
+        const selectorsToAdd = Object.keys(needed).filter(
+            (selector) => !current[selector],
+        )
+
+        const addressToSelectors = _.groupBy(selectorsToAdd, (selector) => needed[selector])
```
cryptoAtwill, line 88: curious, why not use `import {groupBy} from "lodash"` (or is it supported) but use `_`?

contracts/tasks/deploy-gateway.ts
raulk: It's done below.

contracts/tasks/lib/diamond.ts
raulk: It's just a preference. I tend to dislike importing external functions in ways that lose the namespace at the usage sites.

contracts/extensions.ts
raulk: @xBalbinus @rk-rishikesh I'll submit this to the Hardhat FEVM kit so Filecoin devs can also benefit.

cryptoAtwill (approved):

PR #1113, opened 2024-08-19
docs(specs): subnet genesis v2.

raulk:


specs/drafts/genesis-v2.md
```diff
@@ -0,0 +1,174 @@
...
+      ## Contract deployment permissions.
+      contract_deployment_permissions:
+        ## Contract deployment permissioning mode. Values: 'unrestricted', or 'allowlist'.
+        mode: unrestricted | allowlist
+        ## If the mode is 'allowlist', specify a list of allowed addresses, or blank if none is allowed.
+        allowlist:
+          - addresses
+         
+## CIDs of actor manifests for built-in actors and for custom actors.
+actor_bytecode:
```
cryptoAtwill, line 116: so each time after processing, the data will be written to `blockstore`?

specs/drafts/genesis-v2.md
```diff
@@ -0,0 +1,174 @@
...
+For the most part, developers customize the genesis specification by manually editing the file, except for the
+following elements:
+
+- actor_bytecode/builtin
+- actor_bytecode/custom
+- system_contracts
+- initial_state_tree
+- blockstore
+
+Actor bytecode and system contracts are added using the `ipc genesis spec update` tool one or several of the following
```
cryptoAtwill, line 152: The update is not incremental update right? Write all for each call.

specs/drafts/genesis-v2.md
raulk: So when you update the genesis spec to add the bytecodes, you would be operating on the blockstore, yeah. The key idea here is that you import the genesis blockstore into the node as-is.

specs/drafts/genesis-v2.md
raulk: I didn't follow your question entirely, but the idea is that whatever thing you specify will overwrite the previous value for that thing. So if you set builtin actors bytecode, and then you go and call this command again, it will overwrite the previous value for that field. And it will maintain the blockstore succinct.

PR #1114, opened 2024-08-20
feat(node): gas market default initialisation in genesis

cryptoAtwill:
Adding a default initialisation of gas market in genesis. Once the spec for genesis is implemented, the parameters can be taken from genesis params. Currently it's following eth default params.


fendermint/actors/gas_market/src/lib.rs
```diff
@@ -8,103 +8,120 @@ use fil_actors_runtime::{actor_dispatch, ActorError};
...
 /// Elasticity multiplier as defined in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
-const EIP1559_ELASTICITY_MULTIPLIER: u64 = 2;
-/// Initial base fee as defined in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
-pub const EIP1559_INITIAL_BASE_FEE: u64 = 1_000_000_000;
+const ELASTICITY_MULTIPLIER: u64 = 2;
+lazy_static! {
+    /// Initial base fee as defined in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
+    static ref INITIAL_BASE_FEE: TokenAmount = TokenAmount::from_atto(1_000_000_000);
+    static ref MINIMAL_BASE_FEE: TokenAmount = TokenAmount::from_atto(100);
+}
```
raulk, line 26: How about we put these constants into an object in state, making them configurable via `SetConstants`?

The downside is that they'll be rewritten on every utilization update, and we'll need to read state on every update, thus leading to some minor overhead, but it's pretty negligible and leads to much better customization, so I think it's a tradeoff worth taking.

You can also take this object in the constructor, and feed these during genesis.

fendermint/actors/gas_market/src/lib.rs
```diff
@@ -118,15 +135,15 @@ impl ActorCode for EIP1559GasMarketActor {
 
     actor_dispatch! {
         Constructor => constructor,
-        SetBlockGasLimit => set_block_gas_limit,
-        GetState => get_state,
-        UpdateBlockGasConsumption => update_block_gas_consumption,
+        SetConstants => set_constants,
+        CurrentGasReading => current_gas_reading,
```
raulk, line 139: We know this is about gas, we don't need to be pedantic :-)

```suggestion
        CurrentReading => current_reading,
```

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -152,21 +150,28 @@ where
...
         let (apply_ret, emitters, latency) = if msg.from == system::SYSTEM_ACTOR_ADDR {
             let (execution_result, latency) = measure_time(|| state.execute_implicit(msg.clone()));
             let (apply_ret, emitters) = execution_result?;
 
             (apply_ret, emitters, latency)
         } else {
-            // TODO: maybe compare the gas limits is better?
-            msg.gas_limit = msg.gas_limit.min(self.gas.available_block_gas());
+            if msg.gas_limit > state.gas_market().available_block_gas() {
+                tracing::warn!("gas limit exceed available block gas limit");
```
raulk, line 162: Let's always try to be as informative as we can with log messages. With the previous text, an operator would've had to search for the log line to fully understand the context. This is more descriptive:

```suggestion
		// This is panic-worthy, but we suppress it to avoid liveness issues.
                tracing::warn!("[ASSERTION FAILED] message gas limit exceed available block gas limit; consensus engine is misbehaving");
```

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -152,21 +150,28 @@ where
...
             let (execution_result, latency) = measure_time(|| state.execute_explicit(msg.clone()));
             let (apply_ret, emitters) = execution_result?;
 
-            self.gas.consume_gas(apply_ret.msg_receipt.gas_used)?;
+            if state
+                .gas_market_mut()
+                .record_gas_used(apply_ret.msg_receipt.gas_used)
+                .is_err()
+            {
+                tracing::warn!("should not have exceeded block gas limit");
```
raulk, line 173: This is poor error handling logic. This code is making an assumption about what the nature of the error was, which may not hold up in the future. Remember: the GasMarket is customizable. Instead, this warning should print the inner error instead:

```suggestion
                tracing::warn!("[ASSERTION FAILED] gas market failed while recording utilization: {err}");
```

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,105 @@
...
-        self.block_gas_limit.load(Ordering::SeqCst) - self.block_gas_used.load(Ordering::SeqCst)
+        self.block_gas_limit - self.block_gas_used
     }
 
-    fn consume_gas(&self, gas: Gas) -> anyhow::Result<()> {
-        let block_gas_used = self.block_gas_used.load(Ordering::SeqCst);
-
-        if block_gas_used + gas >= self.block_gas_limit.load(Ordering::SeqCst) {
-            anyhow::bail!("out of block gas")
+    fn record_gas_used(&mut self, gas: Gas) -> anyhow::Result<()> {
```
raulk, line 26:
```suggestion
    fn record_utilization(&mut self, gas: Gas) -> anyhow::Result<()> {
```

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,105 @@
...
     /// The block gas limit
-    block_gas_limit: AtomicGas,
+    block_gas_limit: Gas,
     /// The accumulated gas usage so far
-    block_gas_used: AtomicGas,
+    block_gas_used: Gas,
 }
 
 impl GasMarket for ActorGasMarket {
     fn available_block_gas(&self) -> Gas {
```
raulk, line 22: These semantics are more extensible going forward, as you could have gas markets that treat different transaction types differently.

```suggestion
    struct Available {
        block_gas: Gas
    }

    fn available(&self) -> Available {
```

fendermint/vm/interpreter/src/fvm/gas/actor.rs
raulk, line 26: 1. So the error path above will never be triggered because this can never return an error. Let's remove the Result from the return type here?

2. I'm thinking we shouldn't saturate here. Instead, we should keep track of the gas usage we are being told, as is. Then we clamp at minimum (or do a saturating subtraction) on `available()` and clamp at maximum on `commit`.

3. I think that leads to better design, as we are tracking the truth, but we only externalise the value according to expectations.

4. Let's make the warning more useful and log our cumulative gas usage, block gas limit, and new usage reported?

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,105 @@
...
             from: system::SYSTEM_ACTOR_ADDR,
-            to: GAS_ACTOR_ADDR,
-            sequence: chain_state.block_height() as u64,
+            to: GAS_MARKET_ACTOR_ADDR,
+            sequence: block_height as u64,
             // exclude this from gas restriction
-            gas_limit: u64::MAX,
-            method_num: fendermint_actor_gas_market::Method::UpdateBlockGasConsumption as u64,
-            params,
+            gas_limit: fvm_shared::BLOCK_GAS_LIMIT,
```
raulk, line 46: I think other parts use max u64 for implicit executions?

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,105 @@
...
 
-    fn update_params<DB: Blockstore + Clone + 'static>(
-        &self,
-        chain_state: &mut FvmExecState<DB>,
-    ) -> anyhow::Result<()> {
-        let block_gas_used = self.block_gas_used.load(Ordering::SeqCst);
-        let params = fvm_ipld_encoding::RawBytes::serialize(block_gas_used)?;
-
+impl ActorGasMarket {
+    pub fn new<E: Executor>(
```
raulk, line 37: `new` tends to denote a light constructor, but here we're even invoking the FVM, so we better use something else here.

```suggestion
    pub fn create<E: Executor>(
```

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -1,123 +1,105 @@
...
+            params: fvm_ipld_encoding::RawBytes::default(),
             value: Default::default(),
             version: Default::default(),
             gas_fee_cap: Default::default(),
             gas_premium: Default::default(),
         };
 
-        let (apply_ret, _) = chain_state.execute_implicit(msg)?;
+        let raw_length = fvm_ipld_encoding::to_vec(&msg).map(|bz| bz.len())?;
+        let apply_ret = executor.execute_message(msg, ApplyKind::Implicit, raw_length)?;
```
raulk, line 56: I wish we could perform this call with `readonly`, as it would result in better safety guarantees. It's not possible with the current FVM APIs, right?

raulk (changes requested):

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: yeah, I dont see any readonly execution, but I think we can pass the store as readonly for some txns

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: But it has to be `i64::MAX`, otherwise an error is thrown. But I think `BLOCK_GAS_LIMIT` should be fine for this case, should not be very gas costly.

fendermint/actors/gas_market/src/lib.rs
cryptoAtwill: Can, the miminal base fee perhaps is constant worthy. They are just BigInt, should be ok. If the constants grow larger, I think asking the state to hold a hamt mapping using CID as key can work too.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
raulk: Using this value is incorrect. It is neither the gas limit configured for the subnet, nor MAX. It's just another arbitrary value inherited from Filecoin. I'll push a commit to switch to `i64::MAX`. Implicit executions are critical system executions; they cannot fail ever, so we must give them unbounded gas.

raulk: For example, Lotus uses `block_gas_limit * 10000` for Cron execution: 

https://github.com/filecoin-project/lotus/blob/a5cb674bb11d3a438e5b9d9ca3da218f2941a7ac/chain/consensus/compute_state.go#L132

fendermint/actors/gas_market/src/lib.rs
```diff
@@ -8,28 +8,42 @@ use fil_actors_runtime::{actor_dispatch, ActorError};
...
+    minimal_base_fee: TokenAmount,
+    /// Elasticity multiplier as defined in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
+    elasticity_multiplier: u64,
+    /// Base fee max change denominator as defined in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
+    base_fee_max_change_denominator: u64,
+}
 
 #[derive(Serialize_tuple, Deserialize_tuple, Debug, Clone)]
 pub struct EIP1559GasState {
+    block_gas_limit: Gas,
```
raulk, line 33: I don't get why the gas limit hasn't been moved into the constants. With this latest change, we can now configure everything _but_ the block gas limit. Is that what you intended? If so, why?

fendermint/vm/interpreter/src/genesis.rs
```diff
@@ -417,6 +417,23 @@ impl GenesisBuilder {
             )
             .context("failed to replace built in eam actor")?;
 
+        // currently hard code them for now, once genesis V2 is implemented, should be taken
+        // from genesis.
+        // initial base fee as defined in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
+        let initial_base_fee = TokenAmount::from_atto(1_000_000_000);
```
raulk, line 423: 1. These are Ethereum parameters, which are dimensioned very differently to Filecoin/FVM parameters.
2. In Ethereum, this parameter was set to transition from the legacy gas market to EIP-1559. 1e9 is 1 gwei. For comparison purposes, the base fee right now is 1.29gwei, so that's pretty standard for them.
3. However, this is happening in our genesis, so there is no reason to try to conform to some running market expectation, because there is none.
4. I would simplify and just set the minimum base fee as the initial base fee here.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -90,7 +87,7 @@ impl ActorGasMarket {
             to: GAS_MARKET_ACTOR_ADDR,
             sequence: block_height as u64,
             // exclude this from gas restriction
-            gas_limit: u64::MAX,
+            gas_limit: i64::MAX as u64,
```
raulk, line 90: Hm, what's the deal here?

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -45,9 +46,9 @@ impl ActorGasMarket {
             to: GAS_MARKET_ACTOR_ADDR,
             sequence: block_height as u64,
             // exclude this from gas restriction
-            gas_limit: u64::MAX,
-            method_num: fendermint_actor_gas_market::Method::CurrentGasReading as u64,
-            params: fvm_ipld_encoding::RawBytes::serialize(())?,
+            gas_limit: i64::MAX as u64,
```
raulk, line 49: What's the deal here?

raulk (commented):
Leaving a few comments here, we're 99.9% done ;-) I'm happy to push the remaining changes myself.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: You mean why `i64::MAX`? Because `u64::MAX` will cause an error, I think internally it's checking the u64 can be casted safely to u64.

PR #1118, opened 2024-08-21
fix(node): remove genesis interpreter

cryptoAtwill:
This is a follow up to consistent genesis PR. In that PR, the implementation has been migrate, but some tests are still using the old `GenesisInterpreter` and quite coupled/leaked the implementation into unit tests.

This PR removes `GenesisInterpreter` completely, update tests and encapsulates all the genesis requirements into new module.


fendermint/vm/interpreter/src/fvm/mod.rs
```diff
@@ -91,7 +87,7 @@ where
     pub fn new(
         client: C,
         validator_ctx: Option<ValidatorContext<C>>,
-        contracts_dir: PathBuf,
+        _contracts_dir: PathBuf,
```
raulk, line 90: Any reason we don't remove this arg?

cryptoAtwill: Good catch, I missed it. Was using it as a placeholder so that not to break so many code at once. Will remove it.

raulk (approved):
-900 LoC reduction, score! 🎉🎉🎉

PR #1122, opened 2024-08-22
feat(node): prepare + process + end block abci gas handling

cryptoAtwill:
This PR add gas handling to `prepare`, `process` and `end_block` abci calls. Some key changes:
- `prepare`: Prepare the proposal with gas limit check. From the [doc](https://docs.cometbft.com/v0.37/spec/abci/abci++_app_requirements#gas), cometbft will be able to ensure the block_gas_limit is enforce (that's what I inferred from the doc, somehow not 100% clearly), but still added as a sanity check.
- `process`: Same as `prepare`, checks the  block gas limit.
- Added `SetConstants` into the gas market trait. Currently they are not used, added `allow(dead_code)` for future usage. The idea is `SetConstants` will buffer the update and in end block will commit the changes. Before each `end_block` finishes, it will check if there are `SetConstants` called and if there is, it will fetch latest consensus params from cometbft, update accordingly and push to cometbft.

Update:
- Shifted tendermint import from interpreters to `App` so that interpreters are more tendermint agnostic.
- Added `ValidatorTracker` and `Client` to `App` so that communication with cometbft can be made to query consensus parameters.
- Replaced `tendermint::account::Id` from `FvmExecState` with `PublicKey`. The `id` is the first 20 bytes of public key hashed, which is not useful for deriving fvm addresses. Converting to `PublicKey` will make the downstream process much more helpful.
- Added `MessageSelector` trait to handle more generic message selection.


fendermint/vm/interpreter/src/chain.rs
```diff
@@ -526,3 +547,136 @@ fn relayed_bottom_up_ckpt_to_fvm(
...
+/// - Order by gas limit in descending order
+/// - Make sure total gas limit does not exceed the `total_gas_limit` parameter
+fn messages_selection(
+    msgs: Vec<ChainMessage>,
+    total_gas_limit: Gas,
+) -> anyhow::Result<Vec<ChainMessage>> {
+    let mut msgs_with_gas_limit = signed_msgs_with_gas_limit(msgs)?;
+
+    // sort by gas limit descending
+    msgs_with_gas_limit.sort_by(|a, b| b.1.cmp(&a.1));
```
cryptoAtwill, line 573: probably we dont need this or handle if block gas limit is hit.

fendermint/vm/interpreter/src/fvm/cometbft.rs
raulk, line 1: Could you elaborate on why we need all these abstractions? To me, this looks pretty much self-contained to the `App::end_block` method, so I'm skeptical about extracting it from there and creating new public types that will ever only be used in a single site.

fendermint/vm/interpreter/src/fvm/cometbft.rs
```diff
@@ -0,0 +1,115 @@
...
+        let params = self
+            .client
+            .consensus_params(tendermint::block::Height::try_from(state.block_height())?)
+            .await?
+            .consensus_params;
+        end_block.with_consensus(updates.apply(params));
+
+        Ok(())
+    }
+}
```
raulk, line 81: This looks rather over-engineered. What's wrong with having this code in `App::end_block`?

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -17,9 +18,21 @@ pub struct ActorGasMarket {
...
+    type Constant = SetConstants;
+
+    fn get_constants(&self) -> anyhow::Result<Self::Constant> {
+        todo!()
+    }
+
+    fn set_constants(&mut self, constants: Self::Constant) {
+        self.constant_update = Some(constants);
+    }
+
```
raulk, line 35: Are `GetConstants` and `SetConstants` ever used by the App? These are policy-specific methods for the user to invoke off band.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -69,13 +82,56 @@ impl ActorGasMarket {
...
+            params: fvm_ipld_encoding::RawBytes::serialize(constants)?,
+            value: Default::default(),
+            version: Default::default(),
+            gas_fee_cap: Default::default(),
+            gas_premium: Default::default(),
+        };
+        self.call_fvm(msg, executor)?;
+
+        Ok(())
+    }
```
raulk, line 129: We don't need to handle constant changes at all, these happen entirely as standard chain messages, right? Or that's what I thought?

The only thing we need to be careful with is not changing constants _in the middle of block execution_. So I think we'll need a message selection / prepare proposal / process proposal rule to enforce messages directed to the gas market actor to always be at the top of the block.

Does that make sense?

So given a data structure:

```rust
struct MessageOrderRule {
    from: Option<Address>,
    to: Option<Address>,
    precedence: i32,
}
```

We could have a lazy static:

```
lazy_static! {
    pub static ref MESSAGE_ORDER_POLICY: Vec<MessageOrderRule> = vec![
        MessageOrderRule{from: None, to: Some(Address::from_id(GAS_MARKET_ID), precedence: 1000},
}
```

And then we'd reoder messages during selection to satisfy the policy?

Actually, we could have this policy be a trait with methods to sort (prepare proposal) and validate order (process proposal).

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -96,13 +152,18 @@ impl ActorGasMarket {
             gas_premium: Default::default(),
         };
 
+        self.call_fvm(msg, executor)?;
```
raulk, line 155:
```suggestion
        self.apply_implicit_message(msg, executor)?;
```

fendermint/vm/interpreter/src/fvm/gas/actor.rs
```diff
@@ -96,13 +152,18 @@ impl ActorGasMarket {
...
         let raw_length = fvm_ipld_encoding::to_vec(&msg).map(|bz| bz.len())?;
         let apply_ret = executor.execute_message(msg, ApplyKind::Implicit, raw_length)?;
 
         if let Some(err) = apply_ret.failure_info {
             anyhow::bail!("failed to update EIP1559 gas state: {}", err)
         } else {
-            Ok(())
+            Ok(apply_ret)
         }
     }
```
raulk, line 168: Can be simplified (GitHub won't let me make a suggestion):
 
```rust
apply_ret.failure_info
    .map_or(Ok(apply_ret), |err| anyhow::bail!("failed to update EIP1559 gas state: {}", err))
```

fendermint/vm/interpreter/src/fvm/gas/mod.rs
```diff
@@ -11,6 +11,17 @@ pub struct Available {
...
+    /// The constant parameters that determines the readings of gas market, such as block gas limit.
+    type Constant;
+
+    #[allow(dead_code)]
+    fn get_constants(&self) -> anyhow::Result<Self::Constant>;
+
+    /// Update the constants of the gas market. If the gas market is actor based, then it's recommended
+    /// to flush at EndBlock.
+    #[allow(dead_code)]
+    fn set_constants(&mut self, constants: Self::Constant);
```
raulk, line 23: We don't need these.

raulk (changes requested):

fendermint/app/src/app.rs
```diff
@@ -610,13 +611,25 @@ where
...
             .interpreter
-            .prepare(self.chain_env.clone(), txs)
+            .prepare((self.chain_env.clone(), state), txs)
             .await
             .context("failed to prepare proposal")?;
 
         let txs = txs.into_iter().map(bytes::Bytes::from).collect();
+        // TODO: This seems leaky placed here, should be done in `interpreter`, that's where it's ipc
+        // TODO: aware, here might have filtered more important messages.
         let (txs, size) = take_until_max_size(txs, request.max_tx_bytes.try_into().unwrap());
```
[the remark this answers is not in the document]
cryptoAtwill: For example, if you checkout https://github.com/consensus-shipyard/ipc/blob/f4c67489325ec6aa9ded608b2be9118c89cd55e5/fendermint/vm/interpreter/src/chain.rs#L166
`ipc` critical messages are appended to the end of the list. If there are quite some messages, the `ipc` messages might be filtered out as messages are selected from the start.

Since we are introduce some form of message selection, I was looking at some way to put all the selection logic in one place instead of split into different parts of the codebase. I think passing the block limit into the interpreter works for me, at least we have message selection in one place, easier to debug, onboard and maintain.

fendermint/vm/interpreter/src/fvm/cometbft.rs
cryptoAtwill: Currently the cometbft `Client` is held by interpreter, so in[ App::end_block](https://github.com/consensus-shipyard/ipc/blob/f4c67489325ec6aa9ded608b2be9118c89cd55e5/fendermint/app/src/app.rs#L124), there is no access to client.

Also `state` is held in a [closure](https://github.com/consensus-shipyard/ipc/blob/f4c67489325ec6aa9ded608b2be9118c89cd55e5/fendermint/app/src/app.rs#L813), so adding to interpreter is less of a change now.

The `EndBlockUpdate` is added because `interpreter` expected `Output` associated type, currently it's [PowerUpdate](https://github.com/consensus-shipyard/ipc/blob/f4c67489325ec6aa9ded608b2be9118c89cd55e5/fendermint/vm/interpreter/src/fvm/exec.rs#L51). Having a struct is easier to pass around.

Now that you mentioned, maybe the direction to go is move all `tendermint` related operation to `App` crate (currently tendermint crates and fendermint crates are coupled), pave way for consensus agnostic implementation.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: Yep, they are not used by the `App` and called by system upgrade, maybe.

I added these two for "reference" implementation I guess. The `SetConstants` is also related to how `state` could detect potential gas market changes. I'm thinking of holding the changes in memory first, as reflected in the option field, then in `end_block` performs the commit in interpreter. Then send a summary of the changed to `App` which then can be propagated to cometbft. If we dont control `SetConstants`, we might lost the change or have to query the gas market over and over to detect changes.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: I think a trait is better, more general and flexible. Just holding a vec of `MessageOrderRule` traits implementations.

fendermint/vm/interpreter/src/fvm/gas/actor.rs
cryptoAtwill: I tried, seems the above does not compile due to partial move because `map_or` takes self which consumes `apply_ret.failure_info` and Ok(apply_ret) is partially moved. I tried the same with `as_ref`, but have the same issue. It works if I do `apply_ret.failure_info.take`, but perhaps not desired as it changes the failure info.

PR #1123, opened 2024-08-22
fix(core): set the default log level to INFO.

raulk:
The observability caused us to regress by setting the default level to trace. This also cleans up type conversion logic.


cryptoAtwill (approved):
nice, this will clean up console of the genesis bytes.

PR #1112, opened 2024-08-19
fix: genesis pr CI cleanup

cryptoAtwill:
Some clean up for CI.

A few issues that makes the CI very slow and causing rebuild:
- `ipc_actors_abis` built is not deterministic, this is probably from `ethers-rust`? The same rust binding, if one builds twice, the rust binding generated sometimes is different.
- Calling makefiles in different folders actually triggers duplicated compilation of solidity abis.

PR #1124, opened 2024-08-22
fix(ci): significantly speed up CI.

raulk:
- Consolidates the full build of contracts and Rust in a single job.
- Make Rust caches work, make contracts/ABI/bindings caches work.
- Produce a Docker image and publish as an artifact, to reuse in e2e tests.
- Separate e2e tests from unit tests.
- Delete independent build for ipc-resolver; everything is built at once to minimise overhead.
- Remove unused workflows.

## Notes

- The contract cache is still not working 100%. Not sure why Foundry doesn't accept the artifacts in out/ and cache/.

Supersedes https://github.com/consensus-shipyard/ipc/pull/1112.


cryptoAtwill (approved):

raulk:
Well, this first build was slow because it didn't get to leverage caches. The merge to master will be slow too. And after that, all builds will speed up.

PR #1125, opened 2024-08-23
chore(core): fix clippy/fmt + actually use caches

raulk:


raulk:
Great! Got a full cached build down to 25min, and I think we can do better by:
1. Building all targets (including tests, benches, etc.) at the top, so we cache their dependencies too.
2. Running e2e tests with the release profile so it utilises caches.

PR #1130, opened 2024-09-06
feat(node): support configurable staking collateral token

cryptoAtwill:
This PR makes an effort to support configuration different token for collateral. Previously only native token is allowed, now ERC20 can be used.

This PR introduces the following changes:
- Rename `SupplySource` to `GenericToken`, `SupplySourceHelper` to `GenericTokenHelper`.
- Introduced `collateralSource` parameter, this is where the token for collateral can be configured.
- Aligned `preFund`, `preRelease` and everything related to token to use `GenericToken`, removed all hardcoding of native tokens, including `LibStaking` and `ReleaseQueue`.

List of todos:
- [ ] Feature implementation
- [ ] Unit tests and integration tests
- [ ] Deployment script udpate
- [ ] `ipc-cli` update
- [ ] Calirabtion network tests


contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -116,14 +120,15 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
...
+    /// @param amount The amount of collateral provided as stake
+    function join(bytes calldata publicKey, uint256 amount) external payable nonReentrant whenNotPaused notKilled {
         // Adding this check to prevent new validators from joining
         // after the subnet has been bootstrapped, if the subnet mode is not Collateral.
         // We will increase the functionality in the future to support explicit permissioning.
         if (s.bootstrapped) {
             LibSubnetActor.enforceCollateralValidation();
         }
-        if (msg.value == 0) {
+        if (amount == 0) {
```
JulissaDantes, line 131: if the supplySource is native user is still obligated to send the msg.value number in the amount? is there a piece of logic somewhere that auto populates this field is the supplysource is native?

contracts/contracts/lib/GenericTokenHelper.sol
```diff
@@ -83,12 +88,12 @@ library SupplySourceHelper {
     /// triggered from the execution.
     /// This function the `safeTransfer` function used before.
```
JulissaDantes, line 89: Could you clarify this sentence?

cryptoAtwill: nice catch

contracts/contracts/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: yep, have to do that. This just standardises the interface. So maybe one day nft can be used as well.

contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -213,10 +222,10 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
 
         if (!s.bootstrapped) {
             LibStaking.withdrawWithConfirm(msg.sender, amount);
-            return;
+            s.collateralSource.transferFunds(payable(msg.sender), amount);
+        } else {
+            LibStaking.withdraw(msg.sender, amount);
```
JulissaDantes, line 227: This else statement should also contain `s.collateralSource.transferFunds(payable(msg.sender), amount);` no? I ran a test that unstakes when the subnet is bootstrapped and the validator didnt get their erc20 back

cryptoAtwill: so after the subnet is bootstrapped, that means any validator stake withdraw should not have occurred immediately because of safety issues. It has to go through the checkpointing and confirmed by the current commitee.

cryptoAtwill: @JulissaDantes check this out: https://github.com/consensus-shipyard/ipc/blob/6d2cc3a52451a04e31970484a482a6251b08ee8f/contracts/contracts/lib/LibStaking.sol#L667

It's buffered in a release queue. 

See tests: https://github.com/consensus-shipyard/ipc/blob/6d2cc3a52451a04e31970484a482a6251b08ee8f/contracts/test/integration/SubnetActorDiamond.t.sol#L1902

To withdraw stakes, a lot of checks have to go through.

JulissaDantes: oh, I think I'm little lost here, 
- what is the `current commitee.`? 
- does that mean there is somewhere else in the code where it gets check if any tokens are owed to a validator? 
- or does that mean that at the checkpoint library this is handled where they check unstake request and perform the token transfers?

And Im not clear here, my understanding is that a `bootstrapped` subnet is an active subnet that has the minimum required collateral and validators amount to run. So I still think that there is a path for a validator to just want to unstake certain amount of their stake after a while, where there are several checkpoints being made, and they just wish to reduce their staked amount, although if this is handled elsewhere I understand why its not the case, depending on the answers to the previous questions I guess.

contracts/test/integration/GatewayDiamond.t.sol
```diff
@@ -896,28 +896,29 @@ contract GatewayActorDiamondTest is Test, IntegrationTestBase {
...
-                    subnetId: SubnetID({root: ROOTNET_CHAINID, route: new address[](0)}),
-                    rawAddress: FvmAddressHelper.from(caller)
-                }),
-                IPCAddress({subnetId: destinationSubnet, rawAddress: FvmAddressHelper.from(caller)}),
-                1,
-                0
-            )
-        );
-    }
+    // TODO: this is no longer possible because EOA cannot be subnet
```
karlem, line 899: What is EOA? In any case - since it's not possible anymore we can remove this entirely?

ipc/api/src/evm.rs
```diff
@@ -180,6 +180,45 @@ macro_rules! bottom_up_msg_batch_conversion {
     };
 }
 
+/// The type conversion between different generic token types
+macro_rules! generic_token_conversion {
```
karlem, line 184:
```suggestion
macro_rules! asset_conversion {
```

ipc/cli/src/commands/subnet/create.rs
```diff
@@ -70,6 +64,35 @@ impl CreateSubnet {
...
+        token_address,
+    })
+}
+
+fn parse_collateral_source(arguments: &CreateSubnetArgs) -> anyhow::Result<Asset> {
+    let Some(ref kind) = arguments.collateral_source_kind else {
+        return Ok(Asset::default());
+    };
+
+    let token_address = if let Some(addr) = &arguments.collateral_source_address {
```
karlem, line 84: Do we want to enforce that when the token is ERC20 the address should be set?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -344,8 +345,10 @@ impl SubnetManager for EthSubnetManager {
         let contract =
             subnet_actor_manager_facet::SubnetActorManagerFacet::new(address, signer.clone());
 
-        let mut txn = contract.join(ethers::types::Bytes::from(pub_key));
-        txn.tx.set_value(collateral);
+        let mut txn = contract.join(ethers::types::Bytes::from(pub_key), U256::from(collateral));
+        txn = self.handle_txn_token(&subnet, txn, collateral, 0).await?;
+        // txn.tx.set_value(collateral);
```
karlem, line 350: Remove?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -369,8 +372,10 @@ impl SubnetManager for EthSubnetManager {
         let contract =
             subnet_actor_manager_facet::SubnetActorManagerFacet::new(address, signer.clone());
 
-        let mut txn = contract.pre_fund();
-        txn.tx.set_value(balance);
+        let mut txn = contract.pre_fund(U256::from(balance));
+        txn = self.handle_txn_token(&subnet, txn, 0, balance).await?;
+        // txn.tx.set_value();
```
karlem, line 377: Remove?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -419,8 +424,10 @@ impl SubnetManager for EthSubnetManager {
         let contract =
             subnet_actor_manager_facet::SubnetActorManagerFacet::new(address, signer.clone());
 
-        let mut txn = contract.stake();
-        txn.tx.set_value(collateral);
+        let mut txn = contract.stake(U256::from(collateral));
+        txn = self.handle_txn_token(&subnet, txn, collateral, 0).await?;
+        // txn.tx.set_value();
```
karlem, line 429: Remove?

ipc/provider/src/manager/evm/manager.rs
```diff
@@ -1020,6 +1041,29 @@ impl EthSubnetManager {
         }
     }
 
+    pub async fn handle_txn_token<B, D, M>(
```
karlem, line 1044: I think I am missing the point of this function. Could you please explain?

karlem (commented):
In general it looks good.

I just think that we do not have any tests with the collateral supply ERC20 source right?

contracts/contracts/subnet/SubnetActorRewardFacet.sol
```diff
@@ -7,10 +7,17 @@ import {ReentrancyGuard} from "../lib/LibReentrancyGuard.sol";
...
 
 contract SubnetActorRewardFacet is SubnetActorModifiers, ReentrancyGuard, Pausable {
+    using AssetHelper for Asset;
+
     /// @notice Validator claims their released collateral.
     function claim() external nonReentrant whenNotPaused {
-        LibStaking.claimCollateral(msg.sender);
+        uint256 amount = LibStaking.claimCollateral(msg.sender);
+        if (amount > 0) {
+            s.collateralSource.transferFunds(payable(msg.sender), amount);
```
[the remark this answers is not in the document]
cryptoAtwill: but you have to pass `collateralSource` into the function then?

cryptoAtwill:
@karlem what tests are you referring to? There are plenty of unit and integration tests for the contracts.

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: because different combination of collateral source and supply source will have different txn value, if that's the question?

contracts/contracts/lib/LibSubnetActor.sol
```diff
@@ -151,7 +153,34 @@ library LibSubnetActor {
...
+            s.supplySource.increaseAllowance(s.ipcGatewayAddr, g);
+            IGateway(s.ipcGatewayAddr).register{value: collateral}(g, collateral);
+        } else if (supplySourceNative && !collateralSourceNative) {
+            s.collateralSource.increaseAllowance(s.ipcGatewayAddr, collateral);
+            IGateway(s.ipcGatewayAddr).register{value: g}(g, collateral);
+        } else {
+            s.supplySource.increaseAllowance(s.ipcGatewayAddr, g);
+            s.collateralSource.increaseAllowance(s.ipcGatewayAddr, collateral);
+            IGateway(s.ipcGatewayAddr).register(g, collateral);
+        }
```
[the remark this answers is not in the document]
cryptoAtwill: Does the above syntax compile? 

Yeah, I dont like this as well. I prefer removing this function completely with the refactoring. If the above syntax works, then worth a try.

contracts/contracts/lib/AssetHelper.sol
```diff
@@ -192,15 +197,29 @@ library SupplySourceHelper {
...
+    }
+
+    function increaseAllowance(Asset memory self, address spender, uint256 amount) internal {
+        if (self.kind == AssetKind.ERC20) {
+            IERC20 token = IERC20(self.tokenAddress);
+            uint256 allowance = token.allowance(address(this), spender);
+            if (!token.approve(spender, allowance + amount)) {
+                revert IncreaseAllowanceFailed();
+            }
+        }
```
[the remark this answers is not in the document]
cryptoAtwill: I think this way might be more specific to handle `AssetKind` for ERC20. Like you mentioned below if there is a new variant, then the handling would be wrong.

contracts/contracts/lib/LibStaking.sol
```diff
@@ -506,9 +506,6 @@ library LibStaking {
         s.validatorSet.recordWithdraw(validator, amount);
         // confirm deposit that updates the confirmed collateral
         s.validatorSet.confirmWithdraw(validator, amount);
-
-        // release stake from gateway and transfer to user
-        payable(validator).sendValue(amount);
```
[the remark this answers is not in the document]
cryptoAtwill: I think we can leave `LibStaking` to just bookkeeping. Transfer outside might be better because even though we confirmed it, we might not transfer immediately.

contracts/test/integration/GatewayDiamond.t.sol
cryptoAtwill: I donno who wrote this test though. If I just remove it, not sure if there are missing tests.

ipc/cli/src/commands/subnet/create.rs
cryptoAtwill: I think contract should have this tested. I dont see much of a point in checking, simply that we donno if the address is indeed correct.

contracts/contracts/lib/LibStaking.sol
```diff
@@ -87,8 +88,6 @@ library LibStakingReleaseQueue {
             delete self.releases[validator];
         }
 
-        payable(validator).sendValue(amount);
```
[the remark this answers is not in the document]
cryptoAtwill: This comment is repeated four times, answering here once only.

Do you have a reason why you prefer release done here? Does that mean LibStaking should also handle accept funds?

To release in this method, one has to pass collateralSource into this function then?

I just think `LibStaking` should just be bookkeeping and that's it. Fund handling is handled by caller, because:
1. fund transfer can be aggregated from other parts of the codebase. 
2. even if we confirmed it, we might not transfer immediately.

contracts/contracts/subnet/SubnetActorRewardFacet.sol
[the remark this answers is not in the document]
cryptoAtwill: resolve first as it's duplicated below

contracts/contracts/lib/LibStaking.sol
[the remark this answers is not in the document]
cryptoAtwill: resolve first as it's duplicated below

contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -249,11 +258,12 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
                 s.genesisBalance[msg.sender] == 0;
                 s.genesisCircSupply -= genesisBalance;
                 LibSubnetActor.rmAddressFromBalanceKey(msg.sender);
-                payable(msg.sender).sendValue(genesisBalance);
+                s.collateralSource.transferFunds(payable(msg.sender), genesisBalance);
             }
 
             // interaction must be performed after checks and changes
             LibStaking.withdrawWithConfirm(msg.sender, amount);
+            s.collateralSource.transferFunds(payable(msg.sender), amount);
```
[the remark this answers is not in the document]
cryptoAtwill: resolve first as it's duplicated below

contracts/contracts/lib/AssetHelper.sol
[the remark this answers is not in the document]
karlem: In this case it could never be wrong even if you add more type. Because it always deals with one specific type.

contracts/contracts/lib/LibStaking.sol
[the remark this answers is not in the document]
karlem: 2 things here. If you want to use the LibStaking just for book keeping that's fine. But I would rename the functions at least. When the function is called 'withdrawWithConfirm' but you are still expected to the to transfer it is a bit confusing. Given that now thew transfer was included.

contracts/contracts/lib/LibStaking.sol
```diff
@@ -578,14 +575,21 @@ library LibStaking {
...
             } else {
                 uint256 amount = abi.decode(change.payload, (uint256));
+                address gateway = s.ipcGatewayAddr;
 
                 if (change.op == StakingOperation.Withdraw) {
                     s.validatorSet.confirmWithdraw(validator, amount);
                     s.releaseQueue.addNewRelease(validator, amount);
-                    IGateway(s.ipcGatewayAddr).releaseStake(amount);
+                    IGateway(gateway).releaseStake(amount);
                 } else {
```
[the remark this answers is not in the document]
karlem: What if we change the code and forget to adjust this line?

contracts/contracts/lib/LibSubnetActor.sol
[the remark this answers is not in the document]
karlem: It does. Give it a try

karlem:
> @karlem what tests are you referring to? There are plenty of unit and integration tests for the contracts.

@cryptoAtwill but they all use the native collateral source not the ERC20 right? Correct me if I'm wrong.

cryptoAtwill:
@karlem not really, there are permutation and combination, say: https://github.com/consensus-shipyard/ipc/blob/012fa69a95410962a1fe967e56d5bd531e1b6871/contracts/test/integration/SubnetActorDiamond.t.sol#L1942

ipc/provider/src/manager/evm/manager.rs
karlem: Alright I get it now. It is a bit unclear at first sight. Maybe comment would do.

ipc/cli/src/commands/subnet/create.rs
karlem: True, but we can also fail fast with nicer error here.

contracts/contracts/gateway/GatewayManagerFacet.sol
```diff
@@ -90,8 +95,7 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         }
 
         subnet.stake -= amount;
-
-        payable(subnet.id.getActor()).sendValue(amount);
+        SubnetActorGetterFacet(msg.sender).collateralSource().transferFunds(payable(msg.sender), amount);
```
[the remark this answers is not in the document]
cryptoAtwill: The fund flow is from Validator -> SubnetActor -> Gateway. Release is from Gateway -> SubnetActor -> ReleaseQueue (Locking) -> Validator.

You can argue that SubnetActor can manage the funds themselves and it makes sense. Anyways, this is planned as part of the contract refactor.

contracts/contracts/lib/LibStaking.sol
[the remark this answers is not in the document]
cryptoAtwill: any names you have in mind, actually I thought the current one is not bad, haha.

contracts/contracts/lib/LibStaking.sol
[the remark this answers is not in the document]
cryptoAtwill: updated accordingly

contracts/contracts/lib/LibSubnetActor.sol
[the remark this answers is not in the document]
cryptoAtwill: updated accordingly

ipc/provider/src/manager/evm/manager.rs
cryptoAtwill: comment added!

contracts/contracts/lib/AssetHelper.sol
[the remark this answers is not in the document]
cryptoAtwill: understand where you are coming from. But say if we have another asset, NFT, then perhaps the current style can prevent accidently calling allowance on NFT? Actually I dont see much diff here because either way it will fail due to no allowance/permission. I will take note to check this out if we have any new type, but in the foreseeable short/long run, I dont think much will change.

contracts/contracts/gateway/GatewayManagerFacet.sol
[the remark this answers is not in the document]
karlem: Yeah, I think storing it in gateway is quite impractical.

contracts/contracts/lib/AssetHelper.sol
```diff
@@ -192,15 +197,29 @@ library SupplySourceHelper {
...
```
raulk, line 223: 1. Just use `SafeERC20#safeIncreaseAllowance()`?
2. Singleton ifs like this are cleaner if written as a negative if with an early return.

contracts/contracts/lib/LibSubnetActor.sol
```diff
@@ -90,7 +92,7 @@ library LibSubnetActor {
                 emit SubnetBootstrapped(s.genesisValidators);
 
                 // register adding the genesis circulating supply (if it exists)
-                IGateway(s.ipcGatewayAddr).register{value: totalCollateral + s.genesisCircSupply}(s.genesisCircSupply);
```
raulk, line 93: Looks like we had a bug here, where we sent native coins representing the genesis circulating supply, even if the supply source was ERC20. Nice catch!

contracts/contracts/lib/LibSubnetActor.sol
```diff
@@ -151,7 +153,36 @@ library LibSubnetActor {
...
+            s.supplySource.increaseAllowance(s.ipcGatewayAddr, genesisCircSupply);
+        } else {
+            msgValue += genesisCircSupply;
+        }
+
+        if (!collateralSourceNative) {
+            s.collateralSource.increaseAllowance(s.ipcGatewayAddr, collateral);
+        } else {
+            msgValue += collateral;
+        }
```
raulk, line 183: This is leaky as mentioned elsewhere. I'm going to take a stab to contain this complexity inside the AssetHelper.

raulk (approved):
Will submit a PR to:

1. Enclose the allowance-related complexity inside the `AssetHelper`.
2. Remove the unnecessary error types.
3. Use `SafeERC20#safeIncreaseAllowance()` instead of implementing this logic ourselves.

This approval is conditional to reviewing and merging the upcoming PR.

raulk (approved):
Will submit a PR to:

1. Enclose the allowance-related complexity inside the `AssetHelper`.
2. Remove the unnecessary error types.
3. Use `SafeERC20#safeIncreaseAllowance()` instead of implementing this logic ourselves.

This approval is conditional to reviewing and merging the upcoming PR.

raulk:
@cryptoAtwill all blocking review comments addressed in: https://github.com/consensus-shipyard/ipc/pull/1147

Please review and merge if you agree, then feel free to merge this one.

@karlem when that's done, could you please make a release?

contracts/contracts/lib/AssetHelper.sol
```diff
@@ -60,35 +60,40 @@ library SupplySourceHelper {
             }
             // Safe arithmetic is not necessary because underflow is not possible due to the check above
             return finalBalance - initialBalance;
+        } else {
+            // now we are handling native token
+            if (msg.value < value) {
+                revert NoBalanceIncrease();
+            }
```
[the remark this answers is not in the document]
cryptoAtwill: if `msg.value == value`, we should pass the check. If `msg.value > value`, someone wants to do charity, maybe can just take it...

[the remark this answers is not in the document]
raulk: Best not to. It's stronger to have an invariant that gateway balance across all tokens and the native coin is zero for initialized subnets, than to have one where it can be non zero.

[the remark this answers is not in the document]
cryptoAtwill: Oh, I realized we have to use `msg.value > value`. The reason is caller might have msg.value more than just locked value, for other purposes as well. In that way, we can only enforce msg.value > value.

PR #1131, opened 2024-09-06
chore: update IntegrationTestBase.sol to delete unused variable

JulissaDantes:
Hi, I was going through this file and noticed that the subnet actor is initialized below, and this variable is not used.


JulissaDantes:
Nevermind, just saw it is used by other contracts

PR #1136, opened 2024-09-18
feat(node): new topdown process

cryptoAtwill:
This PR is the start of in total 4 PRs that implements the new topdown flow. The implementation is broken down as follows:
- [x] Implement the vote tally skeleton state machine (This PR). The skeleton state machine only outlines the state transition from `Paused -> Active -> Paused`. The other two recovery modes are yet to be implemented.
- [x] Implement the vote tally logic: #1139 
- [x] Implement the new syncer: #1151 
- [x] Integrate certified topdown checkpoint proposal into fendermint execution (prepare, process, execute): #1187 

For this PR, the key changes are:
- Using tokio single thread reactor mode
- Use an enum to represent the different states of the state machine
- Introduce operation mode state machine skeleton with `Active` and `Paused` operation mode
- Use events emitted from syncer to communicate with the state machine: `TopDownSyncEvent`

The state machine is as follow:
```mermaid
stateDiagram-v2
  [*] --> Paused : Process start
  Paused --> Active : Synced
  Active --> Recovery : Checkpoints quiet
  Recovery --> Active : Checkpoints observed
  Active --> Paused : Stopping
  Recovery --> Paused : Stopping
	
  state Recovery {
    [*] --> SoftRecovery
    SoftRecovery --> HardRecovery : Still no new topdown checkpoints
    SoftRecovery --> [*] : New checkpoints
    HardRecovery --> [*] : New checkpoints
  }
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1136)


fendermint/vm/topdown/src/vote/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn handle_request(&self, _req: VoteReactorRequest) {}
+
+    fn record_vote(&self, _vote: VoteRecord) {}
+
+    fn handle_event(&self, _event: TopDownSyncEvent) {}
+
+    fn process_external_request(&mut self, _metrics: &OperationMetrics) -> usize {
+        let mut n = 0;
+        while n < self.config.req_batch_processing_size {
+            match self.req_rx.try_recv() {
```
LePremierHomme, line 87: I'm not sure I understand the purpose of doing `try_recv()`. If the channel is empty, it will simply proceed to a fixed-duration sleep, which seems inefficient.

fendermint/vm/topdown/src/vote/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+                Ok(req) => {
+                    self.handle_request(req);
+                    n += 1
+                }
+                Err(mpsc::error::TryRecvError::Disconnected) => {
+                    tracing::warn!("voting reactor tx closed unexpected");
+                    break;
```
LePremierHomme, line 94: `break` will only break the batch loop, before re-entering. There's no shutdown / error handling mechanism.

fendermint/vm/topdown/src/vote/operation/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+mod active;
+mod paused;
+
+use crate::vote::operation::active::ActiveOperationMode;
+use crate::vote::operation::paused::PausedOperationMode;
+use crate::vote::VotingHandler;
+use std::fmt::Display;
+
+pub type OperationMode = &'static str;
```
LePremierHomme, line 12: Why using plain strings and not enum?

fendermint/vm/topdown/src/vote/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+struct VotingHandler {
+    /// Handles the requests targeting the vote reactor, could be querying the
+    /// vote tally status and etc.
+    req_rx: mpsc::Receiver<VoteReactorRequest>,
+    /// Receiver from gossip pub/sub, mostly listening to incoming votes
+    gossip_rx: broadcast::Receiver<VoteRecord>,
+    gossip_tx: mpsc::Sender<VoteRecord>,
+    /// Listens to internal events and handles the events accordingly
+    internal_event_listener: broadcast::Receiver<TopDownSyncEvent>,
```
LePremierHomme, line 73: `internal_event_rx`?

fendermint/vm/topdown/src/vote/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    QueryVotes(BlockHeight),
+}
+
+struct VotingHandler {
+    /// Handles the requests targeting the vote reactor, could be querying the
+    /// vote tally status and etc.
+    req_rx: mpsc::Receiver<VoteReactorRequest>,
+    /// Receiver from gossip pub/sub, mostly listening to incoming votes
+    gossip_rx: broadcast::Receiver<VoteRecord>,
+    gossip_tx: mpsc::Sender<VoteRecord>,
```
LePremierHomme, line 71: Currently unused, undocumented.

fendermint/vm/topdown/src/vote/operation/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+            current_mode: PAUSED,
+            previous_mode: INITIALIZED,
+        };
+        Self::Paused(PausedOperationMode { metrics, handler })
+    }
+
+    pub fn step(self) -> Self {
+        match self {
+            OperationStateMachine::Paused(p) => p.advance(),
+            OperationStateMachine::Active(p) => p.advance(),
```
LePremierHomme, line 65: The rationale behind the sequential phases within a single `advance()`, for both modes, is unclear. Can you elaborate more on that?

fendermint/vm/topdown/src/vote/operation/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+/// TODO: Soft and Hard recovery mode to be added
+pub enum OperationStateMachine {
+    Paused(PausedOperationMode),
+    Active(ActiveOperationMode),
+}
+
+/// Tracks the operation mdoe metrics for the voting system
+pub(crate) struct OperationMetrics {
+    current_mode: OperationMode,
+    previous_mode: OperationMode,
```
LePremierHomme, line 45: What's the purpose of tracking `previous_mode`?

fendermint/vm/topdown/src/sync/mod.rs
```diff
@@ -21,6 +21,14 @@ use fendermint_vm_genesis::{Power, Validator};
 
 pub use syncer::fetch_topdown_events;
 
+#[derive(Clone)]
+pub enum TopDownSyncEvent {
```
LePremierHomme, line 25: Can you elaborate more on this? it's unclear how it may be produced.

fendermint/vm/topdown/src/vote/operation/mod.rs
cryptoAtwill: mostly for debugging

fendermint/vm/topdown/src/vote/mod.rs
cryptoAtwill: Initially I'm thinking of using `select!`, but then the drawback I think would be unable to decide execution order. Especially when different channels need to be processed simultaneously. Another way I'm thinking of is directly implement `Future` to poll ourselves. But this way is similar to handling poll.

fendermint/vm/topdown/src/vote/operation/mod.rs
cryptoAtwill: I will wrap it into `strum`.

fendermint/vm/topdown/src/vote/operation/mod.rs
cryptoAtwill: sorry, I dont quite get the question.

fendermint/vm/topdown/src/sync/mod.rs
cryptoAtwill: The following PRs are explained more extensively. I will add some doc to this.

fendermint/vm/topdown/src/vote/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+            config,
+        };
+        let mut machine = OperationStateMachine::new(inner);
+        loop {
+            machine = machine.step();
+            tokio::time::sleep(sleep).await;
+        }
+    });
+
+    VoteReactorClient { tx }
```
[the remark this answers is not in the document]
cryptoAtwill: do you have an example that elaborates this more?

fendermint/vm/topdown/src/vote/mod.rs
cryptoAtwill: It's possible to reduce the sleep time to make the processing more responsive. In a way, the reactor loop should be priority based. I think `select!` is fair. I have encountered cases where more important messages are delayed.

LePremierHomme: This comment spans also https://github.com/consensus-shipyard/ipc/pull/1136#discussion_r1882055401 & https://github.com/consensus-shipyard/ipc/pull/1136#discussion_r1882029783.

`select!` would indeed imply random order/priority between channels processing. 
That's why in https://github.com/consensus-shipyard/ipc/pull/1136#discussion_r1882055401 I asked about the need for sequential steps: `req_rx.try_recv()`, then `gossip_rx.try_recv()`, then `internal_event_listener.try_recv()`. 

Unless there's a good reason for that, it just creates a more complicated design.

fendermint/vm/topdown/src/vote/operation/mod.rs
LePremierHomme: We can continue here: https://github.com/consensus-shipyard/ipc/pull/1136#discussion_r1886673600

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
LePremierHomme: Something like having some `new()` function, and a separate `run()` function which blocks indefinitley, and should only be run within a background-task runner like `tokio::spawn`. 

They can then be packaged into one function, like currently. but it will help to better define the in/out flow.

fendermint/vm/topdown/src/vote/mod.rs
LePremierHomme: Can you elaborate more on that? 

You don't really get this guarantee with the current design, only in the scope of a single `advance()` iteration, for a given batch processing size config (`req_channel_buffer_size`, `req_batch_processing_size`).

It makes sense to employ a design like this only if there's a timeline dependency between the state machine handling of different channels, which is a fragile design, and I don't think is the case. 

If the channels output timing assumes to be non-deterministic, it can similarly be handled by separate actors, or using `select!`.
If they are expected to be timed/ordered, please elaborate more on that.

fendermint/vm/topdown/src/vote/operation/active.rs
```diff
@@ -0,0 +1,58 @@
...
+        );
+
+        n = self.handler.process_gossip_subscription_votes();
+        tracing::debug!(num = n, status = self.to_string(), "handled gossip votes");
+
+        if n == 0 {
+            todo!("handle transition to soft recover")
+        }
+
+        while let Some(v) = self.handler.poll_internal_event() {
```
LePremierHomme, line 42: Why `while` is used here, while in `paused.rs`'s equivalent it's a simple `if`, and in other channels a batch size is defined?

fendermint/vm/topdown/src/vote/mod.rs
LePremierHomme: * i’m not sure about the need for `step()`, `advance()`, etc. it seems to be used to control the Paused/Active modes transition and to enable/disable certain operations. but i think it can be simplified, using some observability on that state transition. 
* i assume channels output timing to be arbitrary, and so the ordering of the processing can be considered random as well. access serialization, either with channels, or alternatively with mutex, is only to prevent data races (or, to get your code to compile). what i mean is that when you query a proposal, there’s no logical conditions on including a certain `add_vote` either before or after. that’s why i don’t think that trying to control the channels processing order, or introducing batch processing size, helps. 
* processing multiple channels with `select!` or overloading them onto one with multiple producers is roughly equivalent, because eventually all write events must be serialized to be processed.
* a potential optimization is to separate reads and writes, and allow reads parallelization instead of having it be serialized with the writes. but i'm not sure it makes sense, considering the read/write ratio.
* if we're concerned about the interpreter proposal query latency, we can employ a shared priority channel, and a separate worker to source it from the multiple channels.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
cryptoAtwill: I think for this simple function, it should be fine? If the initialisation becomes complex, I will convert it to a `new`/`run` methods.

fendermint/vm/topdown/src/vote/operation/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+///
+///   State: Recovery {
+///     [*] --> SoftRecovery
+///     SoftRecovery --> HardRecovery : Still no new topdown checkpoints
+///     SoftRecovery --> [*] : New checkpoints
+///     HardRecovery --> [*] : New checkpoints
+///   }
+/// TODO: Soft and Hard recovery mode to be added
+pub enum OperationStateMachine {
+    Paused(PausedOperationMode),
```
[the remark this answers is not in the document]
cryptoAtwill: `Paused` could be triggered by many other condition, mostly not just in syncing. Just not implemented yet.

fendermint/vm/topdown/src/vote/mod.rs
cryptoAtwill: @LePremierHomme I have updated accordingly. This PR only switches to `select!`, following one will update the rest to ensure a stronger guarantee.

LePremierHomme: Few remarks: 
* there should be some cancellation mechanism, at least for a graceful shutdown, even if currently unused.
* i’m still not sure about `OperationStateMachine` as the base state using `step()`/`advance()` to transition, when transitioning is only to disable/enable a subset of the handlings. perhaps this can be structured differently by spawning an additional actor which observes that state.

fendermint/vm/topdown/src/vote/operation/mod.rs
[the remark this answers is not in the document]
LePremierHomme: What I mean is that `Paused` isn't really idle, so this name is a bit confusing. I didn't refer to what triggers that mode, but to how it's best to describe it (given what it's doing).

fendermint/vm/topdown/src/vote/mod.rs
cryptoAtwill: For the first point, I think I can add them in the later PRs, since the structure is setup.
For the second one, I'm not so sure it's actually a problem at the current stage. Because each mode at the current stage is just a slightly different. Given more operation mode, it might be making sense to break it into different struct. But I feel we can do that pretty easily when the use cases are required.

fendermint/vm/topdown/src/vote/mod.rs
cryptoAtwill: if there are error, we just log it or maybe raise an alert, but the loop should continue running. As for shutdown, I think we can add it in `VoteReactorRequest`, should be pretty straightforward to do.

fendermint/vm/topdown/src/vote/mod.rs
cryptoAtwill: added doc

fendermint/vm/topdown/src/vote/operation/mod.rs
cryptoAtwill: updated accordingly.

fendermint/vm/topdown/src/vote/operation/mod.rs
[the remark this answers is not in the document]
cryptoAtwill: Paused pauses the handling internal events.
If we want to change the name, I'm open to any better naming but I feel Idle is not much different either. think we can do this from the doc first or at the final PR.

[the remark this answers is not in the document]
LePremierHomme: I didn't suggest a rename to `Idle`, but rather that `Paused` suggests a state of idleness, while that's not the case. 

Which doc are you referring to?

fendermint/app/src/cmd/run.rs
```diff
@@ -196,36 +197,13 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
-        }
-
-        tracing::info!("subscribing to gossip...");
+        info!("subscribing to gossip...");
         let rx = service.subscribe();
-        let parent_finality_votes = parent_finality_votes.clone();
-        tokio::spawn(async move {
-            dispatch_resolver_events(rx, parent_finality_votes, topdown_enabled).await;
-        });
+        let topdown_gossip_rx = IPLDTopdownGossipReceiver { rx };
```
LePremierHomme, line 202: Isn't `rx`/`client` suppose to be general purpose, and for the entire IPC hierarchy, but here it's being associated with the topdown process?

fendermint/app/src/cmd/run.rs
```diff
@@ -310,29 +259,65 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
-            {
-                Ok(_) => {}
-                Err(e) => tracing::error!("cannot launch polling syncer: {e}"),
-            }
-        });
+
+        let topdown_config = settings.ipc.topdown_config()?;
+        let config = fendermint_vm_topdown::Config {
+            syncer: ParentSyncerConfig {
+                request_channel_size: 1024,
```
LePremierHomme, line 275: Is this the best place to position default values?

fendermint/vm/topdown/src/launch.rs
```diff
@@ -0,0 +1,254 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+
+/// Run the topdown checkpointing in the background. This consists of two processes:
+/// - syncer:
+///     - syncs with the parent through RPC endpoint to obtain:
+///         - parent block hash/height
+///         - topdown messages
+///         - validator changes
+///     - prepares for topdown observation to be braodcasted
```
LePremierHomme, line 28: typo: braodcasted

fendermint/vm/topdown/src/launch.rs
```diff
@@ -0,0 +1,254 @@
...
+/// - voting:
+///     - signs/certifies and broadcast topdown observation to p2p peers
+///     - listens to certified topdown observation from p2p
+///     - aggregate peer certified observations into a quorum certificate for commitment in fendermint
+pub async fn run_topdown<CheckpointQuery, GossipTx, GossipRx, Poller, ParentClient>(
```
LePremierHomme, line 33: Can you add docs for the params?

fendermint/app/src/cmd/run.rs
cryptoAtwill: This function is topdown specific, even in the previous implementation.

PR #1139, opened 2024-09-20
feat(node): migrate vote tally 

cryptoAtwill:
Following previous PR, this one migrates the vote tally process over.

In this PR, the following key changes are made:
- Migrating current main VoteTally over. Most of the logic is unchanged. The key changes are removing the `stm` module, removing `chain` field (not relevant to vote tally anymore) and remove `pause_votes` (not needed). Also `dump_votes` that dump all the votes currently collected is added. This is mainly to aid debugging in case of stuck finality.
- Implemented the `VoteTallyClient` that utilises tokio channels to communicate with the underlying reactor.
-  Introduced `VoteStore` trait that stored votes collected. Currently it's in memory store by default. Will merge with #897 when recovery modes are introduced.
- Introduced `GossipClient` trait that vote tally uses to communicate with the underlying libp2p gossip channel. Previously it's using `ipld-resolver` directly. But that crate is under utilised and quite heavy. Introducing this trait to decouple with `ipld-resolver` for easier testing as well.
- Added `RecoverableECDSASignature` to validate signed votes from peers and sign votes for peers.
- Vote is now versioned. The first byte is the version number, the rest of the bytes are the payload depending on the version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1139)


fendermint/vm/topdown/src/observation.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    pub fn ensure_valid(&self) -> anyhow::Result<ValidatorKey> {
+        let to_sign = fvm_ipld_encoding::to_vec(&self.observation)?;
+        let (pk1, _) = self.observation_signature.recover(&to_sign)?;
+
+        let p = Self::envelop_payload(&self.observation_signature, self.certified_at)?;
+        let (pk2, _) = self.signature.recover(p.as_slice())?;
+
+        if pk1 != pk2 {
+            return Err(anyhow!("public keys not aligned"));
```
LePremierHomme, line 65: Maybe add the pks into the err string, to allow some debugging.

fendermint/vm/topdown/src/vote/operation/active.rs
```diff
@@ -35,15 +34,19 @@ impl ActiveOperationMode {
...
+{
+    pub(crate) async fn advance(mut self) -> OperationStateMachine<T, S, V> {
         loop {
             select! {
                 Some(req) = self.handler.req_rx.recv() => {
-                    self.handler.handle_request(req);
+                    self.handler.handle_request(req, &self.metrics);
                 },
-                Ok(vote) = self.handler.gossip_rx.recv() => {
+                Ok(vote) = self.handler.gossip_rx.recv_vote() => {
```
LePremierHomme, line 49: Where does the `Vote::v1_checked` validation is taken place?

fendermint/crypto/src/secp.rs
```diff
@@ -0,0 +1,100 @@
...
+use crate::SecretKey;
+use anyhow::Context;
+use libsecp256k1::{recover, sign, PublicKey, RecoveryId};
+use multihash::{Code, MultihashDigest};
+
+#[cfg(feature = "with_serde")]
+use serde::de::Error;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct RecoverableECDSASignature((u8, [u8; 64]));
```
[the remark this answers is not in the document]
cryptoAtwill: This is the certificate, so we have to include them. 

Because the private key is secp. If you are concerned about the size, there is a PR to change the scheme to schnorr signature scheme. But at the time of my research, seems no stable library for that one. Will replace this accordingly when that happens.

fendermint/vm/topdown/src/observation.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub(crate) parent_hash: Bytes,
+    /// A rolling/cumulative commitment to topdown effects since the beginning of
+    /// time, including the ones in this block.
+    pub(crate) cumulative_effects_comm: Bytes,
+}
+
+/// A self-certified observation made by a validator.
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+pub struct CertifiedObservation {
+    observation: Observation,
```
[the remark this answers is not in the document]
cryptoAtwill: It's required in `Observation` because we need it to publish a cert. But for `certified_at` field, we also need it to be nonforgeable but no need to be part of the certificate. In the future, there might be more fields to be added to counter different attacks.

fendermint/vm/topdown/src/vote/operation/active.rs
cryptoAtwill: I delegated to `gossip`, maybe it's a good idea to handle it here.

fendermint/vm/topdown/src/vote/mod.rs
```diff
@@ -1,74 +1,307 @@
...
             machine = machine.step().await;
         }
     });
 
-    VoteReactorClient { tx }
+    Ok(VoteReactorClient { tx })
+}
+
+impl VoteReactorClient {
+    async fn request<T, F: FnOnce(oneshot::Sender<T>) -> VoteReactorRequest>(
```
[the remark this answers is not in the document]
cryptoAtwill: would you mind elaborate a bit?

fendermint/vm/topdown/src/observation.rs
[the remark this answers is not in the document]
LePremierHomme: If the latter is just part of the messaging protocol for authentication, perhaps it's better to abstract it away from that struct which defines the message content.

fendermint/vm/topdown/src/observation.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::{BlockHeight, Bytes};
+use anyhow::anyhow;
+use arbitrary::Arbitrary;
+use fendermint_crypto::secp::RecoverableECDSASignature;
+use fendermint_crypto::SecretKey;
+use fendermint_vm_genesis::ValidatorKey;
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+
+/// The content that validators gossip among each other.
```
LePremierHomme, line 13: This isn't the gossiped message, but rather contained in the `CertifiedObservation` message, which was documented with "A self-certified observation made by a validator", which seems like the reverse.

fendermint/vm/topdown/src/vote/store.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        for v in self.0.iter() {
+            let validator = v.voter();
+
+            let power = power_table.get(&validator).cloned().unwrap_or(0);
+            if power == 0 {
+                continue;
+            }
+
+            if let Some(w) = votes.iter_mut().find(|w| w.0 == v.observation()) {
```
LePremierHomme, line 112: Can use a hashmap to avoid this iteration (and use `or_insert_with` to simplify).

fendermint/vm/topdown/src/vote/store.rs
```diff
@@ -0,0 +1,187 @@
...
+impl VoteStore for InMemoryVoteStore {
+    fn earliest_vote_height(&self) -> Result<Option<BlockHeight>, Error> {
+        Ok(self.votes.first_key_value().map(|(k, _)| *k))
+    }
+
+    fn latest_vote_height(&self) -> Result<Option<BlockHeight>, Error> {
+        Ok(self.votes.last_key_value().map(|(k, _)| *k))
+    }
+
+    fn store_vote(&mut self, height: BlockHeight, vote: Vote) -> Result<(), Error> {
```
LePremierHomme, line 48: Why storing `Vote`, and not merely `Observation`? AFAIU only the latter is needed going forward.

fendermint/vm/topdown/src/vote/tally.rs
```diff
@@ -0,0 +1,418 @@
...
+        let quorum_threshold = self.quorum_threshold();
+        let Some(max_height) = self.votes.latest_vote_height()? else {
+            tracing::info!("vote store has no vote yet, skip finding quorum");
+            return Ok(None);
+        };
+
+        for h in ((self.last_finalized_height + 1)..=max_height).rev() {
+            let votes = self.votes.get_votes_at_height(h)?;
+
+            for (observation, weight) in votes.observation_weights(&self.power_table) {
```
LePremierHomme, line 156: What about doing this indexing when when storing, and not when attempting to fetch, which could be latency sensitive?

fendermint/vm/topdown/src/vote/store.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn into_owned(self) -> Vec<Vote> {
+        self.0.into_iter().cloned().collect()
+    }
+
+    pub fn observation_weights(&self, power_table: &PowerTable) -> Vec<(&Observation, Weight)> {
```
LePremierHomme, line 101: Better to make the returned list ordered by the weight, so you won't need to iterate over it for to find a `> quorum_threshold` item.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
LePremierHomme: Something like having `VoteReactorClient::request` to construct and pass the response channel without having to `VoteReactorRequest` to embed it. However after trying this out myself, it ended up with more verbose than your version ;) 

So I'd only suggest to embed the response type in a uniform way and reduce the enum polymorphism, something like:
```rust
type Response<T> = oneshot::Sender<T>;

enum VoteReactorRequest {
    /// A new child subnet block is mined, this is the fendermint block
    NewLocalBlockMined {
        height: BlockHeight,
        res: Response<()>,
    },
    /// Query the current operation mode of the vote tally state machine
    QueryOperationMode {
        res: Response<OperationMetrics>
    },
    /// Query the current validator votes at the target block height
    QueryVotes {
        height: BlockHeight,
        res: Response<Result<Vec<Vote>, Error>>,
    },
}
```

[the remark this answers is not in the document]
LePremierHomme: Something like having `VoteReactorClient::request` to construct and pass the response channel without having to `VoteReactorRequest` to embed it. However after trying this out myself, it ended up more verbose than your version ;) 

So I'd only suggest to embed the response type in a uniform way and reduce the enum polymorphism, something like:
```rust
type Response<T> = oneshot::Sender<T>;

enum VoteReactorRequest {
    /// A new child subnet block is mined, this is the fendermint block
    NewLocalBlockMined {
        height: BlockHeight,
        res: Response<()>,
    },
    /// Query the current operation mode of the vote tally state machine
    QueryOperationMode {
        res: Response<OperationMetrics>
    },
    /// Query the current validator votes at the target block height
    QueryVotes {
        height: BlockHeight,
        res: Response<Result<Vec<Vote>, Error>>,
    },
}
```

fendermint/vm/topdown/src/vote/tally.rs
cryptoAtwill: Good call. I was thinking of create a field:
 `quorum: Arc<RwLock<Option<Vec<CertifiedObservation>>>`
that is created during storing, so when calling `find_quorum`, it immediate returns. The only consideration I has was it would change the current logic of always finding the highest height. But on the other hand, we can enforce the quorum should not reduce in height. I'm not sure if we should change the implementation now. But this is a good idea.

fendermint/vm/topdown/src/vote/store.rs
cryptoAtwill: We still need the validator key

fendermint/vm/topdown/src/vote/store.rs
cryptoAtwill: I just returned the max weight that's what we care really.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
cryptoAtwill: yeah, it's more or less the same

fendermint/vm/topdown/src/observation.rs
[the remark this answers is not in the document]
cryptoAtwill: I tried something like:
```rust
struct RecoverablePayload<T> {
  t: T,
  sig: RecoverableSig
}
```
But it becomes more nested than the above. If there is a use case for reuse, I think it's worth changing, but so far I feel not that much though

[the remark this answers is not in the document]
LePremierHomme: It's mostly about decoupling different application layers, not about optimizing or immediate code reuse. The wire protocol is usually defined on top of the underlying content messages.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
LePremierHomme: I would still keep my suggestion. The current design leads to multiple usage patterns (`self.request` vs direct `self.tx.send`), which isn't a good sign.

fendermint/vm/topdown/src/vote/store.rs
LePremierHomme: What about `CertifiedObservation.certified_at` & `CertifiedObservation.signature`? are they needed?

fendermint/vm/topdown/src/vote/tally.rs
cryptoAtwill: yep, I have updated already, it's implemented.

fendermint/vm/topdown/src/vote/store.rs
cryptoAtwill: we might need those to prevent some relay attacks for validation.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
cryptoAtwill: updated

fendermint/vm/topdown/src/observation.rs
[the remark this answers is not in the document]
LePremierHomme: Any comment here?

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
LePremierHomme: I don't see any diff pushed..

fendermint/vm/topdown/src/vote/store.rs
LePremierHomme: AFAIU these are to be used prior to storing the vote/observation. If otherwise, please elaborate.

fendermint/vm/topdown/src/vote/tally.rs
LePremierHomme: `check_quorum_at_height` is doing the sorting, and it's called on every write (`add_vote`). What I suggested is to maintain something like `BTreeMap<Weight, Observation>`, instead of recalculating it from `Vec<Observation>` each time.

LePremierHomme (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1139)* status: 0 of 25 files reviewed, 12 unresolved discussions (waiting on @cryptoAtwill)

___
*[`fendermint/vm/topdown/src/vote/tally.rs` line 29 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1139#-OIkQw_q2re6Z9DiJZXl:-OIkQw_q2re6Z9DiJZXm:b-fz74cg) ([raw file](https://github.com/consensus-shipyard/ipc/blob/68c6659b6332bedbbe14693459299233f6d03188/fendermint/vm/topdown/src/vote/tally.rs#L29)):*
> ```XML
> 
>     /// The latest observation with a quorum formed
>     latest_quorum_formed: Option<Observation>,
> ```

Does a proposer only cares about his local-highest block height quorum? 
___
*[`fendermint/vm/topdown/src/vote/tally.rs` line 156 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1139#-OIkMlIzAHhgKOEwOuoC:-OIkMlIzAHhgKOEwOuoD:bk1cqjy) ([raw file](https://github.com/consensus-shipyard/ipc/blob/68c6659b6332bedbbe14693459299233f6d03188/fendermint/vm/topdown/src/vote/tally.rs#L156)):*
> ```XML
>             // we only check quorum if the incoming vote height is bigger than latest quorum formed,
>             // no point choosing a lower parent height
>             .map(|v| v.parent_height < parent_height)
> ```

Seeing `.map()` here is a bit confusing.
What about something like
```rust
let check_quorum = match self.latest_quorum_formed.as_ref() {
    Some(v) => v.parent_height < parent_height, 
    None => true, 
};
```

fendermint/vm/topdown/src/vote/tally.rs
LePremierHomme: `check_quorum_at_height` is doing the sorting, and it's called on every write (`add_vote`). What I suggested is to maintain the weight per `Observation`, instead of recalculating it from `Vec<Observation>` each time.

cryptoAtwill:
@LePremierHomme Yep, only care about in coming votes with higher block number.

fendermint/vm/topdown/src/observation.rs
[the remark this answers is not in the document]
cryptoAtwill: so the question is why both the inner and envelop signatures are needed? As mentioned above, the double signatures are needed because they form the complete Observation. If you break them, but you still have to represent them somehow together. Might as well put them in one struct.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
cryptoAtwill: it is there: https://github.com/consensus-shipyard/ipc/pull/1139/files#diff-7fd156766435de4455e58d886f300f8a25e91ed701cea183ac6155b66b6e5eedR168

fendermint/vm/topdown/src/vote/store.rs
cryptoAtwill: `certified_at` is currently not validated, but introduced for recovery modes that's currently under discussion. If we include `certified_at`, then we need to include the signature for validation later in the process. Otherwise how do we know `certified_at` is not changed.

fendermint/vm/topdown/src/observation.rs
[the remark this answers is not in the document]
LePremierHomme: My latest question is about why not to decouple the different application layers, instead of bundling them in one flat struct.

fendermint/vm/topdown/src/vote/mod.rs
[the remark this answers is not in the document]
LePremierHomme: It's still using both `self.request` & `self.tx.send`. 
If you want to proceed with this design it's fine.

PR #1147, opened 2024-09-24
feat(node): address review comments against #1130

raulk:
See changelog.

### DIFF
--- a/contracts/contracts/lib/LibStaking.sol
+++ b/contracts/contracts/lib/LibStaking.sol
@@ -584,7 +584,7 @@ library LibStaking {
                 } else if (change.op == StakingOperation.Deposit)  {
                     s.validatorSet.confirmDeposit(validator, amount);
                     uint256 msgValue = s.collateralSource.makeAvailable(gateway, amount);
-                    IGateway(gateway).addStake{value: amount}(amount);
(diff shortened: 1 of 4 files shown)


contracts/contracts/lib/LibStaking.sol
```diff
@@ -583,15 +583,10 @@ library LibStaking {
...
                     s.validatorSet.confirmDeposit(validator, amount);
-
-                    if (s.collateralSource.isNative()) {
-                        IGateway(gateway).addStake{value: amount}(amount);
-                    } else {
-                        s.collateralSource.increaseAllowance(gateway, amount);
-                        IGateway(gateway).addStake(amount);
-                    }
+                    uint256 msgValue = s.collateralSource.makeAvailable(gateway, amount);
+                    IGateway(gateway).addStake{value: amount}(amount);
```
cryptoAtwill: I think this is not correct, it should be:
```
IGateway(gateway).addStake{value: msgValue}(amount);
```
Pushed a PR to this.

cryptoAtwill (approved):

contracts/contracts/lib/LibStaking.sol
raulk: Well spotted! Copy pasta

Issue #1080, opened 2024-07-18
L3+ unflagging


raulk:
Currently, the ability to spawn L3 subnets and beyond is behind a feature flag. We believe things are mostly working, but there are some loose ends.

## Criteria to unflag

Product:

- [x] Message propagation paths reviewed and well-tested, including response paths.
- [x] Make sure everything as battle-tested as it can be: unit tests, integration tests, scenario tests, etc.
- [x] Produce good demos of L3+ features.
- [x] Update documentation.

Loose ends:

- [x] Figure out how to deal with combinations of supply sources across the hierarchy.
- [ ] Figure out the relayer story -- can a single relayer support multiple subnets?
- [ ] Optimize the addressing scheme, as addresses get larger the deeper we go.

karlem:
### Action Plan for Testing and Improving L2+

1. **Integration Tests in Solidity (L3-Level Testing)**
   - Write and run integration tests to cover the following scenarios:
     - **Happy Path Scenarios**:
       - From root to L3.
       - From L3 to root.
       - Between sibling L3 nodes.
     - **Failure Scenarios**:
       - Errors at the destination.
       - Errors midway through execution.
     - **Edge Case Testing**:
       - Supply source mismatches.
       - Non-existent destinations.
     - Ensure all error cases propagate correctly.

2. **End-to-End Testing with Fendermint**
   - Create end-to-end tests that simulate message propagation:
     - From root to L3 and back.
     - From L3 to L3.
     - Verify that errors propagate properly across these paths.

3. **Advanced Testing (L4+)**
   - Implement **Parameterized Testing**: Use varied inputs to test subnet paths.
   - Apply **Generative Testing (Fuzzing)** for L3+ subnets to cover a wide range of scenarios based on randomized inputs.

**Continuous Improvement and Atomic PRs**
   - Optimize path routing.
   - Investigate if a relayer can be shared across multiple subnets, and explore relayer batching.
   - Focus on small, incremental code improvements (e.g., automatic postbox propagation).
   - Submit small, focused pull requests for each improvement to ensure the codebase remains clean and maintainable.

PR #1157, opened 2024-09-26
feat: support l2 plus

karlem:
Part of #1080 

This PR focuses on:

**Integration Tests in Solidity (L3-Level Testing)**
   - Write and run integration tests to cover the following scenarios:
     - **Happy Path Scenarios**:
       - From root to L3. - DONE
       - From L3 to root. - DONE
       - Between sibling L3 nodes. - DONE
     - **Failure Scenarios**:
       - Errors at the destination. - DONE
       - Errors midway through execution. Not sure this is really an issue.
     - **Edge Case Testing**:
       - Supply source mismatches. - This should really not be an issue - as the only cross net messages allowed are Call kind. (not trasfers)
       - Non-existent destinations. - DONE
     - Ensure all error cases propagate correctly. - DONE

**Improvements**

- Improved visibility by adding new events into the cross-message propagation process. A cross-message ID has been added to each event for better tracking and transparency:
  - `CheckpointCommitted`
  - `NewBottomUpMessage`
  - `MessageStoredInPostbox`
  - `MessagePropagatedFromPostbox`
- Automated cross-message propagation from the postbox. Previously, this process required an external caller to propagate the message, but now the IPC automatically propagates them after a checkpoint has been committed or after top-down messages have been applied.
- Added a check to verify that the destination address is indeed a contract and not an EOA (Externally Owned Account).
- Changed the behavior when applying messages: instead of reverting when a message is invalid, a result receipt is returned. This ensures that one invalid message does not disrupt the entire checkpoint submission process.
- Extracted message validation into a separate function to ensure validation is completed before committing the message. The validation function does *not* revert since reversion is only desirable when the message is directly triggered by a user. In scenarios where the message originates from a different subnet (e.g., in the form of a checkpoint), reversion is not appropriate. For the same reason, the commit functions (top-down, bottom-up) should also not revert, as they are triggered by automatic postbox propagation. Instead, validation is performed beforehand.

## TODOs

- [ ] Adding a doc on who, how is invoking cross messages calls. Because there were some assumption/requirements not properly documented yet, all the components needed to ensure correct functioning.
- [ ] Currently, there is missing a debug tool for cross message calls. It would be really helpful to have a tool, maybe external, that tracks the lifecycle of the cross network messages, i.e. where the message submitted, which subnet is the message now, is the message committed into the postbox and etc... This would help debug cross network message execution.
- [x] Checking subnet.circSupply. Not sure if this variable is tracked correctly. This value is [deducted](https://github.com/consensus-shipyard/ipc/blob/1585e1befe662d0671f99c945ba9cb019db554c3/contracts/contracts/gateway/router/CheckpointingFacet.sol#L151C16-L151C26) by IPCEnvelop.value regardless message type? But only [increased](https://github.com/consensus-shipyard/ipc/blob/1585e1befe662d0671f99c945ba9cb019db554c3/contracts/contracts/lib/LibGateway.sol#L254) (after bootstrap) when message is Transfer.  **DONE**: only deducted for [Transfer](https://github.com/consensus-shipyard/ipc/blob/1585e1befe662d0671f99c945ba9cb019db554c3/contracts/contracts/gateway/router/CheckpointingFacet.sol#L137)
- [ ] There should be a debug tool to check if the message execution is successful or failed. If failed, should be possible to show the error.
- [x] Consistent way to encode ret bytes. The way that encodes the error is using [encodeWithSelector](https://github.com/consensus-shipyard/ipc/blob/7572a6aac3b47a30e4d2146dee02e201acb865de/contracts/contracts/lib/LibGateway.sol#L362), but for L2+ error is just [encode](https://github.com/consensus-shipyard/ipc/blob/1585e1befe662d0671f99c945ba9cb019db554c3/contracts/contracts/lib/LibGateway.sol#L452), it's kind of confusing for debugging to tell which decode method to use. **DONE** - now maps to custom error. https://github.com/consensus-shipyard/ipc/pull/1157/commits/6b5901dafae390b6dadd02a527a2dcf9f9331c2c
- [x] [Emit event in test](https://github.com/consensus-shipyard/ipc/blob/1585e1befe662d0671f99c945ba9cb019db554c3/contracts/test/integration/L2PlusSubnet.t.sol#L494) should be removed. Events should be emitted in actual code, not from test, otherwise the expect becomes useless. @cryptoAtwill https://book.getfoundry.sh/cheatcodes/expect-emit it is not emitted from test. This is how the testing framework works you emit what you expect to see in the next call.
- [x] Fix duplicate nonce issue. **DONE** - https://github.com/consensus-shipyard/ipc/pull/1157/commits/34102cc92c6ac40b83aa5fd0485db4dea0107661
- [x] Do not transfer funds for call message **DONE** - https://github.com/consensus-shipyard/ipc/pull/1157/commits/6a6c53fd8c03539551892e279c3d083633cbd131


contracts/contracts/gateway/GatewayMessengerFacet.sol
```diff
@@ -13,12 +13,15 @@ import {AssetHelper} from "../lib/AssetHelper.sol";
...
 
+import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+
 string constant ERR_GENERAL_CROSS_MSG_DISABLED = "Support for general-purpose cross-net messages is disabled";
 string constant ERR_MULTILEVEL_CROSS_MSG_DISABLED = "Support for multi-level cross-net messages is disabled";
 
 contract GatewayMessengerFacet is GatewayActorModifiers {
     using FilAddress for address payable;
     using SubnetIDHelper for SubnetID;
+    using EnumerableSet for EnumerableSet.Bytes32Set;
```
cryptoAtwill, line 24: why not use `EnumerableMap` directly, feels like this is implementing `EnumerableMap`.

cryptoAtwill: oh, they dont support the value type.

contracts/contracts/gateway/GatewayMessengerFacet.sol
```diff
@@ -62,6 +66,18 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
             nonce: 0 // nonce will be updated by LibGateway.commitCrossMessage
         });
 
+        CrossMessageValidationOutcome outcome = LibGateway.validateCrossMessage(committed);
```
cryptoAtwill, line 69: why not just revert inside `validateCrossMessage`? Then there is no need to handle it outside.

raulk: @karlem does it makes sense to have a helper function here for syntactic sugar? I'd keep the logic in `LibGateway#validateCrossMessage`, but would add a helper in `CrossMsgHelper` [bound to the `CrossMsg` type](https://solang.readthedocs.io/en/latest/language/using.html#binding-methods-to-types-with-using) so we can do:

```solidity
CrossMessageValidationOutcome outcome = committed.validateCrossMessage();
```

contracts/contracts/gateway/GatewayMessengerFacet.sol
```diff
@@ -75,23 +91,9 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
...
-
-        bool shouldBurn = LibGateway.commitCrossMessage(crossMsg);
-        // We must delete the message first to prevent potential re-entrancies,
-        // and as the message is deleted and we don't have a reference to the object
-        // anymore, we need to pull the data from the message to trigger the side-effects.
-        uint256 v = crossMsg.value;
-        delete s.postbox[msgCid];
-
-        LibGateway.crossMsgSideEffects({v: v, shouldBurn: shouldBurn});
+    function propagateAllPostboxMessages() external payable {
```
raulk, line 96: Do we need to leak the "postbox" concept through the method name here? Why not just `propagateAll`?

contracts/contracts/gateway/router/CheckpointingFacet.sol
```diff
@@ -41,6 +44,8 @@ contract CheckpointingFacet is GatewayActorModifiers {
         LibGateway.checkMsgLength(checkpoint.msgs);
 
         execBottomUpMsgs(checkpoint.msgs, subnet);
+
+        emit CheckpointCommitted({subnet: checkpoint.subnetID.getAddress(), blockHeight: checkpoint.blockHeight});
```
raulk, line 48: Let's remove ambiguity here:

```suggestion
        emit CheckpointCommitted({subnet: checkpoint.subnetID.getAddress(), subnetHeight: checkpoint.blockHeight});
```

contracts/contracts/lib/AssetHelper.sol
```diff
@@ -196,6 +209,15 @@ library AssetHelper {
         }
     }
 
+    // @notice Gets the balance of the account.
+    function balanceOf(Asset memory asset, address account) internal view returns (uint256 ret) {
```
raulk, line 213: Not necessarily an account -- could be a contract:

```suggestion
    function balanceOf(Asset memory asset, address holder) internal view returns (uint256 ret) {
```

raulk (commented):
Some minor comments.

contracts/contracts/gateway/GatewayMessengerFacet.sol
karlem: I think it makes sense to call it propagateAll here. But on the gateway interface it is then a bit vague. That's why I made it so specific. But I can revert it back.

contracts/contracts/gateway/GatewayMessengerFacet.sol
karlem: @raulk yeah good idea.

contracts/contracts/gateway/GatewayMessengerFacet.sol
karlem: Maybe `propagateAllMessages`?

contracts/contracts/gateway/router/XnetMessagingFacet.sol
```diff
@@ -27,6 +27,7 @@ contract XnetMessagingFacet is GatewayActorModifiers {
     /// @dev It requires the caller to be the system actor.
     /// @param crossMsgs The array of cross-network messages to be applied.
     function applyCrossMessages(IpcEnvelope[] calldata crossMsgs) external systemActorOnly {
-        LibGateway.applyMessages(s.networkName.getParentSubnet(), crossMsgs);
+        LibGateway.applyTopDownMessages(s.networkName.getParentSubnet(), crossMsgs);
```
cryptoAtwill, line 30: `applyCrossMessages` only handles topdown messages?

contracts/contracts/lib/CrossMsgHelper.sol
```diff
@@ -199,4 +203,8 @@ library CrossMsgHelper {
 
         return true;
     }
+
+    function validateCrossMessage(IpcEnvelope memory crossMsg) internal view returns (CrossMessageValidationOutcome)  {
```
cryptoAtwill, line 207: caller why not just invoke `LibGateway` directly instead, it's internal anyways.

contracts/contracts/lib/SubnetIDHelper.sol
```diff
@@ -130,13 +128,13 @@ library SubnetIDHelper {
     /// down in the path from the subnet id given as argument.
     /// subnet2 needs to be a prefix of the subnet1.
     /// If subnet1 is /a/b/c/d and subnet2 is /a/b, then the returned ID should be /a/b/c.
-    /// @dev Revert will be triggered if subnet2 is an invalid input.
+    /// @dev Returns an empty SubnetID if subnet2 is not a prefix of subnet1 or if the roots are different.
     function down(SubnetID calldata subnet1, SubnetID calldata subnet2) public pure returns (SubnetID memory) {
         if (subnet1.root != subnet2.root) {
-            revert DifferentRootNetwork();
+            return SubnetID({root: 0, route: new address[](0)});
```
cryptoAtwill, line 134: is there a reason why it's not reverted? Feels like it's an error that should be handled instead of introducing `root = 0` as invalid subnet. Probably in some use cases `root: 0, route: []` is a valid use case.

contracts/contracts/lib/LibGateway.sol
```diff
@@ -228,23 +245,14 @@ library LibGateway {
...
-        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
-        SubnetID memory subnetId = crossMessage.to.subnetId.down(s.networkName);
-
-        (bool registered, Subnet storage subnet) = getSubnet(subnetId);
-
-        if (!registered) {
-            revert NotRegisteredSubnet();
-        }
-
+    function commitTopDownMsg(Subnet storage subnet, IpcEnvelope memory crossMessage) internal {
```
cryptoAtwill, line 248: feels like we just need the subnet id as the param

contracts/contracts/lib/CrossMsgHelper.sol
```diff
@@ -199,4 +203,8 @@ library CrossMsgHelper {
...
+        return LibGateway.validateCrossMessage(crossMsg);
```
cryptoAtwill, line 208: I still kind of feel `CrossMessageValidationOutcome` should not be exposed, unless different callers might handle the outcomes differently? Now every call of `validateCrossMessage` must handle the outcome, dont feel it's necessary though.

contracts/contracts/lib/LibGateway.sol
```diff
@@ -355,6 +380,19 @@ library LibGateway {
     /// @param arrivingFrom - the immediate subnet from which this message is arriving
     /// @param crossMsg - the cross message to be executed
     function applyMsg(SubnetID memory arrivingFrom, IpcEnvelope memory crossMsg) internal {
```
cryptoAtwill, line 382: this method seems confusing though, because it's called `applyMsg` but actually it's expecting bottom up only.

contracts/contracts/lib/LibGateway.sol
```diff
@@ -404,11 +444,20 @@ library LibGateway {
         // should increase the appliedNonce to allow the execution of the next message
         // of the batch (this is way we have this after the nonce logic).
         if (!crossMsg.to.subnetId.equals(s.networkName)) {
+            CrossMessageValidationOutcome outcome = validateCrossMessage(crossMsg);
```
cryptoAtwill, line 447: Feels like this could help: https://solidity-by-example.org/try-catch/

contracts/contracts/lib/LibGateway.sol
```diff
@@ -547,4 +559,86 @@ library LibGateway {
...
+
+     /**
+     * @dev Propagates the populated cross-net message for the given `msgCid`.
+     * @param msgCid - the cid of the cross-net message
+     */
+    function propagatePostboxMessage(bytes32 msgCid) internal {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        IpcEnvelope storage crossMsg = s.postbox[msgCid];
+
+        if (crossMsg.isEmpty()) {
```
cryptoAtwill, line 625: will this ever happen？This should be checked before commitment?

contracts/test/integration/L2PlusSubnet.t.sol
```diff
@@ -0,0 +1,857 @@
...
+        GatewayDiamond rootGateway = createGatewayDiamond(gatewayParams(rootNetworkName));
+
+        rootNetwork = RootSubnetDefinition({
+            gateway: rootGateway,
+            gatewayAddr: address(rootGateway),
+            id: rootNetworkName
+        });
+
+        address[] memory nativeSubnetPath;
+        nativeSubnet = createNativeSubnet(nativeSubnetPath, rootNetwork.gatewayAddr, rootNetwork.id);
```
cryptoAtwill, line 69: maybe call this `nativeL2Subnet`?

contracts/test/integration/L2PlusSubnet.t.sol
```diff
@@ -0,0 +1,857 @@
...
+        vm.prank(params.callerAddr);
+
+        fundSubnet(params.root.gateway, params.subnet, params.callerAddr, 100000);
+        fundSubnet(params.subnet.gateway, params.subnetL3, params.callerAddr, 100000);
+
+        // create the xnet message on the subnet L3 - it's local gateway
+        IpcEnvelope memory crossMessage = TestUtils.newXnetCallMsg(
+            IPCAddress({subnetId: params.subnetL3.id, rawAddress: FvmAddressHelper.from(params.callerAddr)}),
+            IPCAddress({subnetId: params.root.id, rawAddress: FvmAddressHelper.from(params.recipientAddr)}),
+            params.amount,
```
cryptoAtwill, line 397: the param amount could exceed the `fundSubnet` amount?

contracts/test/integration/L2PlusSubnet.t.sol
```diff
@@ -0,0 +1,857 @@
...
+        );
+
+        // apply the cross message in the L3 subnet
+        executeTopDownMsgs(msgs, params.subnetL3.gateway);
+        assertEq(params.recipientAddr.balance, params.expectedAmount);
+
+        // submit checkoint so the result message can be propagated to L2
+        submitBottomUpCheckpoint(
+            callCreateBottomUpCheckpointFromChildSubnet(params.subnetL3.id, params.subnetL3.gateway),
+            params.subnetL3.subnetActor
```
cryptoAtwill, line 506: this should be `subnetL2.subnetActor`? The checkpoint should be propagated to L2 right? Not to L3 subnet actor.

contracts/test/integration/L2PlusSubnet.t.sol
```diff
@@ -0,0 +1,857 @@
...
+            0
+        );
+
+        vm.prank(params.callerAddr);
+        params.subnetL3.gateway.messenger().sendContractXnetMessage{value: params.amount}(crossMessage);
+
+        // this would normally submitted by releayer. It call the subnet actor on the L2 network.
+        submitBottomUpCheckpoint(
+            callCreateBottomUpCheckpointFromChildSubnet(params.subnetL3.id, params.subnetL3.gateway),
+            params.subnetL3.subnetActor
```
cryptoAtwill, line 407: the `params.subnetL3.subnetActor` refers to the `SubnetActorDiamond` in L2?

contracts/test/integration/L2PlusSubnet.t.sol
```diff
@@ -0,0 +1,857 @@
...
+        );
+        tokenL3SubnetsWithTokenParent.push(
+            createTokenSubnet(address(tokenL3), tokenSubnet.path, tokenSubnet.gatewayAddr, tokenSubnet.id)
+        );
+
+        printActors();
+    }
+
+    function createNativeSubnet(
+        address[] memory subnetPath,
```
cryptoAtwill, line 98: shouldnt this be parent subnet id? How do you know the subnet path before it's created?

cryptoAtwill (changes requested):

cryptoAtwill:
@karlem Another question I have is regarding nonce. Say a bottom up message is created in L3, the nonce will be set by the local network. Then in L2, will the cross network message nonce change? If so, then the id (`keccak`) will change too?

raulk (commented):
Some initial comments.

contracts/contracts/gateway/router/XnetMessagingFacet.sol
karlem: Yes, it is only used to apply top-down messages locally after they have been finalized in Fendermint.

contracts/contracts/lib/CrossMsgHelper.sol
karlem: Raul has suggested a nice syntax sugar update. So I have added it after his comment.

contracts/contracts/lib/SubnetIDHelper.sol
karlem: The problem here is that if I revert, I lose control elsewhere. The method is also used in`validateCrossMessage`, where reverting is not desirable. Additionally, `commonParent` method here returns the empty subnet ID instead of reverting. But I understand your point—if this were deployed on Ethereum, the root could be zero. Would returning a boolean (indicating whether it was found or not) help?

contracts/contracts/lib/LibGateway.sol
karlem: Yeah, I can pass the subnetID down, but then I would need to look up the subnet again with getSubnet(subnetId), since this function actually updates the subnet state (and has been).

contracts/contracts/lib/LibGateway.sol
karlem: It does not expect only bottom-up messages; it expects both. But I understand it can be confusing. :)

contracts/contracts/lib/LibGateway.sol
```diff
@@ -457,65 +506,27 @@ library LibGateway {
...
-            }
-        }
+        bool isLCA = to.commonParent(crossMessage.from.subnetId).equals(s.networkName);
 
         // If the directionality is top-down, or if we're inverting the direction
         // because we're the LCA, commit a top-down message.
         if (applyType == IPCMsgType.TopDown || isLCA) {
             ++s.appliedTopDownNonce;
-            LibGateway.commitTopDownMsg(crossMessage);
+            (, Subnet storage subnet) = getSubnet(to.down(s.networkName));
```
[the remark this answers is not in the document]
karlem: Yes, at this point this should be validated by the caller by using the `validateCrossMessage`.

contracts/contracts/lib/LibGateway.sol
karlem: I think the only reason I kept this here is because Solidity doesn’t have any other way to check if a record in the mapping exists, right? Correct me if I’m wrong. Although, I guess we can trust that the keys we iterate through in postboxKeys are correctly in sync. So, it should be fine to remove it?

contracts/test/integration/L2GatewayDiamond.t.sol
```diff
@@ -45,11 +47,11 @@ contract L2GatewayActorDiamondTest is Test, L2GatewayActorDiamond {
         IpcEnvelope[] memory topDownMsgs = new IpcEnvelope[](n);
         for (uint64 i = 0; i < n; i++) {
             topDownMsgs[i] = TestUtils.newXnetCallMsg(
-                IPCAddress({subnetId: id, rawAddress: FvmAddressHelper.from(address(this))}),
                 IPCAddress({
                     subnetId: gatewayDiamond.getter().getNetworkName().getParentSubnet(),
                     rawAddress: FvmAddressHelper.from(receipient)
                 }),
+                IPCAddress({subnetId: id, rawAddress: FvmAddressHelper.from(address(this))}),
```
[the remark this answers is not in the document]
karlem: Because it was labeled as a top-down message, but the path (from, to) made it a bottom-up message, the test was incorrect. Since we now have checks ensuring that top-down messages are indeed top-down (as a parent can't send a message to itself), this would cause a revert.

contracts/test/integration/L2PlusSubnet.t.sol
karlem: @raulk Solved by using the parent route instead.

contracts/test/integration/L2PlusSubnet.t.sol
karlem: Yes.

contracts/test/integration/L2PlusSubnet.t.sol
karlem: params.subnetL3.subnetActor corresponds to the L2 subnet actor. I agree it can be confusing, but I'm building around the already existing `TestSubnetDefinition`.

karlem: I guess if you find it super confusing I can rename it. But it might affect a lot of files :)

contracts/test/integration/L2PlusSubnet.t.sol
karlem: Good catch!

contracts/contracts/lib/CrossMsgHelper.sol
karlem: `unless different callers might handle the outcomes differently? ` - that's exactly the case :)

karlem:
> @karlem Another question I have is regarding nonce. Say a bottom up message is created in L3, the nonce will be set by the local network. Then in L2, will the cross network message nonce change? If so, then the id (`keccak`) will change too?

Hmm that's a good point. I will think about how to mitigate this.

karlem:
@cryptoAtwill should be ready for another review :)

contracts/contracts/lib/CrossMsgHelper.sol
```diff
@@ -131,6 +135,22 @@ library CrossMsgHelper {
         return keccak256(abi.encode(crossMsg));
     }
 
+    /// @notice Returns a deterministic hash for the given cross message. The hash remains the same accross different networks
+    /// because it doesn't include the network-specific nonce.
+    function toDeterministicHash(IpcEnvelope memory crossMsg) internal pure returns (bytes32) {
```
cryptoAtwill, line 140: I guess most of the time this is ok, but will this cause conflict if say someone sends the same message twice?

contracts/contracts/lib/LibGateway.sol
```diff
@@ -355,6 +380,19 @@ library LibGateway {
...
+    
+    /// @notice executes a cross message if its destination is the current network, otherwise adds it to the postbox to be propagated further
+    /// This function assumes that the relevant funds have been already minted or burnt
+    /// when the top-down or bottom-up messages have been queued for execution.
+    /// This function is not expected to revert. If a controlled failure happens, a new
+    /// cross-message receipt is propagated for execution to inform the sending contract.
+    /// `Call` cross-messages also trigger receipts if they are successful.
+    /// @param arrivingFrom - the immediate subnet from which this message is arriving
+    /// @param crossMsg - the cross message to be executed
+    /// @param expectTopDownOnly - whether the message should be top-down only. Reverts if it is not.
```
cryptoAtwill, line 394: I think I was misled by the comment here. The idea is `whether the message should be top-down only. Reverts if this flag is true but the crossmsg is actually bottom up`?

cryptoAtwill (approved):
LGTM. But I do have a few follow up questions though:
1. Does it work in calibration with a live 3 level network? Is there a testnet we can try or planned?
2. `propagateAll` is called in the parent as part of bottom up checkpoint. Bottom up checkpoint is already quite gas costly, not sure if this will burst the gas even more.

contracts/contracts/lib/CrossMsgHelper.sol
karlem: It might collide, but it is only for logging so it should not matter.

contracts/contracts/lib/LibGateway.sol
karlem: Reverts if it's both - not only top down :)

cryptoAtwill (changes requested):
I think this is no longer correct: https://github.com/consensus-shipyard/ipc/pull/1157/files#diff-71559bfe927afaf1a3980c01ffe9f4d15c77d58650b2d850bbae821de9c009d0R49.

This implicitly assumes the token behind `envelop.value` is the native token in the origin subnet. If say I want to trigger transfer from address A to address B in L1 and I'm sending from L3, why should I attach that amount in `msg.value`?

It comes back to the definition of `envelop.value`, is it just the native token transfer in the target subnet. If this is the case, it does not make sense to have this check.

karlem:
@cryptoAtwill The issue was resolved by not using or supporting native token transfers in the Call message.

contracts/contracts/lib/LibGateway.sol
```diff
@@ -405,11 +447,24 @@ library LibGateway {
...
+                sendReceipt(
+                    crossMsg,
+                    OutcomeType.SystemErr,
+                    abi.encodeWithSelector(InvalidXnetMessage.selector, validationOutcomeToInvalidXnetMsgReason(outcome))
+                );
+                return;
+            }
+
             bytes32 cid = crossMsg.toHash();
+            s.postboxKeys.add(cid);
```
cryptoAtwill, line 461: if the cross network message is topdown, there is no point inserting into postbox? Why not just emit a topdown event or call commit topdown directly?

contracts/test/integration/GatewayDiamondToken.t.sol
```diff
@@ -233,7 +232,7 @@ contract GatewayDiamondTokenTest is Test, IntegrationTestBase {
         vm.prank(address(saDiamond));
         vm.expectCall(recipient, abi.encodeCall(IIpcHandler.handleIpcMessage, (msgs[0])), 1);
         gatewayDiamond.checkpointer().commitCheckpoint(batch);
-        assertEq(token.balanceOf(recipient), 8);
+        assertEq(token.balanceOf(recipient), 0);
```
cryptoAtwill, line 235: why change this to 0?

karlem: Because I have disable the transfer for Call kind.

cryptoAtwill:
This looks good to me. Just one more thing regarding the tests, shall we also add nonce check to bottom up, just make sure it's correct?

cryptoAtwill (approved):

raulk (changes requested):
Some initial feedback. Removing the ability to send value in calls is a no-go. See inline comments for more context. It will be very common for subnets denominated in the same asset to want to interact with one another sending value and calling some logic in an atomic manner.

@karlem can you please detail the rationale for having removed this? Was it due to the risk of supply source mismatches across the hierarchy during a multi-hop? We had logic to prevent this here: https://github.com/consensus-shipyard/ipc/blob/4243ff9378e66850f7df4d41ec98e44d388d86c4/contracts/contracts/lib/LibGateway.sol#L499-L513

So curious to hear why that was insufficient, or why strengthening the validation to detect supply source mismatches was a no-go?

contracts/contracts/lib/LibGateway.sol
```diff
@@ -548,4 +566,105 @@ library LibGateway {
...
+
+        bool shouldBurn = LibGateway.commitValidatedCrossMessage(crossMsg);
+
+        // Cache value before deletion to avoid re-entrancy
+        uint256 v = crossMsg.value;
+        bytes32 deterministicId = crossMsg.toDeterministicHash();
+
+        // Remove the message to prevent re-entrancy and clean up state
+        delete s.postbox[msgCid];
+        s.postboxKeys.remove(msgCid);
```
cryptoAtwill, line 662: this is buggy, if `msgCid` is removed from postboxKeys, the length of the set changes and this line: https://github.com/consensus-shipyard/ipc/pull/1157/files#diff-b26838f2cc5f8bf764313ca069618747d9171766dc768ece47de2d9c38ffddfbR632 will panic due to array out of bound.

contracts/contracts/structs/CrossNet.sol
```diff
@@ -74,15 +74,26 @@ struct IpcEnvelope {
...
+    /// Similarly, in a `Result` message, `value` might represent the outcome of a transaction, such as the total tokens minted or refunded.
+    ///
+    /// For `Transfer` messages, `value` represents the actual amount of native tokens being transferred across networks.
     uint256 value;
     /// @dev abi.encoded message
     bytes message;
+    /// @dev original nonce of the message from the source network.
+    /// It is set once at the source network and remains unchanged during propagation.
+    /// It is used to generate a unique tracing ID across networks, which is useful for debugging and auditing purposes.
+    uint64 originalNonce;
```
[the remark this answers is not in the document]
karlem: Because we wanted to safe space.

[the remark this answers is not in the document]
raulk: @karlem Will has a reasonable question. We made this decision after a private discussion, so it would be helpful if you explained the rationale here for Will (and for posterity).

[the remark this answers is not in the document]
karlem: The decision has been made that, since this is not a core part of the system and is "only" used for debugging, it may be worth saving the 24 bytes on each network the message flows through. This is especially relevant given that storage is more costly than the relatively inexpensive `abi.encode` operation.

Issue #1079, opened 2024-07-18
[Collateral-based membership] Validator rewards


raulk:
The flip side of #1051 is validator rewards. To keep validators rational and aligned with correct behaviour, IPC needs to deliver cryptoeconomic incentives for subnet validation.

## Framework approach

IPC as a framework should be fairly unopinionated by how and which rewards are distributed, as this will depend on concrete subnet tokenomics, especially in the presence of custom ERC20 subnet tokens. Our framework should allow subnet operators to configure, parameterize, and customize the reward mechanics. This implies design choices that favour modularity via hooks and user-pluggable functions and policies. We should provide one or several basic policy/strategies out of the box as quickstarts / examples.

## Considerations

- Coin/token choice. Subnet operators should choose which coin/token rewards get paid in.
- Disbursement site. Can happen at the parent when checkpoints are presented (in which case they must carry a digest of proposers), or at the child (where more information is available and disbursement can happen immediately; handicap: we have no link between a validator and an account yet, but that needs to be fixed for #925 anyway).
- Vault. In both cases, since the subnet does not control the supply, there must be the concept of a rewards vault/treasury, and _possibly_ the concept of convertible debt to handle continuity in case of temporary vault depletion.
- Vault transfers. Because subnets can be created dynamically, amounts in vault must be transferred down at subnet genesis, and any remnants back up at subnet EOL.
- Customizable reward function. Depending on the disbursement site, this may have access to more or less data. We should keep this agnostic of the concrete consensus algorithm (e.g. do not model rounds, since some consensus protocols don't have rounds).
- Vesting curve. Some protocols may want to keep some portion of rewards locked-up for some time, which may be forfeited or not if the validator leaves the network early. TBD.

raulk:
@sanderpick @bigs @carsonfarmer @angelo-schalley-p @hannahhoward @boneyard93501 would appreciate your input here!

boneyard93501:
Looks pretty good to me. Thanks for writing this up.

Couple things, bottom up:
* Vesting considerations are pretty critical and an operator should be able to to specify two schedules: one for stake, one for reward
[...]

PR #1162, opened 2024-10-02
Rrototype the validator rewards feature.

raulk:
This PR prototypes the Validator Rewards feature (#1079). I tried to keep it as simple as possible for now. I've authored the data structures and inserted `TODO(rewards)` comments everywhere I think we need to extend the implementation, but there may be other spots that I haven't covered.

## IValidatorRewarder

This is the interface that should be implemented to disburse validator rewards for a subnet on its parent.

## Summaries and commitments

The ActivitySummary is the central object that is created by the subnet and committed in the checkpoint in a hash. The full payload is emitted as an event on the local chain so that the relayer can pick it up and submit it to the root network once the commitment has fully travelled upwards.

## Subnet Actor storage

There are new fields added to specify the address of a ValidatorRewarder, and to track commitments to summaries that are pending forwarding and presentation.

## Relayer

The relayer should be augmented to watch multiple subnets at once. It will also track summaries that are emitted locally and will present them in the root chain once the relevant commitment has travelled all the way up.

## Fendermint

In `maybe_create_checkpoint` we should craft the ValidatorSummary object by querying the chain in CometBFT. That said, this is not the ideal solution as it likely adds latency. Instead, we should keep a cache (queue) that's updated when new blocks are committed, and pruned when new checkpoints are generated. However, that might be more complex, so maybe we start simple and only implement the cache if we observe performance issues?


cryptoAtwill:
@raulk Regarding `fendermint` construct `ActivitySummary`, I think querying the chain might be too slow, imaging 600 bottom up checkpoint period. Instead, maybe we can create an actor to track this, `end_block` hook, solidity contract can do too.

cryptoAtwill:
@raulk So in summary, the high level flow diagram is:
- Fendermint mines a new block
- The validator that proposed this is recorded
- When a bottom up checkpoint is created, create `ActivitySummary` as well and attach it to bottom up checkpoint. 
- Quorum formation in fendermint, emit Activity event
- Relayer push the rewards distribution with `ActivitySummary`

A few questions I have:
1. Actually why not ask validators to claim? They can always query fendermint to obtain ActivitySummary and claim their rewards in batch. So relayer does not really have to worry about gas. Compared to the validator rewards, the gas imposed on validators should be quite minimal.
2. RelayedActivity seems not clear how it's used. Currently validators are coming from parent, so as for the use case now, it really does not seems to be used. Maybe I'm missing sth.

contracts/contracts/interfaces/IValidatorRewarder.sol
```diff
@@ -0,0 +1,18 @@
...
+///
+/// @dev Implement this interface and supply the address of the implementation contract at subnet creation to process
+/// subnet summaries at this level, and disburse rewards to validators based on their block production activity.
+///
+/// This interface will be called by the subnet actor when a relayer presents a
+interface IValidatorRewarder {
+    /// @notice Called by the subnet manager contract to instruct the rewarder to process the subnet summary and
+    /// disburse any relevant rewards.
+    /// The
+    /// @dev This method should revert if the summary is invalid; this will cause the
```
JulissaDantes, line 16: `this will cause the` what will it cause? And what criteria makes a summary invalid?

contracts/contracts/structs/CrossNet.sol
```diff
@@ -29,6 +29,31 @@ struct BottomUpCheckpoint {
...
+    /// @dev Summaries relayed upwards from descendants of this subnet.
+    /// NOTE: Not merkelized to keep it simple, but we will merkelize later to scale better.
+    RelayedSummary[] relayedSummaries;
+}
+
+struct ActivitySummary {
+    /// @dev The block range the activity summary spans; these are the local heights of the start and the end, inclusive.
+    uint256[2] blockRange;
+    /// @dev The validators whose activity we're reporting about.
+    address[] validators;
```
JulissaDantes, line 43: Just curious why 2 arrays instead of a mapping?

JulissaDantes: I understand the arrays if when distributing the rewards we have to process them all but I agree with @cryptoAtwill, validators could claim their own rewards, so a mapping would make sense.

contracts/contracts/structs/CrossNet.sol
```diff
@@ -29,6 +29,31 @@ struct BottomUpCheckpoint {
...
+    /// @dev The number of blocks committed by each validator in the position they appear in the validators array.
+    /// If there is a configuration change applied at this checkpoint, this carries information about the _old_ validator set.
+    uint64[] blocksCommitted;
```
JulissaDantes, line 46: Could we have a metadata bytes typed member here to allow to pass additional info?

carsonfarmer: Nice suggestion @JulissaDantes. @raulk, in particular, we have additional information we'd like to propagate up the hierarchy. I know we had talked about specific "zones" for this type of information in the checkpoints at one point. What is your thinking here at this stage?

contracts/contracts/structs/CrossNet.sol
cryptoAtwill: This is a DTO, so mapping cannot, but I have updated the data struct slightly for storage

contracts/contracts/structs/CrossNet.sol
cryptoAtwill: At least for solidity, there is no generic, then perhaps only bytes array? To let the subnet/fendermint and parent IValidatorRewarder handle the bytes deserialization + distribution. What the framework does is just checking:
- Validator has claimed before?
- Does the claim actually exists?
- The lifecycle of the validator reward

raulk: Yes, this is the direction I wanted to move things in. This was the idea:
1. We add a region in checkpoints to carry extensible subnet activity reports in the form of `{ namespace/type/discriminator: bytes, payload: bytes }[]` -- we can call this checkpoint extensions.
2. IPC ships with a set of default subnet activity reporters as extensions, e.g. the block production one here.
3. Users can plug in custom subnet activity reporters to, e.g. report on quality of service, retrieval rate, scoring, etc.
4. Ideally the reporters are Wasm actors, so they can compose easily and be upgraded through governance in the future. However, to truly make this design client/node-agnostic, users would likely need to add custom syscalls to access the metadata actors will want to report on.
5. We'd need a master actor to orchestrate "hooks". The `extend_checkpoint` hook would be called by the IPC node, injecting the raw Checkpoint and expecting an array of extension fragments in return. The client would assemble them into the Checkpoint and submit that on chain for local quorum signature.

@carsonfarmer @JulissaDantes In my head we could defer this feature, but it sounds like it would be beneficial to you guys to have it now, right?

If so, I think we can simplify the above on several fronts to try to meet the time constraints while delivering the feature. How about:

1. We don't bother with developing the checkpoint extensions as Wasm-land actors for now; we just add well-bounded hook callbacks in Fendermint with the intention to migrate them to actors in the future.
2. We have the node orchestrate the hooks instead of a master actor.

---

Open points:

1. Namespacing.
2. Parent execution.

raulk:

PR #1163, opened 2024-10-02
fix: tracing deserialization

sanderpick:
- Allow passing `domain_filter` and `events_filter` via env
- Allow passing lowercase log level, e.g., "info", instead of "Info"

### DIFF
--- a/docs/fendermint/observability.md
+++ b/docs/fendermint/observability.md
@@ -357,7 +357,7 @@ directory = "/path/to/log/directory"
 max_log_files = 5 # Number of files to keep after rotation
 rotation = "daily" # Options: minutely, hourly, daily, never
 ## Optional: filter events by domain
-domain_filter = ["Bottomup", "Consenesus", "Mpool", "Execution", "Topdown", "System"]
+domain_filter = ["Bottomup", "Consensus", "Mpool", "Execution", "Topdown", "System"]
 ## Optional: filter events by event name
 events_filter = ["ParentFinalityAcquired", "ParentRpcCalled"]
 ```
--- a/docs/fendermint/observability.md
+++ b/docs/fendermint/observability.md
@@ -357,9 +357,9 @@ directory = "/path/to/log/directory"
 max_log_files = 5 # Number of files to keep after rotation
 rotation = "daily" # Options: minutely, hourly, daily, never
 ## Optional: filter events by domain
-domain_filter = "Bottomup, Consenesus, Mpool, Execution, Topdown, TracingError"
+domain_filter = ["Bottomup", "Consenesus", "Mpool", "Execution", "Topdown", "System"]
 ## Optional: filter events by event name
-events_filter = "ParentFinalityAcquired, ParentRpcCalled"
+events_filter = ["ParentFinalityAcquired", "ParentRpcCalled"]
 ```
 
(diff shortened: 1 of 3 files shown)


ipc/observability/src/config.rs
```diff
@@ -12,6 +12,7 @@ use tracing_subscriber::filter::EnvFilter;
 #[serde_as]
 #[derive(Debug, Deserialize, Clone, Default, strum::EnumString, strum::Display)]
 #[strum(serialize_all = "snake_case")]
+#[serde(rename_all = "lowercase")]
```
sanderpick, line 15: apparently `#[strum(serialize_all = "snake_case")]` only applies to and from string, not with serde.

raulk (approved):

fendermint/app/settings/src/lib.rs
```diff
@@ -326,12 +326,16 @@ impl Settings {
...
+                    .with_list_parse_key("tracing.file.events_filter")
                     .with_list_parse_key("resolver.connection.external_addresses")
                     .with_list_parse_key("resolver.discovery.static_addresses")
                     .with_list_parse_key("resolver.membership.static_subnets")
                     .with_list_parse_key("eth.cors.allowed_origins")
                     .with_list_parse_key("eth.cors.allowed_methods")
-                    .with_list_parse_key("eth.cors.allowed_headers"),
+                    .with_list_parse_key("eth.cors.allowed_headers")
+                    .with_list_parse_key("eth.tracing.file.domain_filter")
+                    .with_list_parse_key("eth.tracing.file.events_filter"),
```
raulk, line 338: Are these necessary for the Eth API process? I think it should just reuse the general tracing configuration, and if we want to enable/disable Eth API tracing, we should have an Eth domain we can disable through the `domain_filter`.

raulk (approved):

docs/fendermint/observability.md
```diff
@@ -357,9 +357,9 @@ directory = "/path/to/log/directory"
 max_log_files = 5 # Number of files to keep after rotation
 rotation = "daily" # Options: minutely, hourly, daily, never
 ## Optional: filter events by domain
-domain_filter = "Bottomup, Consenesus, Mpool, Execution, Topdown, TracingError"
```
sanderpick, line 360: looking for instances of `impl_traceables!`, I didn't find one with `"TracingError"`, but there was one with `"System"`

docs/fendermint/observability.md
```diff
@@ -357,9 +357,9 @@ directory = "/path/to/log/directory"
...
+domain_filter = ["Bottomup", "Consenesus", "Mpool", "Execution", "Topdown", "System"]
```
raulk, line 360:
```suggestion
domain_filter = ["Bottomup", "Consensus", "Mpool", "Execution", "Topdown", "System"]
```

raulk (approved):

fendermint/app/settings/src/lib.rs
sanderpick: Nice, that sounds better. So maybe `EthSettings` should have a subset of the tracing settings? as in:
```
pub domain_filter: Option<Vec<String>>,
pub events_filter: Option<Vec<String>>,
```

sanderpick: I'm happy to make that change if sounds right... would prefer to do in a diff PR.

PR #1165, opened 2024-10-07
feat(node): validator reward skeleton implementation

cryptoAtwill:
This PR extends #1162 with the following:

- Implemented the validator reward distribution flow in the parent subnet. Instead of adding directly to the subnet storage, this PR uses library and facet to make reward distribution more modular. 
- Added a metadata field to validator summary for extra data required. 
- Added `ValidatorActivityTracker` trait `fendermint` so that the implementation can decide how to track validator's activity.
- Implemented a solidity contract based `ValidatorActivityTracker`, only contract is implemented at the moment.

Todos:
- [x] Gather initial feedbacks on this approach.
- [x] Adding integration for `ValidatorActivityTracker` and solidity contracts.
- [ ] Tests


raulk:

Issue #1150, opened 2024-09-26
Genesis validator set is not updated completely


cryptoAtwill:
### Description
When the child gateway is created, the membership is forwarded to the child, but however the validator tracker is not aware of the changes.

The validator tracker is a module in the child gateway to apply the validator changes coming from the parent. This is achieved by storing validator changes from the top down checkpoint, then confirmed in bottom up checkpoint creation, see https://github.com/consensus-shipyard/ipc/blob/1fb733355c68379b80d65013c33fc292a681a939/contracts/contracts/lib/LibStaking.sol#L641. 

What happens is when someone bootstraps a subnet with some federated power, the initial validator's federated power is not synced to the child gateway validator tracker. So once the next batch of validator changes arrive, the tracker's validator set does not have the initial validators' information and hence their powers are lost.

### Proposed fixes
Adding `permissionMode` to the gateway constructor. Based on the genesis validator, we can initialized the validator tracker. Currently we cannot do that is because validator tracker requires specifying `collateral` and `federatedPower` and genesis validators only have `weight` which is `collateral + federatedPower`. Without `permissionMode`, one cannot tell whether the power should go to collateral or federated.

As a work around without updating the constructor, one can:
- Initial power update set full validator sets' power. This is only feasible for `federatedPower`.

Considering we will have a major contract refactor such that the validators no longer need the split between `federatedPower` and `collateral`, but just power, it would be really simple to fix this.

On the other hand, perhaps it would be useful to let the child gateway be aware of where the power comes from. Maybe, say for example, an L3 is collateral based subnet, but L2 is federated, then a validator penalty propagation will stop at L2.

gvelez17:
I don't have a strong opinion here, I think it might be ok to prohibit some kinds of inheritance for now.  The main thing is to have it simple, robust and clearly documented.  If we say only federated can inherit from federated that's fine, it would simplify the test paths.  

It is helpful to not have to know the full set of validators when adding one

but its ok if we do have to, as long as that requirement is clear, and maybe a warning if you are adding one only, so that you will know you are replacing full set with a single one

PR #1166, opened 2024-10-08
fix: include genesis validators in gateway parent validators tracker

sanderpick:
This PR is the result of trying to track down the cause of the following behavior:

- Create a new subnet with min validators = 2
- Create 3 validators and join
- The first 2 validators will be considered "genesis validators", and the third one will be added to the power table (membership in the subnet gateway contract) via a top down staking change
- The subnet receives and stores the validator change, then checkpointing applies those changes. The next power table excludes the first two validators [here](https://github.com/consensus-shipyard/ipc/blob/d0892f21a1d4c7e22106abe6786fb68477e3c37e/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L118). We now have only one validator with power.
- Since the genesis validators are not included in the subnet gateway's parent validator tracker, the new membership set by the top down finalizer excludes the genesis validators, ie, `listActiveValidators()` will be zero when applying this first change [here](https://github.com/consensus-shipyard/ipc/blob/d0892f21a1d4c7e22106abe6786fb68477e3c37e/contracts/contracts/gateway/router/TopDownFinalityFacet.sol#L64).
- Bottom up checkpoints fail because they are only signed by this one validator with power, but the subnet contract on the parent still thinks (correctly) that the total collateral is from three validators, ie, the quorum threshold is higher than what's included in the checkpoint. Checkpoint fails with [`WeightsSumLessThanThreshold`](https://github.com/consensus-shipyard/ipc/blob/d0892f21a1d4c7e22106abe6786fb68477e3c37e/contracts/contracts/lib/LibMultisignatureChecker.sol#L68)
- You can see this reflected in the power as known by cometbft by looking at the nodes' status endpoint (`http://localhost:<comet_rpc_port>/status`). Power is zero for the first two nodes. 

I'm not sure if this fix is ideal, but it works. It just adds the genesis validators to the parent validator tracker before setting the initial membership in the gateway constructor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1166)

### DIFF
--- a/contracts/contracts/GatewayDiamond.sol
+++ b/contracts/contracts/GatewayDiamond.sol
@@ -11,7 +11,7 @@ import {InvalidCollateral, InvalidSubmissionPeriod, InvalidMajorityPercentage} f
 import {LibDiamond} from "./lib/LibDiamond.sol";
 import {LibGateway} from "./lib/LibGateway.sol";
 import {SubnetID} from "./structs/Subnet.sol";
-import {LibStaking} from "./lib/LibStaking.sol";
+import {LibStaking, LibValidatorSet} from "./lib/LibStaking.sol";
 import {BATCH_PERIOD, MAX_MSGS_PER_BATCH} from "./structs/CrossNet.sol";
 
 error FunctionNotFound(bytes4 _functionSelector);
@@ -73,6 +73,19 @@ contract GatewayDiamond {
         // The startConfiguration number is also 1 to match with nextConfigurationNumber, indicating we have
         // empty validator change logs
         s.validatorsTracker.changes.startConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
+        // Add genesis validators to the parent validators tracker.
+        uint256 vLength = params.genesisValidators.length;
+        for (uint256 i; i < vLength; ) {
+            address addr = params.genesisValidators[i].addr;
+            uint256 amount = params.genesisValidators[i].weight;
+            bytes memory metadata = params.genesisValidators[i].metadata;
+            LibValidatorSet.setMetadataFromConstructor(s.validatorsTracker.validators, addr, metadata);
+            LibValidatorSet.recordDeposit(s.validatorsTracker.validators, addr, amount);
+            LibValidatorSet.confirmDeposit(s.validatorsTracker.validators, addr, amount);
+            unchecked {
+                ++i;
+            }
+        }
         // set initial validators and update membership
         Membership memory initial = Membership({configurationNumber: 0, validators: params.genesisValidators});
         LibGateway.updateMembership(initial);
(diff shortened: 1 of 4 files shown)


contracts/contracts/lib/LibStaking.sol
```diff
@@ -203,6 +203,11 @@ library LibValidatorSet {
         validators.validators[validator].metadata = metadata;
     }
 
+    /// @notice Set validator data using metadata from memory bytes
+    function setMetadataFromConstructor(ValidatorSet storage validators, address validator, bytes memory metadata) internal {
```
sanderpick, line 207: Ugly... we can only have `bytes memory` in the constructor.

sanderpick:
Hunch: This issue is also the reason that in federated mode, you have to always include _all_ validators, instead of being able to add them one-by-one.

contracts/contracts/GatewayDiamond.sol
```diff
@@ -73,6 +73,19 @@ contract GatewayDiamond {
...
         // empty validator change logs
         s.validatorsTracker.changes.startConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
+        // Add genesis validators to the parent validators tracker.
+        uint256 vLength = params.genesisValidators.length;
+        for (uint256 i; i < vLength; ) {
+            address addr = params.genesisValidators[i].addr;
+            uint256 amount = params.genesisValidators[i].weight;
+            bytes memory metadata = params.genesisValidators[i].metadata;
+            LibValidatorSet.setMetadataFromConstructor(s.validatorsTracker.validators, addr, metadata);
+            LibValidatorSet.recordDeposit(s.validatorsTracker.validators, addr, amount);
```
cryptoAtwill, line 83: this technically works for `Staking` backed power, but will not work for federated. Ideally it would call record federated based on the permission mode.

cryptoAtwill (commented):
@sanderpick The fix is correct for staking, but need to call record federated for permission mode. But adding permission mode into the constructor is actually quite a big change propagating to fendermint too.

sanderpick:
Ah, makes sense. Somehow I missed https://github.com/consensus-shipyard/ipc/issues/1150. I'm happy to take a stab at passing the mode in the constructor here, but sounds like from slack convo you're picking this up. If so, do you want access to this branch or will you just recreate it? Nbd either way.

cryptoAtwill:
@sanderpick yeah, it's quite tedious to fix the constructor, quite some code needs to be updated in fendermint. @raulk mentioned yesterday that he will check this out to see if there are easier fixes. But I think for collateral based subnets, this is the fix.

cryptoAtwill:
closes this PR as it's fixed in https://github.com/consensus-shipyard/ipc/pull/1302

PR #1167, opened 2024-10-08
refactor: replace LogLevel type with String

sanderpick:
This allows for more flexible log filters like `info,fendermint=debug`.

This also includes a small solhint fix to get `make lint` to work for me locally. My compiler uses solidity `0.8.26`.


karlem (approved):
Thanks for the change @sanderpick

PR #1168, opened 2024-10-08
feat: ipc cli util command to convert f4 to eth address

sanderpick:
This is useful when you need to get the eth contract address from a subnet ID to do, for example, a token approval from a token collateral erc20 contract to the subnet actor contract.


raulk:
Thanks for this @sanderpick!

PR #1170, opened 2024-10-11
feat(node): integration validator reward of contract and bottom up

cryptoAtwill:
This builds on top of previous PR with:
- [x] Adding contract tests
- [x] Updating gateway constructor to accept validator rewarder as parameters. This is to encourage deploy the rewarder first.
- [x] Adding a setter for updating `ValidatorRewarder` contract address. This is needed especially for L2 and L2+, because at genesis, the rewarder will not be deployed or present.
- [x] Integrate caller fvm state to call getters to obtain validator activities
- [x] Update deployment scritps


raulk:

PR #1173, opened 2024-10-15
feat: customizable gas markets (with EIP-1559 default), base fee oscillation, premium distribution

raulk:
This PR integrates these draft and partial PRs into a fully reviewed solution.

- Closes https://github.com/consensus-shipyard/ipc/pull/1098
- Closes https://github.com/consensus-shipyard/ipc/pull/1114
- Closes https://github.com/consensus-shipyard/ipc/pull/1122
- Closes https://github.com/consensus-shipyard/ipc/pull/1132

# New features

## Gas markets in actor space

- Customizable gas markets via a Wasm actor, adhering to the GasMarket standard interface which currently requires implementing two methods: `CurrentReading` and `UpdateUtilization`.
- The gas market is a singleton actor sitting at ID 98. Users can deploy alternative implementations during genesis or an upgrade.
- EIP-1559 as the default gas market, with default constants (minimum base fee, block gas, elasticity, multiplier) inherited from Filecoin.
- EIP-1559 constants can be updated by calling the `SetConstants` method. By default this requires a system caller, thus limiting execution to migrations only. But users can override the caller requirement to something different to allow updates from authorized actors/accounts.

## Premium distributions

Gas premiums previously collected under the Reward actor sitting at address f02 are now distributed to the block producer, crediting them to an f410/0x address derived from their secp256k1 public key.

## Base fee oscillation

The base fee now oscillates based on gas utilization, as determined by the gas market implementation.

## Block gas limit updates

Whenever the block gas limit is updated in the gas market actor, we inform the consensus layer during `end_block` so CometBFT begins to pack blocks adhering to the new block gas limit.

## Message selection

We introduced an incipient message selection function, which users could override by patching the codebase. In the future, we intend to make this modular/pluggable.

# Review

Many adjustments were done directly on the branch throughout review. I tried to atomize the changelog as separate commits, but this was done post-facto so the diffs aren't too clean. The first review commit is 040e18e26fb9857fafbd92402dc11542a86944e1.

# TODO

- [ ] Deal with penalties? https://github.com/consensus-shipyard/ipc/issues/896

---

Closes https://github.com/consensus-shipyard/ipc/issues/925


fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -232,7 +254,7 @@ where
 
     /// Identity of the block creator, if we are indeed executing any blocks.
     pub fn validator_pubkey(&self) -> Option<PublicKey> {
```
cryptoAtwill, line 256: rename this to `block_producer`?

cryptoAtwill (approved):

fendermint/vm/interpreter/src/fvm/state/exec.rs
raulk: Good catch!

raulk: Addressed in https://github.com/consensus-shipyard/ipc/pull/1173/commits/5d96099fb019f60014abadeb77bc3ae72e9d48d4.

PR #1174, opened 2024-10-17
docs: add docs for gas markets.

raulk:


JulissaDantes:
I think its very clear, the only thing I would say is to keep in mind linking the axon-r06 release notes, when it goes live of course.

sanderpick:
+1, this looks great. one silly question: this mechanism applies to transactions coming from the EthAPI and "raw" FVM transactions, correct? I have a vague memory of gas being handled differently in the EthAPI in the current code base, but could be misremembering.

PR #1178, opened 2024-10-19
Docs: Fix a broken reference for the subnets webpage

francoisthire:
While reading the IPC docs, I realized the subnets link from the main page at https://docs.ipc.space/ is broken.

I suspect the link should redirect to https://docs.ipc.space/concepts/subnets and this is what this MR achieves.

### How to test

On the https://docs.ipc.space/ , clicking on the "subnets" link goes to a 404 like page.

On my local fork https://github.com/francoisthire/ipc/tree/main/docs-gitbook we can check the link works as expected.

### Note

I have tried to test it locally with [gitbook-cli](https://github.com/GitbookIO/gitbook-cli) but the tool is deprecated and does not compile anymore with a recent version of node. Maybe there is a way to test it with their online editor at https://www.gitbook.com/ but I did not find how.

### DIFF
--- a/docs-gitbook/README.md
+++ b/docs-gitbook/README.md
@@ -22,7 +22,7 @@ IPC is a scaling solution intentionally designed to achieve considerable perform
 
 It achieves scaling through the permission-less spawning of new blockchain sub-systems, which are composed of subnets.&#x20;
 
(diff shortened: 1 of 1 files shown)


raulk (approved):
Thank you!

PR #1182, opened 2024-10-23
fix: improve Gas Fee Estimation by Integrating Filecoin's EIP-1559-Compatible APIs

### LINKED ISSUE #982: ipc relayer uses fixed gas parameters
From @raulk 

Make it calculate the gas fee cap and gas premium using:
- eth_feeHistory
- eth_maxPriorityFeePerGas

### LINKED ISSUE #1175: Review ipc-cli gas parameters
Akave folks have pointed out that ipc-cli transactions against Filecoin mainnet (e.g. subnet creation, subnet join, fund) seem to remain queued for long. This could be due to us miscalculating gas parameters (concretely the gas fee cap or premium). Even though we calculate our values based on the data returned by the Eth RPC API, it's possible that our code is carrying Ethereum assumptions that work for Calibrationnet (due to it being mostly vacant), but not for Mainnet.

karlem:
Close #1175 
Close #982 

This PR updates our gas fee estimation to leverage the `eth_maxPriorityFeePerGas` and `eth_getBlockByNumber` (to retrieve the current `base_fee_per_gas` from the latest mined block) API calls, using Eth-standard APIs available through Filecoin. Since Filecoin supports EIP-1559, these calls are fully compatible with Filecoin's implementation, aligning with standard EVM practices and ensuring interoperability across EVM chains.

**Key Changes:**
- The `max_priority_fee_per_gas()` and `base_fee_per_gas()` functions retrieve gas fee estimates based on Filecoin's adherence to EIP-1559.
- Gas fee calculation buffers the base fee with a multiplier of 2 to accommodate potential cumulative increases, aligning with the 12.5% maximum cap on base fee growth per block. By doubling the base fee, we account for possible surges in consecutive blocks, ensuring that estimates remain responsive to rapid market changes while providing a buffer for handling block-to-block volatility.

This approach simplifies our fee estimation logic by utilizing available EIP-1559-compatible APIs, improving chances of capturing better gas market dynamics on specific networks (Filecoin in this case) to increase chance of fast transaction inclusion within a block.


ipc/provider/src/manager/evm/gas_estimator_middleware.rs
```diff
@@ -0,0 +1,160 @@
...
+
+            // Set the gas fees directly on the transaction.
+            let tx_req = inner
+                .clone()
+                .max_fee_per_gas(gas_fees.max_fee_per_gas)
+                .max_priority_fee_per_gas(gas_fees.max_priority_fee_per_gas);
+
+            *tx = TypedTransaction::Eip1559(tx_req);
+        } else {
+            return Err(Eip1559GasEstimatorError::failed_to_estimate_gas_not_supported());
```
cryptoAtwill, line 126: why not just skip the error so that other types of txns can go through so that it's more general. Or the purpose is only supporting EIP1559?

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
```diff
@@ -0,0 +1,160 @@
...
+        &self,
+        tx: &mut TypedTransaction,
+        block: Option<BlockId>,
+    ) -> Result<(), Self::Error> {
+        if let TypedTransaction::Eip1559(inner) = tx {
+            let gas_fees = self.estimate_gas_fees().await?;
+
+            // Set the gas fees directly on the transaction.
+            let tx_req = inner
+                .clone()
```
cryptoAtwill, line 120: why need to clone here instead of directly assign?

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
```diff
@@ -0,0 +1,160 @@
...
+                .max_priority_fee_per_gas(gas_fees.max_priority_fee_per_gas);
+
+            *tx = TypedTransaction::Eip1559(tx_req);
+        } else {
+            return Err(Eip1559GasEstimatorError::failed_to_estimate_gas_not_supported());
+        }
+
+        // Delegate to the inner middleware for filling remaining transaction fields.
+        self.inner()
+            .fill_transaction(tx, block)
```
cryptoAtwill, line 131: will the gas be overwritten by inner fill transaction? Will call inner first then overwrite the gas params be safer?

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
```diff
@@ -0,0 +1,160 @@
...
+                // Populate missing gas fees with `fill_transaction`.
+                self.fill_transaction(&mut tx, block).await?;
+            }
+        } else {
+            return Err(Eip1559GasEstimatorError::failed_to_estimate_gas_not_supported());
+        }
+
+        // Proceed to send the transaction with the inner middleware.
+        self.inner()
+            .send_transaction(tx, block)
```
cryptoAtwill, line 156: same question here, the inner will not overwrite the gas params updated right?

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
karlem: > Or the purpose is only supporting EIP1559?

Yes. As these endpoints are only available from EIP1559 upgrade.

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
karlem: I think - from what I have seen, the gas parameter is only set internally when it's not defined. But we define it here.

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
karlem: Nope it should not. Answer above.👆🏻

ipc/provider/src/manager/evm/gas_estimator_middleware.rs
karlem: yeah, even thought the inner is mutable reference the compiler complains

```
ethers::types::Eip1559TransactionRequest::max_fee_per_gas` takes ownership of the receiver `self`, which moves `*inner`
```

cryptoAtwill: I see, the `tx` is `&mut TypedTransaction`.

cryptoAtwill (approved):

PR #1185, opened 2024-10-25
feat(node): txn prioritization based on gas parameters 

### LINKED ISSUE #1183: Enforce new base fee in mempool after gas market base fee update
With the introduction of gas market, block gas limit and base fee rotation were introduced. However, there is a missing piece when the base fee is updated, it should be enforced in the mempool for message selections: i.e. message in mempool below the new base fee, should be deprioritised for message selection.

cryptoAtwill:
This is a follow up to gas market feature and closes #1183.

The implementation follows the method 2 described. A high-level review is as follows:
- Check txn loads the past 5 blocks' base fee and constructs a priority calculator.
- Finds the lowest base fee in the past 5 blocks.
- The txn priority is calculated as outlined in https://github.com/consensus-shipyard/ipc/blob/17796a0f156121606186119aea59bb29a17e26f8/fendermint/vm/interpreter/src/fvm/state/priority.rs#L47

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -589,6 +589,9 @@ where
             .await
             .context("error running check")?;
 
+        // Update the check state.
+        *guard = Some(state);
+
         let mut mpool_received_trace = MpoolReceived::default();
 
         let response = match result {
@@ -604,9 +607,6 @@ where
                 }
             },
         };
-
-        // Update the check state.
-        *guard = Some(state);
 
         mpool_received_trace.accept = response.code.is_ok();
         if !mpool_received_trace.accept {
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -397,6 +397,40 @@ where
             _ => Err(anyhow!("invalid app state json")),
         }
     }
+
+    fn create_check_state(&self) -> anyhow::Result<FvmExecState<ReadOnlyBlockstore<SS>>> {
+        let db = self.state_store_clone();
+        let state = self.committed_state()?;
+
+        // This would create a partial state, but some client scenarios need the full one.
+        // FvmCheckState::new(db, state.state_root(), state.chain_id())
+        //     .context("error creating check state")?
+
+        let mut state = FvmExecState::new(
+            ReadOnlyBlockstore::new(db),
+            self.multi_engine.as_ref(),
+            state.block_height.try_into()?,
+            state.state_params,
+        )
+        .context("error creating check state")?;
+
+        // load txn priority calculator
+        let end = state.block_height() as u64;
+        let start = end
+            .saturating_sub(state.txn_priority_calculator().len() as u64)
+            .max(1);
+        for h in start..=end {
+            let Ok((state_params, _)) = self.state_params_at_height(FvmQueryHeight::Height(h))
+            else {
+                continue;
+            };
+            state
+                .txn_priority_calculator_mut()
+                .base_fee_updated(state_params.base_fee);
+        }
+
+        Ok(state)
+    }
 }
 
 // NOTE: The `Application` interface doesn't allow failures at the moment. The protobuf
@@ -561,22 +595,7 @@ where
 
         let state = match guard.take() {
             Some(state) => state,
-            None => {
-                let db = self.state_store_clone();
-                let state = self.committed_state()?;
-
-                // This would create a partial state, but some client scenarios need the full one.
-                // FvmCheckState::new(db, state.state_root(), state.chain_id())
-                //     .context("error creating check state")?
-
-                FvmExecState::new(
-                    ReadOnlyBlockstore::new(db),
-                    self.multi_engine.as_ref(),
-                    state.block_height.try_into()?,
-                    state.state_params,
-                )
-                .context("error creating check state")?
-            }
+            None => self.create_check_state()?,
         };
 
         let (state, result) = self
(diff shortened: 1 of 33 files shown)


fendermint/app/src/app.rs
```diff
@@ -561,22 +595,7 @@ where
...
-
-                FvmExecState::new(
-                    ReadOnlyBlockstore::new(db),
-                    self.multi_engine.as_ref(),
-                    state.block_height.try_into()?,
-                    state.state_params,
-                )
-                .context("error creating check state")?
-            }
+            None => self.create_check_state()?,
```
cryptoAtwill, line 598: We can avoid loading the 5 blocks state params for `check_tx`, to do this, we must change the `commit` to hold some historical `check_state` instead of purging everything. Not sure if we should do in this PR.

fendermint/vm/interpreter/src/fvm/state/priority.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        if msg.gas_fee_cap >= base_fee {
+            let i = msg.gas_fee_cap.clone() - base_fee + msg.gas_premium.clone();
+            i.atto()
+                .min(&BigInt::from(i64::MAX))
+                .to_i64()
+                .expect("clipped to i64 max")
+        } else {
+            0
+        }
```
raulk, line 51: 1. These kinds of conditionals are easier to follow when written as an early return. If the gas_fee_cap is below the base fee, return 0.
2. Are we sure 0 is the right value here? Why not return i64::MIN if the message is non-includable? Right now messages with gas_fee_cap == base_fee && gas_premium == 0 would rank exactly the same as non-includable messages. Seems wrong?
3. Adding the gas premium to the delta between the gas_fee_cap and the base fee is incorrect. The gas_fee_cap _includes_ the payable base fee **and** the gas premium. When the message is includable, our priority should be proportional to the **effective gas premium**. See how the [effective gas premium](https://github.com/filecoin-project/ref-fvm/blob/a8f9be18e76fcbe5d4b326d8c213eedb58121996/fvm/src/executor/default.rs#L91) is calculated in ref-fvm; copy the logic here.
4. i64::MAX atto is equal to 9.223372036854775807 base units. We are probably safe in clipping the effective premium to i64::MAX, since nobody sane will set such a high premium, but I'm uncomfortable with the panic here. I suggest this instead, which saturates to i64::MAX:

	```rust
	effective_premium.atto().to_i64().unwrap_or(i64::MAX);
	```

cryptoAtwill:
@raulk Updated, please help take another look.

fendermint/vm/interpreter/src/fvm/state/priority.rs
```diff
@@ -0,0 +1,204 @@
...
+use fvm_shared::bigint::BigInt;
+use fvm_shared::econ::TokenAmount;
+use num_traits::{ToPrimitive, Zero};
+
+/// The transaction priority calculator. The priority calculated is used to determine the ordering
+/// in the mempool.
+#[derive(Clone, Debug)]
+pub struct TxnPriorityCalculator {
+    /// Ring buffer of base fee history
+    base_fee_history: Vec<Option<TokenAmount>>,
```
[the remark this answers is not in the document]
cryptoAtwill: I think this was quite early discussion on one of the huddles to use the lowest base fee in recent history to let slightly more transaction to go through to avoid empty blocks. An array is also tuneable in how base fee can be selected. The requirement is updated slightly and base fee selection is now includable/non-includable. Then the most recent base fee should be enough. I can update this accordingly.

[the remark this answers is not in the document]
raulk: We'll need to fix this. The base fee determines inclusion and the burn rate (network fee). By no means should a transaction be included if its gas fee cap is below the current base fee for the block its included in. That's a consensus fault.

cryptoAtwill:
@raulk updated accordingly. Now the priority calculation is based on the latest base fee from the block gas market reading. Anything below will have priority i64::MIN.

fendermint/vm/interpreter/src/fvm/state/priority.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        let msg = create_msg(TokenAmount::from_atto(10), TokenAmount::from_atto(20));
+        assert_eq!(cal.priority(&msg), i64::MIN);
+
+        let msg = create_msg(TokenAmount::from_atto(35), TokenAmount::from_atto(20));
+        assert_eq!(cal.priority(&msg), 5);
+
+        let msg = create_msg(TokenAmount::from_atto(50), TokenAmount::from_atto(20));
+        assert_eq!(cal.priority(&msg), 20);
+
```
raulk, line 70: nit: missing test case for when base fee is below gas fee cap (i.e. includable), and the _effective_ premium is below the indicated premium.

```suggestion

let msg = create_msg(TokenAmount::from_atto(60), TokenAmount::from_atto(20));
        assert_eq!(cal.priority(&msg), 20);
```

raulk (approved):

raulk (approved):
An integration test here would be great.

1. Submit many transactions with high gas limit to stuff blocks and make the base fee spike to X or above.
2. Submit a transaction with gas fee cap way below X but above the minimum base fee.
3. Stop the flood so the base fee decreases, and ensure the transaction does not appear until the base fee is below the gas fee cap.

I will merge this and open an issue for these extra tests.

raulk:
@sanderpick feel free to integrate this fix on your side. There are still some edge cases for us to deal with (e.g. rejecting transactions below minimum base fee).

sanderpick:
> @sanderpick feel free to integrate this fix on your side. There are still some edge cases for us to deal with (e.g. rejecting transactions below minimum base fee).

great, ty!

PR #1186, opened 2024-10-26
fix: for "Nonce Has Already Been Used" Issue During Deployment on Calibrationnet

karlem:
#### Issue
- **Problem:** The error arises because the nonce, which uniquely identifies each transaction from the sender, is reused while deploying multiple contracts on the calibrationnet. This issue does not occur on a local node, likely because transactions are processed immediately in a single-node environment.
- **Cause:** On a public testnet like calibrationnet, network latency, transaction propagation time, and node synchronization can lead to situations where a new transaction is sent before the previous one is fully confirmed, resulting in nonce reuse.
- **Local Node Success:** On a local node, deployments succeed because transactions are mined instantly, preventing any nonce conflicts.

#### Solution
The solution is to manage the nonce automatically by calling getTransactionCount with the pending argument to align with Filecoin.

### DIFF
--- a/.github/workflows/contracts-deployment-test.yaml
+++ b/.github/workflows/contracts-deployment-test.yaml
@@ -1,4 +1,4 @@
-name: 'Contracts: Smoke test deployment'
+name: "Contracts: Smoke test deployment"
 
 # This workflow is triggered from the main CI workflow.
 on:
@@ -17,8 +17,8 @@ jobs:
     - name: Setup Node
       uses: actions/setup-node@v4
       with:
-        node-version: '18.x'
-        cache: 'pnpm'
+        node-version: "18.x"
+        cache: "pnpm"
 
     - name: Install dependencies
       run: cd contracts && make deps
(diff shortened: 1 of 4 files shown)


cryptoAtwill:
Interesting. Even though we are targeting the same node, ideally with one confirmation, it should have updated the nonce usage. Unless the node is load balanced?

karlem:
> Unless the node is load balanced?

It is.


> If this is just about confirmation, then increase to 3 might slow down the deployment 3 times. Or maybe we can try a `try-catch` and handle the error to wait just 1 block time.

Sure we can try. Or we can increase the confirmation to 2. This is just a hot fix.

cryptoAtwill (approved):
@karlem I tried it, and it's working.

But it's really slow deploying all the contracts. We have 27 contracts (maybe more to come?), then 27 * 3 * (30 seconds) ~= 40 min. I would say speed is important too.

cryptoAtwill (commented):
I approved it by accident, my bad. But I think the above comment might be helpful.

karlem:
@cryptoAtwill I think using 2 confirmations with a try-catch mechanism is the sweet spot. I tried 1 confirmation, but it resulted in many retries and also took a while. What do you think?

karlem:
Changed to manually managing nonce.

.github/workflows/contracts-deployment-test.yaml
raulk, line 1: Revert formatting changes?

raulk (approved):

PR #1188, opened 2024-10-30
fix: fix ci clippy error

cryptoAtwill:
Fix current CI clippy error to Box everything.


LePremierHomme (approved):
Works well for me. 

Note: need to manually delete `fendermint:latest` image.

PR #1194, opened 2024-11-05
fix: use eth_getTransactionCount(pending) to get nonce

### LINKED ISSUE #1195: Handling Filecoin Network Nonce Error with Custom Middleware
Issue:

When interacting with the Filecoin Calibration network using ipc-cli, transactions often fail with the error: message: minimum expected nonce is 63, got 62. This typically occurs when sending two consecutive transactions.

Cause:

The error is caused by Filecoin changing the behavior of the eth_getTransactionCount endpoint. To get the correct latest nonce, the "pending" block parameter must now be specified: [Relevant Lotus PR](https://github.com/filecoin-project/lotus/pull/12520).

Solution:

ethers-rs currently does not provide a built-in way to set the "pending" parameter for nonce fetching. The recommended workaround is to implement a custom middleware that explicitly sets the "pending" parameter when fetching the nonce.

karlem:
Close #1195 

Issue:

When interacting with the Filecoin Calibration network using ipc-cli, transactions often fail with the error: message: minimum expected nonce is 63, got 62. This typically occurs when sending two consecutive transactions.

Cause:

The error is caused by Filecoin changing the behavior of the eth_getTransactionCount endpoint. To get the correct latest nonce, the "pending" block parameter must now be specified: [Relevant Lotus PR](https://github.com/filecoin-project/lotus/pull/12520).

Solution:

Fix by setting the transaction block to pending, which makes the get_transaction_count method use the pending parameter, aligning it with Filecoin's expectations.


3benbox:
fwiw, this fix worked for both cross-msg fund-with-token and ipc-cli checkpoint relayer nonce issues we've been seeing on calibration.  🎉

raulk (approved):
Clippy fails, but we're fixing that elsewhere, so good with merging this.

PR #1197, opened 2024-11-06
fix: prevent panic on chain replay

### LINKED ISSUE #1196: `CometBFT` crashes with panic in `fendermint` during `BeginBlock` when `fendermint` is catching up (replaying) from CometBFT
### Description:
We encountered an issue where `cometbft` crashes with a panic caught in `fendermint`. This issue occurs because, in `BeginBlock`, we attempt to resolve the CometBFT validator ID to a public key. However, when `fendermint`’s `data` folder is deleted and `fendermint` is restarted, `cometbft` attempts to start block replay but is not ready for the RPC API connection that `fendermint` requires for this process.

### Steps to Reproduce:
1. Run both `cometbft` and `fendermint`.
2. Wait until a few blocks have been produced.
3. Stop `fendermint` and delete its `data` folder.
4. Restart `fendermint`.

#### `cometbft` Logs Before Crash:
```
I[2024-11-06|15:13:38.920] ABCI Replay Blocks module=consensus appHeight=0 storeHeight=5 stateHeight=5 I[2024-11-06|15:13:51.760] Applying block module=consensus height=1 E[2024-11-06|15:13:51.762] Stopping abci.socketClient for error: read message: EOF module=abci-client connection=consensus I[2024-11-06|15:13:51.762] service stop module=abci-client connection=consensus msg="Stopping socketClient service" impl=socketClient E[2024-11-06|15:13:51.762] consensus connection terminated. Did the application crash? Please restart CometBFT module=proxy err="read message: EOF"
```

#### `fendermint` Panic:
```
2024-11-06T14:13:51.762219Z ERROR fendermint/abci/src/application.rs:212: failed to execute ABCI request: Error { msg: "HTTP error", source: "error trying to connect: tcp connect error: Connection refused (os error 61)", } thread 'tokio-runtime-worker' panicked at /Users/alexei/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-abci-0.7.0/src/v037/server.rs:145:70: called Result::unwrap() on an Err value: HTTP error

Caused by: error trying to connect: tcp connect error: Connection refused (os error 61)

Location: /Users/alexei/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9

Caused by: 0: HTTP error 1: error trying to connect: tcp connect error: Connection refused (os error 61) note: run with RUST_BACKTRACE=1 environment variable to display a backtrace 2024-11-06T14:13:51.995565Z ERROR fendermint/app/src/main.rs:24: panicking stacktrace=" 0: std::backtrace_rs::backtrace::libunwind::trace\n
```

### Cause:
The issue seems to be due to [this line in `validators.rs`](https://github.com/consensus-shipyard/ipc/blob/main/fendermint/app/src/validators.rs#L42), where `fendermint` tries to resolve the validator ID to a public key by connecting to the `cometbft` RPC API during `BeginBlock`. If `cometbft` is not fully ready (due to replay or a fresh start with deleted data), this connection fails, causing `fendermint` to panic and terminate.

karlem:
Close #1196 

Removing the dependency on the CometBFT client in favor of caching. When CometBFT is catching up—replaying from the beginning of the chain to synchronize with the ABCI app—it does not start the RPC API. Unfortunately, our ABCI app relied on the API during consensus events, which made it impossible to replay the chain.

Solved by relying on the exec and app states instead of making calls to CometBFT.


fendermint/app/src/app.rs
```diff
@@ -164,6 +169,8 @@ where
     state_hist_size: u64,
     /// Tracks the validator
     validators: ValidatorTracker<C>,
+
+    latest_consensus_params: Arc<RwLock<Option<TendermintConsensusParams>>>,
     /// The cometbft client
     client: C,
```
LePremierHomme, line 175: Having `client` embedded here (and inside `ValidatorTracker`) creates a circular dependency, which is problematic regardless of the current issue (unless explicit orchestration is implemented). It’s better to remove it entirely.

fendermint/app/src/validators.rs
```diff
@@ -66,4 +63,28 @@ impl<C: Client + Sync> ValidatorTracker<C> {
         let keys = self.public_keys.read().unwrap();
         keys.get(id).copied()
     }
+
+    pub fn hydrate_cache(
```
LePremierHomme, line 67: I assume you added this to prevent `ValidatorTracker` from dispatching calls during `begin_block`. If so, it would be worthwhile to refactor it, as it should no longer use `self.client` at all, and should panic if a validator record is missing.

fendermint/app/src/app.rs
karlem: Yeah, I don't think circular dependency is an issue here, but it's not being used so it should be removed.

karlem:
@LePremierHomme could you have another look please?

fendermint/app/src/app.rs
```diff
@@ -803,22 +813,31 @@ where
...
-                .consensus_params(tendermint::block::Height::try_from(request.height)?)
-                .await?
-                .consensus_params;
-
-            if consensus_params.block.max_gas != new_block_gas_limit as i64 {
-                consensus_params.block.max_gas = new_block_gas_limit as i64;
-                Some(consensus_params)
+        // Updates the validator cache with the new validator set.
+        let validators_ids_with_keys: Vec<(TendermintAccountId, TendermintPublicKey)> =
+            validator_updates
```
cryptoAtwill, line 818: If i rmb correctly, `validator_updates` is the power change, so it's about changes, not setting validators.

cryptoAtwill (commented):
@karlem A quick question, how will the cache be recovered after crashes? Seems not queried when booting?

fendermint/app/src/app.rs
karlem: Hmm. In that case where would be a good place to get the validator changes?

karlem:
> @karlem A quick question, how will the cache be recovered after crashes? Seems not queried when booting?

Yeah good point. We would loose it after the crash. Do you think maybe saving it to the app state might be a way to go?

fendermint/app/src/app.rs
cryptoAtwill: So we just need to check if the power is 0, that means the validator is removed, we should purge instead. See: https://github.com/consensus-shipyard/ipc/blob/409c1387ddf978a76c999bd8a19c7291dd094fd1/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L434. Else it's just an `upsert`.

fendermint/app/src/consensus_params.rs
```diff
@@ -0,0 +1,47 @@
...
+        self.current_block_gas_limit
+            .store(gas_limit, Ordering::Relaxed);
+    }
+
+    pub fn get_current_block_gas_limit(&self) -> i64 {
+        self.current_block_gas_limit.load(Ordering::Relaxed)
+    }
+
+    pub fn get_genesis_consensus_params(&self) -> Option<TendermintConsensusParams> {
+        let consensus_params = self.genesis_consensus_params.lock().unwrap();
```
LePremierHomme, line 44: This would panic if `init_chain` is not called prior to `end_block`. From @cryptoAtwill's comment, I understand that this behavior is expected, and the params will need to be persistent. 

And for validators, if `modify_exec_state` outputs just the diff in power, then the state of the validators will need to be continuously stored, unless there's a simple way to locally fetch it when bootstrapping with no `init_chain`.

cryptoAtwill:
@karlem The rabbit hole gets deeper. But I think it's possible to read the state during boot up. The validators and gas limits are all stored in the contract or actor, so technically we can call the getters from the store directly.

raulk:
@cryptoAtwill that's a good point; it would also take us one tiny step closer to our desired end state of having as much as possible on chain.

fendermint/app/src/app.rs
```diff
@@ -803,22 +813,31 @@ where
...
+        let consensus_param_updates =
+            if self.consensus_params.get_current_block_gas_limit() != new_block_gas_limit {
+                self.consensus_params
+                    .set_current_block_gas_limit(new_block_gas_limit);
+
+                let mut updated_params = self
+                    .consensus_params
+                    .get_genesis_consensus_params()
+                    .ok_or_else(|| anyhow!("missing genesis consensus params"))?;
+                updated_params.block.max_gas = new_block_gas_limit;
```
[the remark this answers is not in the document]
karlem: I think I prefer it here. At least we can see what is going on exactly in the app.

karlem:
@cryptoAtwill Yeah, I think working with the contract makes more sense than this. I would ditch this in favor of using the contracts, as they are part of the app state—much more solid than this.

karlem:
This is related to that: https://github.com/consensus-shipyard/ipc/pull/1166

karlem:
@cryptoAtwill Changed the implementation to rely on app and exec states instead.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -246,7 +246,7 @@ where
         Message = VerifiableMessage,
         DeliverOutput = SignedMessageApplyRes,
         State = FvmExecState<DB>,
-        EndOutput = (PowerUpdates, BlockGasLimit),
+        EndOutput = (PowerUpdates, BlockGasLimit, CurrentValidators),
```
cryptoAtwill, line 249: If the caller maintains power updates, then there is no need to return `CurrentValidators`? It can be derived?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -252,7 +254,17 @@ where
             PowerUpdates::default()
         };
 
-        let ret = (updates, next_gas_market.block_gas_limit);
+        let current_validators = self
+            .gateway
+            .current_membership(&mut state)
+            .context("failed to update membership")?
```
cryptoAtwill, line 260: this comment should be read membership, not update.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -48,7 +50,7 @@ where
     /// Return validator power updates and the next base fee.
     /// Currently ignoring events as there aren't any emitted by the smart contract,
     /// but keep in mind that if there were, those would have to be propagated.
-    type EndOutput = (PowerUpdates, BlockGasLimit);
+    type EndOutput = (PowerUpdates, BlockGasLimit, CurrentValidators);
```
cryptoAtwill, line 53: I think this approach is not correct, because essentially you are query the current list of validators at every block end. Then what is the point of maintaining a cache? Also most of the time, there is no validator change, so the querying of validator list is not necessary.

I would say returning the list of current validators when PowerUpdates is non empty would be more effecient.

cryptoAtwill (dismissed):

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -252,7 +254,17 @@ where
...
         };
 
-        let ret = (updates, next_gas_market.block_gas_limit);
+        let current_validators = self
+            .gateway
+            .current_membership(&mut state)
+            .context("failed to update membership")?
+            .validators
+            .iter()
+            .filter_map(|v| PublicKey::parse_slice(&v.metadata, None).ok())
```
cryptoAtwill, line 263: Might be better to add a method in `gateway` to encapsulate the logic

fendermint/vm/interpreter/src/chain.rs
karlem: But how would that work during the restarts? Since the cache is not stored?

fendermint/vm/interpreter/src/fvm/exec.rs
karlem: Yeah, I have noticed it. Fixed

raulk:
I'm wrapping my head around the problem and the proposed solution to help push things forward. Here are some notes.

If I'm following correctly, this attempts to fix an edge case whereby during a CometBFT startup, the latter may attempts to perform a chain replay either from the WAL or from the network, by feeding blocks to the ABCI app. The issue arises because we recently introduced logic to fetch the block proposer's public key so we could credit gas premiums to their on-chain account. This happens in [`begin_block`](https://github.com/consensus-shipyard/ipc/blob/c74a21cbb1d990ce19e8b5099854f3f3b18d479c/fendermint/app/src/app.rs#L734) and was introduced in #1173. Because our local cache is empty, we fall back to CometBFT, whose RPC API has not started yet, therefore making us fail.

I think all of this can be greatly simplified if instead of querying CometBFT, we query our power table from the gateway actor and match on the block proposer's identity. The gateway already has public keys. Such a call would be entirely inside the state tree, so it has no dependencies on CometBFT, and would break the problematic cycle.

In fact, we already do this in `end_block` in order to calculate power table updates.

1. `end_block` calls `interpreter.end()`: https://github.com/consensus-shipyard/ipc/blob/c74a21cbb1d990ce19e8b5099854f3f3b18d479c/fendermint/app/src/app.rs#L802
2. When we create a checkpoint, it may come with power table updates: https://github.com/consensus-shipyard/ipc/blob/3ad2c0f52fff1323fd32c20d1d3507bcc12094a3/fendermint/vm/interpreter/src/fvm/exec.rs#L222
3. We query the current power table: https://github.com/consensus-shipyard/ipc/blob/3ad2c0f52fff1323fd32c20d1d3507bcc12094a3/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L71
4. We already have a helper that queries the gateway actor in the state tree and returns domain objects, containing validator pubkeys: https://github.com/consensus-shipyard/ipc/blob/03e3ebdca9859bfcf9ea8a1fa69109fde451d6ed/fendermint/vm/interpreter/src/fvm/state/ipc.rs#L178

In a nutshell, I don't think we need one more cache here, nor to manage its lifecycle, nor anything like that. Assuming I'm right, we can kill the `ValidatorTracker` entirely and simply replace it with the above.

@cryptoAtwill does this sound right to you?

karlem:
@cryptoAtwill and @raulk Changed this to a replaceable cache that generates internal state based on execution state. Hopefully, this aligns closely with what we want.

fendermint/app/src/app.rs
```diff
@@ -245,6 +263,7 @@ where
                     power_scale: 0,
                     app_version: 0,
                 },
+                consensus_params: default_consensus_params(),
```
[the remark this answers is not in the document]
raulk: Pushed a refactor turning this into an Option, removing the default dummy parameters (this is a smell in Rust), and minorly reworking some logic.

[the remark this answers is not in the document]
raulk: [`d029337` (#1197)](https://github.com/consensus-shipyard/ipc/pull/1197/commits/d029337e160c08905e8e0a11567cb93f4361069c)

raulk (approved):

fendermint/app/src/app.rs
```diff
@@ -82,6 +86,8 @@ pub struct AppState {
     oldest_state_height: BlockHeight,
     /// Last committed version of the evolving state of the FVM.
     state_params: FvmStateParams,
+    /// Tendermint consensus params.
+    consensus_params: TendermintConsensusParams,
```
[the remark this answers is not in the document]
karlem: Indeed.

fendermint/app/src/validators.rs
```diff
@@ -1,69 +1,52 @@
...
+            .ok_or_else(|| anyhow!("validator not found"))
     }
 }
+
+fn validator_to_map_entry(v: &Validator<Power>) -> Result<(TendermintId, PublicKey)> {
+    let tendermint_pub_key: TendermintPubKey = TendermintPubKey::try_from(v.public_key.clone())?;
+    let id = TendermintId::from(tendermint_pub_key);
+    let key = *v.public_key.public_key();
+    Ok((id, key))
+}
```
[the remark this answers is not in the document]
karlem: I initially had it inside the closure but didn’t like how it looked, so I was moving it back and forth. However, it’s fine to keep it in the closure.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -331,7 +331,7 @@ pub fn emit_trace_if_check_checkpoint_finalized<DB>(
 where
     DB: Blockstore + Clone,
 {
-    if !gateway.enabled(state)? {
+    if !gateway.is_root(state)? {
```
raulk, line 334: (I know this method already existed, but it didn't appear to be called before, so taking a few moments to revisit the naming.) 

I think we should rename this to `is_anchored`, since it's only gonna be called inside subnets (IPC is a client for subnets). We tend to use the term _root_ to refer to existing L1s where we root a hierarchy.

raulk (approved):
A few nits; other than that, LGTM. Glad we're landing this after the several twists and bends this has taken.

Issue #1190, opened 2024-11-05
Metrics validator label in expanded formet


3benbox:
### Issue type

Bug

### Have you reproduced the bug with the latest dev version?

Yes

### Version

main

### Custom code

Yes

### Describe the issue

The `validator` label value for ipc metrics is printing the public key struct instead of the value

Example:
```
ipc_bottomup_checkpoint_signed_height{validator="PublicKey(Affine { x: Field { n: [66693404, 23532371, 40178610, 55223286, 21415173, 15532913, 25444414, 53056045, 57924191, 4134036], magnitude: 1, normalized: false }, y: Field { n: [20460645, 23136642, 11840773, 46331495, 4715384, 27082347, 25105138, 33510637, 47085913, 150247], magnitude: 1, normalized: false }, infinity: false })"}
```

This value should be just the public key, or an address?

### Repro steps

Collect metrics, observe resulting metric values.

PR #1201, opened 2024-11-14
fix: metrics validator label to use address instead of pk struct formatting

LePremierHomme:
Closes https://github.com/consensus-shipyard/ipc/issues/1190

I used address instead of the public key hex encoding. Can change that if wanted.

### DIFF
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -125,7 +125,7 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         .with_max_retries(settings.broadcast.max_retries)
         .with_retry_delay(settings.broadcast.retry_delay);
 
-        ValidatorContext::new(sk, broadcaster)
+        ValidatorContext::new(sk, addr, broadcaster)
     });
 
     let testing_settings = match settings.testing.as_ref() {
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -261,7 +261,7 @@ where
                 role: CheckpointSignedRole::Own,
                 height: height.value(),
                 hash: HexEncodableBlockHash(cp.block_hash.to_vec()),
-                validator: validator_ctx.public_key,
+                validator: validator_ctx.addr,
             });
 
             tracing::debug!(?height, "submitted checkpoint signature");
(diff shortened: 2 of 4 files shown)


raulk (approved):
It's unfortunate we have to do this conversion. Two things:

1. We should make `record_metrics()` fallible by returning a `Result<()>`. Then we can avoid having panicking code, and the higher level component would log errors in the error event type. @karlem could you drop a link to an example where we use that event type?
2. Would add a TODO so we try to propagate the FVM address in the validator context.

raulk:
@LePremierHomme nit: the "closing" keyword is not supported by GitHub to link the issue, but "closes", "close", "closed" are. https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

LePremierHomme:
> 1. We should make `record_metrics()` fallible by returning a `Result<()>`. Then we can avoid having packing code, and the higher level component would log errors in the error event type. @karlem could you drop a link to an example where we use that event type?

Not sure I understand the purpose of having it fallible. To propagate encoding errors? 

> 2. Would add a TODO so we try to propagate the FVM address in the validator context.

The public key is what's used as the validator key for comparisons. I guess it would be better to just derive `Address` as a view, or fix it as an additional field in the context. It seems to be currently needed just for the metrics, so I could also just change `CheckpointSigned` to contain the address instead of the public key.

karlem:
We should make record_metrics() fallible by returning a Result<()>. This way, we can avoid having packing code, and the higher-level component would log errors using the error event type. @karlem, could you provide a link to an example where we use this event type?
In general, I agree—failing to record a metric should not cause a panic that shuts down the whole node. Here's an example of how to emit a tracing error: https://github.com/consensus-shipyard/ipc/blob/409c1387ddf978a76c999bd8a19c7291dd094fd1/fendermint/vm/interpreter/src/fvm/exec.rs#L200.

However, given point 2:

Would add a TODO so we try to propagate the FVM address in the validator context.

I would suggest we include this change in the current PR. It’s a simple modification—see here: https://github.com/consensus-shipyard/ipc/blob/409c1387ddf978a76c999bd8a19c7291dd094fd1/fendermint/app/src/cmd/run.rs#L128.

By including the address in the event directly, we can skip the conversion altogether and eliminate the need to make record_metrics() fallible. At least for now!

karlem:
@LePremierHomme , the address is already being converted in run.rs. I suggest including it in the ValidatorContext and then using it directly in the event.

raulk:
SGTM!

raulk:
Thank you, @karlem! 🙏

karlem (approved):
@LePremierHomme I think it's fine like this now. Since we do not cause panic in the `record_metrics`.

Issue #1153, opened 2024-09-26
Add the current chain id as a label to the metrics


3benbox:
### Issue type

Feature Request

### Have you reproduced the bug with the latest dev version?

Yes

### Version

main

### Custom code

Yes

### Describe the issue

When a host participates on multiple subnets, or changes to a different subnet, it would be really useful to know the chain the metrics are associated with.

Cometbft already does this, so reusing their label and value creation would be preferred.

<img width="801" alt="image" src="https://github.com/user-attachments/assets/ef6d4a45-fb81-45c6-b061-2b4004c28b94">

### Repro steps

Observe the output generated from the metrics endpoint.

PR #1204, opened 2024-11-15
feat: add `subnet_id` to metrics as a global label

LePremierHomme:
Closes https://github.com/consensus-shipyard/ipc/issues/1153


karlem (approved):
Perfect, small changes. Looks good :)

PR #1205, opened 2024-11-15
feat(node): integration rewards, review comments.

raulk:
Calibration testing:
SubnetRegistryDiamond deployed at 0xbCC8CAbbf2f33D5665B176692CDc6B1135bAE05E
GatewayDiamond deployed at 0x689D7fAD1a77d9aDc7d2C724c366C8EF8D919184


contracts/contracts/activities/Activity.sol
```diff
@@ -51,6 +63,7 @@ struct ValidatorClaimProof {
 }
 
 /// The proofs to batch claim validator rewards in a specific subnet
+/// REVIEW(raulk): Delete this type. Make the method just take the subnet ID and the list of claim proofs.
```
cryptoAtwill, line 66: @raulk I think this type is introduced to handle multiple subnet ids at once. With this structure, might be easier to manage types. Or this requirement is not needed?

contracts/contracts/activities/Activity.sol
```diff
@@ -3,54 +3,70 @@ pragma solidity ^0.8.23;
...
+        uint64 blocksCommitted;
+        /// @dev Other metadata
+        bytes metadata;
+    }
 }
 
 /// The proof required for validators to claim rewards
 struct ValidatorClaimProof {
-    ValidatorSummary summary;
+    uint64 checkpointHeight;
```
cryptoAtwill, line 63: @raulk Add in checkpoint height to identify which commitment to use

contracts/contracts/structs/CrossNet.sol
```diff
@@ -31,7 +31,7 @@ struct BottomUpCheckpoint {
     /// @dev Batch of messages to execute.
     IpcEnvelope[] msgs;
     /// @dev The activity summary from child subnet to parent subnet
-    ActivitySummary activities;
+    Consensus.Compressed activities;
```
cryptoAtwill, line 34: @raulk using compressed instead over full as only `Compressed` needs to be persisted with bottom up checkpoint in the parent, there is no need to commit Full report to the storage in the parent subnet.

contracts/contracts/activities/Activity.sol
```diff
@@ -3,54 +3,70 @@ pragma solidity ^0.8.23;
...
+    }
+
+    struct ValidatorDetail {
+        /// @dev The validator whose activity we're reporting about, identified by the Ethereum address corresponding
+        /// to its secp256k1 pubkey.
+        address validator;
+        /// @dev The number of blocks committed by this validator during the summarised period.
+        uint64 blocksCommitted;
+        /// @dev Other metadata
+        bytes metadata;
```
cryptoAtwill, line 57: `metadata` still needed?

cryptoAtwill (commented):
@raulk I have push the changes to the contract. Please help take a quick look as there are some slight adjustments to fit the overall flow. Once this is ok, fendermint is very easy to change.

contracts/contracts/activities/Activity.sol
raulk: I think this should be an unwrapped argument. It is not part of the proof per se, but rather an extra argument we need to pass to the method to identify the height against we're presenting the proof. For the batch claim, we should take an array of heights and an array of proofs.

contracts/contracts/structs/CrossNet.sol
raulk: Absolutely. This was my IDE messing up.

contracts/contracts/activities/Activity.sol
raulk: Not any longer. After discussing with Basin, they found a way to do without. And we prefer to have user data as custom extensions to the activity summary instead of adhoc fields sprinkled around.

raulk: Not any longer. After discussing with Basin, they found a way to do without. And we prefer to have user data as custom extensions to the activity summary instead of adhoc fields sprinkled around.

contracts/contracts/activities/Activity.sol
raulk: So there's two levels of batching:
1. Claim rewards corresponding to multiple checkpoint heights for a single subnet at once.
2. Claim rewards from multiple subnets, over multiple checkpoint heights for each subnet, at once.

I would only support (1) and not (2) as:
1. It limits attack surface by constraining interactions to a single subnet within a transaction.
2. It sounds like premature optimization to me, and once introduced it will be impossible to remove even if nobody uses it.

My suggestion is that we descope (2) and limit ourselves to (1). With that in mind, we can kill the extra type and simply have three arguments:

- subnet ID
- array of checkpoint heights
- array of claims

contracts/contracts/activities/Activity.sol
raulk: See https://github.com/consensus-shipyard/ipc/pull/1205#discussion_r1846530654.

fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -369,22 +359,6 @@ where
         f(&mut self.params);
         self.params_dirty = true;
     }
-
-    pub fn call_state(&self) -> anyhow::Result<FvmCallState<DB>> {
-        let externs = self.executor.externs().read_only_clone();
```
raulk, line 374: @cryptoAtwill can we also clean up and remove the read_only_clone() method?

cryptoAtwill: removed!

contracts/contracts/activities/Activity.sol
cryptoAtwill: removed, resolving now

cryptoAtwill (approved):

Issue #1093, opened 2024-07-26
Migrate Resolver Metrics to New Observability Event-Based Architecture


karlem:
## Summary
Migrate the resolver metrics in the IPLD resolver to use the new observability event-based architecture. This involves revisiting the current metrics and deciding on which ones to keep and which ones to discard.

## Current Code Link
- [Resolver Metrics](https://github.com/consensus-shipyard/ipc/blob/103d68907690c696fb26c4eb0b4f6ad74810068c/ipld/resolver/src/stats.rs#L21)

## Proposed Solution
1. **Migrate Metrics:** Update the current metrics implementation to align with the new observability event-based architecture.
2. **Review and Refactor:** Revisit all existing metrics to evaluate their relevance and usefulness. Decide which metrics to keep and which to discard.
3. **Implement Event-Based Architecture:** Utilize the new event-based architecture to handle the metrics.

PR #1207, opened 2024-11-19
refactor(ipld): migrate metrics scheme

LePremierHomme:
Closes https://github.com/consensus-shipyard/ipc/issues/1093

Current work contains the migration to the event-based architecture. 
Still need to evaluate the usefulness of each metric.


ipld/resolver/src/observe.rs
```diff
@@ -0,0 +1,312 @@
...
+use prometheus::{
+    register_histogram, register_int_counter, register_int_gauge, Histogram, IntCounter, IntGauge,
+    Registry,
+};
+use std::error::Error;
+use std::time::Duration;
+
+register_metrics! {
+    PING_RTT: Histogram =
+        register_histogram!("ping_rtt", "Ping roundtrip time");
```
karlem, line 19: I think we need to prefix all of the metrics with `ipld_resolver`. Otherwise just seeing `ping_rtt` metric on it's own does not say much.

ipld/resolver/src/observe.rs
```diff
@@ -0,0 +1,312 @@
...
+        match self {
+            Self::Received(_) => IDENTIFY_RECEIVED.inc(),
+        }
+    }
+}
+
+#[derive(Debug)]
+#[allow(dead_code)]
+pub enum IdentifyFailureEvent {
+    Failure(PeerId, Box<dyn Error>),
```
karlem, line 148: Do we need to store the `Box<dyn Error>`? We don't really need for anything so I wonder of just a string would suffice?

karlem:
Thanks @LePremierHomme! I took a quick look, and overall it looks good. I may have more comments tomorrow. We will also need to add these events to the [documentation](https://github.com/consensus-shipyard/ipc/blob/main/docs/fendermint/observability.md) and ensure that the traces produce user-friendly output.

ipld/resolver/src/observe.rs
LePremierHomme: Fixed: https://github.com/consensus-shipyard/ipc/pull/1207/commits/2037c80cd86f27ac456534f7d1d8da400df81b25

ipld/resolver/src/observe.rs
LePremierHomme: Yes, it's not needed, I mainly used it to signal that's it's an error, since I used tuples instead of structs. 
Fixed here: https://github.com/consensus-shipyard/ipc/pull/1207/commits/915cd016c13cee6fa58beeb8a77a73bb6d275c5b

LePremierHomme:
I added domain namespace prefix to the traces, although it's not applied elsewhere. 
So it looks like this now:
```
2024-11-20T05:13:52.295725Z  INFO tracing_event: domain="IPLD_Resolver/Ping" event=Success(PeerId("1AkHsDuvAgj8ZAAgGaaqw82CMRU6e7ii2WA4XmXDsj1qkQ"), 500ms)
2024-11-20T05:13:52.296359Z  WARN tracing_event: domain="IPLD_Resolver/Ping" event=Timeout(PeerId("1AkHsDuvAgj8ZAAgGaaqw82CMRU6e7ii2WA4XmXDsj1qkQ"))
2024-11-20T05:13:52.296402Z  WARN tracing_event: domain="IPLD_Resolver/Ping" event=Failure(PeerId("1AkHsDuvAgj8ZAAgGaaqw82CMRU6e7ii2WA4XmXDsj1qkQ"), "err")
```
As for the docs, I'll add it as the final step once we're done with the review.

ipld/resolver/src/observe.rs
```diff
@@ -0,0 +1,331 @@
...
+    Registry,
+};
+use std::time::Duration;
+
+register_metrics! {
+    IPLD_RESOLVER_PING_RTT: Histogram =
+        register_histogram!("ipld_resolver_ping_rtt", "Ping roundtrip time");
+
+    IPLD_RESOLVER_PING_TIMEOUT: IntCounter =
+        register_int_counter!("ipld_resolver_ping_timeouts", "Number of timed out pings");
```
karlem, line 21: I think we could merge these together and have a label instead? Instead of having 3 metrics it can be one with status label Same goes for others (only if it makes sense in the given context). What do you think?

ipld/resolver/src/observe.rs
```diff
@@ -0,0 +1,331 @@
...
+        register_histogram!("ipld_resolver_content_resolve_peers", "Number of peers found for resolution from a subnet");
+
+    IPLD_RESOLVER_CONTENT_CONNECTED_PEERS: Histogram =
+        register_histogram!("ipld_resolver_content_connected_peers", "Number of connected peers in a resolution");
+
+    IPLD_RESOLVER_CONTENT_RATE_LIMITED: IntCounter =
+        register_int_counter!("ipld_resolver_content_rate_limited", "Number of rate limited requests");
+}
+
+impl_traceables!(TraceLevel::Info, "IPLD_Resolver/Ping", PingEvent);
```
karlem, line 87: We actually allow users to filter by the domain. I think having separate domain for each event type makes it impossible to filter for all events from resolver.

ipld/resolver/src/observe.rs
LePremierHomme: Overloading as a label will require secondary indexing and will be counter productive if per-event counters are wanted. I think they are wanted, but I don't have a strong opinion about it.

ipld/resolver/src/observe.rs
LePremierHomme: So are you suggesting that the domain will be simply `IPLD_Resolver`? because previously (before my last commit) they were also segregated (e.g., `Ping`, `Discovery`, etc.).
Another option is to implement domain namespace filtering, but it might be an overkill.

karlem: Yeah, I think having a unified domain might work better. We do allow to filter by it.

karlem: > Another option is to implement domain namespace filtering, but it might be an overkill.

Yes. I think it would be.

ipld/resolver/src/observe.rs
karlem: I just didn’t want to generate a lot of similar metrics. I don’t think we gained enough experience debugging the resolver based on these, so it’s a tough call. We can leave it as it is for now.

ipld/resolver/src/observe.rs
LePremierHomme: fixed here: https://github.com/consensus-shipyard/ipc/pull/1207/commits/8b5051c286a09ec61d3f2c20ffcd0446f87344a1

karlem (approved):

PR #1211, opened 2024-11-21
fix: enforce paris evm version

dtbuchholz:
The newest foundry nightly build now uses cancun as the default EVM version (https://github.com/foundry-rs/foundry/pull/9131). This breaks everything in IPC because transient storage opcodes are not supported.

### DIFF
--- a/contracts/foundry.toml
+++ b/contracts/foundry.toml
@@ -16,6 +16,7 @@ remappings = [
     "murky/=lib/murky/src/",
 ]
 allow_paths = ["../node_modules"]
+evm_version = "paris"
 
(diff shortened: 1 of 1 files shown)


dtbuchholz:
question—is the IPC team planning to support Cancun evm compatibility? common tooling like openzeppling and foundry are now releasing tooling that uses Paris by default, so it requires a bit of vigilance to stay on top of it—which is easier for internal teams like Basin, but might be challenging for newcoming devs using Basin.

note: our tests started failing when using the latest `nightly` foundry, and pinning the evm version fixed it.

Issue #1095, opened 2024-08-01
Add list validator cmd to ipc-cli


karlem:
Currently, in the IPC CLI under the subnet, there is a `get-validator` command but no command to list all validators (`list-validators`) in a specific subnet. Including such a command would arguably be useful.

PR #1221, opened 2024-11-28
feat(cli): add `list-validators` command

LePremierHomme:
Closes https://github.com/consensus-shipyard/ipc/issues/1095


contracts/test/integration/SubnetActorDiamond.t.sol
```diff
@@ -75,6 +75,9 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
         require(saDiamond.getter().bottomUpCheckPeriod() == params.bottomUpCheckPeriod, "unexpected bottom-up period");
         require(saDiamond.getter().majorityPercentage() == params.majorityPercentage, "unexpected majority percentage");
         require(saDiamond.getter().getParent().toHash() == _parentId.toHash(), "unexpected parent subnetID hash");
+        require(saDiamond.getter().genesisValidators().length == 0, "unexpected genesis validators");
+        require(saDiamond.getter().getActiveValidators().length == 0, "unexpected active validators");
+        require(saDiamond.getter().getWaitingValidators().length == 0, "unexpected waiting validators");
```
raulk, line 80: Good work strengthening the assertions here with the new accessors!

ipc/cli/src/commands/subnet/list_validators.rs
```diff
@@ -0,0 +1,42 @@
...
+            println!("{}: {}", addr, info);
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(
+    name = "list validators",
+    about = "List the info of all the validators in the subnet"
```
raulk:
```suggestion
    about = "List the info of all the validators in the subnet, as viewed by the parent"
```

raulk: Important to note this, since one would default to thinking that we're querying the subnet itself.

raulk: Pushed in 8f128686.

raulk (approved):

Issue #1225, opened 2024-12-03
IPC messages are omitted in log filtering


cryptoAtwill:
In `get_logs` api, only [chain message is checked](https://github.com/consensus-shipyard/ipc/blob/6f5521a21a44c747577b58f1e1a15be0f7a384e2/fendermint/eth/api/src/apis/eth.rs#L979) and IPC messages are omitted. So any message coming from gateway executions are omitted.

raulk:
Fixed in https://github.com/consensus-shipyard/ipc/pull/1226.

PR #1226, opened 2024-12-04
fix: patch missing ipc messages in eth get_logs

cryptoAtwill:
This PR addresses https://github.com/consensus-shipyard/ipc/issues/1225.

This PR updates the matching of query addresses against txn from and to to only user signed txns. IPC message does not really need to check that because it's obviously system call.

### DIFF
--- a/fendermint/eth/api/src/apis/eth.rs
+++ b/fendermint/eth/api/src/apis/eth.rs
@@ -974,21 +974,28 @@ where
 
                 let mut log_index_start = 0usize;
                 for ((tx_idx, tx_result), tx) in tx_results.iter().enumerate().zip(block.data()) {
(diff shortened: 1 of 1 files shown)


fendermint/eth/api/src/apis/eth.rs
```diff
@@ -974,21 +974,29 @@ where
...
 
-                    // Filter by address.
-                    if !addrs.is_empty()
-                        && !addrs.contains(&msg.message().from)
-                        && !addrs.contains(&msg.message().to)
-                        && addrs.intersection(&emitters).next().is_none()
-                    {
-                        continue;
-                    }
+                    let no_intersection =
```
karlem, line 979: Nit: `no_overlap_with_emitters` might be better?

PR #1232, opened 2024-12-13
feat(node): support ethereum legacy txn

cryptoAtwill:
The current main does not allow `send_raw_transaction` for legacy transactions. This PR removes this assumption. To achieve it, the following changes are made:
- Adding `OriginKind` enum to explicitly specify the original transaction type. The type information is lost when converting to Fvm messages.
- Support gas params from legacy transactions as specified in https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0091.md
- Any ethereum transaction sent from fendermint client is enforced to be eip1559.


cryptoAtwill:
Favor #1253

PR #1234, opened 2024-12-15
fix: use current exec state when querying validator table

### LINKED ISSUE #1233: Panic During Execution: Validator Not Found in Validators Cache
Issue:
Initially, the validators start and function correctly. However, when a new validator is added via top-down finality (causing the power table to change) and the validators cache is refreshed, a panic occurs for the newly added validator.

Root Cause:
When fetching the validator membership to refresh the validators cache, the read_only_state method is used to retrieve the execution state. However, this method returns the state from the previous block (as it relies on the state store) instead of the current execution state. Since the new set of validators has not yet been committed (flushed to the state store), the newly added validator is missing from the cache.

Proposed Solution:
Replace the use of the read_only_state method with the current execution state to ensure the validators cache is updated correctly with the latest set of validators.

karlem:
Close #1233 

The issue has been resolved by using the current execution state instead of the previous state retrieved through the read-only view. This ensures that the validators cache accurately reflects the latest set of validators, preventing the panic.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -420,12 +420,15 @@ where
 
     /// Replaces the current validators cache with a new one.
     async fn refresh_validators_cache(&self) -> Result<()> {
-        let mut state = self
-            .read_only_view(None)?
-            .ok_or_else(|| anyhow!("exec state should be present"))?;
+        // TODO: This should be read only state, but we can't use the read-only view here
+        // because it hasn't been committed to state store yet.
+        self.modify_exec_state(|mut s| async {
+            let mut cache = self.validators_cache.lock().await;
+            *cache = Some(ValidatorCache::new_from_state(&mut s.1)?);
+            Ok((s, ()))
+        })
+        .await?;
 
-        let mut cache = self.validators_cache.lock().await;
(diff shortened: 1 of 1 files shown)


cryptoAtwill:
@karlem Somehow the first impression that strikes is shouldnt we try to update the `read_only_view` to reflect latest state instead of passing mutable ref of state to a read only function?

raulk (commented):
@cryptoAtwill Unfortunately the FVM executor does not expose read-only semantics through its `execute_message` API. 

So we'd have to add a method to "fork" the executor/machine using the intermediate state tree root (which you can get from `machine.state_tree()`) and a [`BufferedBlockstore`](https://github.com/filecoin-project/ref-fvm/blob/9abb85b8e6b15b945834692f0421ed7ade20e4d9/fvm/src/blockstore/buffered.rs#L201) over the original Blockstore (`machine.blockstore()`), which we discard after we finish the query.

Alternatively, we could forego this state query entirely and:

1. Guarantee we have always have a fully initialized ValidatorCache at the start.
2. Merge the power updates onto our ValidatorCache.

raulk (approved):
Merging this and opening https://github.com/consensus-shipyard/ipc/issues/1236 to fix the right way.

Issue #927, opened 2024-05-13
Support Eth legacy transactions


maciejwitowski:

raulk:
---

We only support legacy transactions partially for estimation purposes:

https://github.com/consensus-shipyard/ipc/blob/ebb743f42882efa7af8f4a31970e9c854893238b/fendermint/eth/api/src/conv/from_eth.rs#L20

---

This is how Filecoin bolted on Eth legacy tx support recently:

https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0091.md

---

This is needed to deploy things like Gnosis Safe on IPC subnets; unfortunately legacy transactions are quite ossified in some libraries / tools.

PR #1235, opened 2024-12-16
feat(node): support legacy transactions

cryptoAtwill:
Closes https://github.com/consensus-shipyard/ipc/issues/927.

The current main does not allow `send_raw_transaction` for legacy transactions. This PR removes this assumption. To achieve it, the following changes are made:
- Adding `OriginKind` enum to explicitly specify the original transaction type. The type information is lost when converting to Fvm messages.
- Support gas params from legacy transactions as specified in https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0091.md
- Any ethereum transaction sent from fendermint client is enforced to be eip1559.
- `TransactionCache` is updated to use `(TypeTransaction, Signature)` instead of the response dto `Transaction`.


fendermint/eth/api/src/conv/from_eth.rs
```diff
@@ -23,37 +24,72 @@ pub fn to_fvm_message(tx: TypedTransaction) -> JsonRpcResult<Message> {
...
+                transaction_type: Some(2u64.into()),
+                access_list: Some(tx.access_list.clone()),
+                other: Default::default(),
+            })
+        }
+        _ => error::error_with_revert(
+            ExitCode::USR_ILLEGAL_ARGUMENT,
+            "txn type not supported",
+            None::<Vec<u8>>,
         ),
```
raulk, line 92: Why are we duplicating this logic here? `send_raw_transaction` is already producing this error?

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -640,10 +640,10 @@ where
     let msghash = et::TxHash::from(ethers_core::utils::keccak256(rlp.as_raw()));
     tracing::debug!(?sighash, eth_hash = ?msghash, ?tx, "received raw transaction");
 
-    if let Some(tx) = tx.as_eip1559_ref() {
-        let tx = from_eth::to_eth_transaction(tx.clone(), sig, msghash);
-        data.tx_cache.insert(msghash, tx);
-    }
+    data.tx_cache.insert(
+        msghash,
+        from_eth::to_eth_transaction_response(&tx, sig, msghash)?,
```
raulk, line 645: What do you mean with transaction _response_ here? This module is an API, so I interpret as if this is returning the response the API sends to the caller, which isn't the case? (unless there's an error)

We are duplicating the transaction type validation in this method and when we calculate the origin kind above. Can we refactor this to make it DRY?

BTW -- how can we have a module called `from_eth` that then has `to_eth` methods? Wouldn't they cancel each other out? 😆

fendermint/eth/api/src/conv/from_eth.rs
```diff
@@ -23,37 +24,72 @@ pub fn to_fvm_message(tx: TypedTransaction) -> JsonRpcResult<Message> {
...
-        value: tx.value.unwrap_or_default(),
-        gas: tx.gas.unwrap_or_default(),
-        max_fee_per_gas: tx.max_fee_per_gas,
-        max_priority_fee_per_gas: tx.max_priority_fee_per_gas,
-        // Strictly speaking a "Type 2" transaction should not need to set this, but we do because Blockscout
-        // has a database constraint that if a transaction is included in a block this can't be null.
-        gas_price: Some(
-            tx.max_fee_per_gas.unwrap_or_default()
-                + tx.max_priority_fee_per_gas.unwrap_or_default(),
+) -> Result<et::Transaction, JsonRpcError> {
```
raulk, line 31: I really dislike that the cache uses `et::Transaction` -- this type is intended as a return type from the JSON-RPC API, not to store canonical signed transactions.

fendermint/eth/api/src/conv/from_eth.rs
```diff
@@ -23,37 +24,72 @@ pub fn to_fvm_message(tx: TypedTransaction) -> JsonRpcResult<Message> {
...
+    match &tx {
```
raulk, line 32: This all looks very repetitive. Can't you populate the common values from `TransactionEssentials` via [`essentials()`](https://github.com/foundry-rs/foundry/blob/eef87de35375311f134fb4261c18455ced8022a5/crates/anvil/core/src/eth/transaction/mod.rs#L690)? And then add the specifics?

Isn't there a method to normalize typed transactions into the Transaction response type in ethers?

raulk (changes requested):

fendermint/eth/api/src/conv/from_eth.rs
cryptoAtwill: > Isn't there a method to normalize typed transactions into the Transaction response type in ethers?
From what I found, not really. My guess is the Transaction response contains transaction index, which does not make sense to do the conversion using `From`.

cryptoAtwill:
@raulk Because the types and `from_eth::*` and `to_eth::*` are used in multiple places, I just did a extension to make it work. I think it's a good chance to do a more in depth refactoring. Will update accordingly.

fendermint/eth/api/src/mpool.rs
```diff
@@ -21,11 +22,12 @@ use crate::{cache::Cache, state::Nonce, HybridClient};
 
 const RETRY_SLEEP_SECS: u64 = 5;
 
+pub type SignedTransaction = (TypedTransaction, et::Signature);
 /// Cache submitted transactions by their Ethereum hash, because the CometBFT
 /// API would not be able to find them until they are delivered to the application
 /// and indexed by their domain hash, which some tools interpret as the transaction
 /// being dropped from the mempool.
-pub type TransactionCache = Cache<et::TxHash, et::Transaction>;
```
raulk, line 28: Honestly, now that I look at the final solution, I think this was better the way it was before. We ended up leaking all of ethers' type complexity into Tendermint. In fact, I don't think TypedTransaction should've leaked anywhere at all. I'll work on rolling this back, removing `to_eth_typed_transaction`, `to_eth_legacy_request`, `to_eth_eip1559_request`, etc.

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -612,6 +615,14 @@ pub async fn get_uncle_by_block_number_and_index<C>(
     Ok(None)
 }
 
+fn normalize_signature(sig: &mut et::Signature) -> JsonRpcResult<()> {
```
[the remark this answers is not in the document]
raulk: Looks like it's not exposed...

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -439,8 +439,11 @@ where
                 .state_params(FvmQueryHeight::Height(header.header.height.value()))
                 .await?;
             let chain_id = ChainID::from(sp.value.chain_id);
+
             let hash = msg_hash(&res.tx_result.events, &res.tx);
-            let mut tx = to_eth_transaction(msg, chain_id, hash)?;
+            let mut tx = to_eth_transaction_response(msg, chain_id)?;
+            debug_assert_eq!(hash, tx.hash);
```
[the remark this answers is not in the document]
cryptoAtwill: they should be the same, `debug_assert_eq` should be fine for release build though, more like a sanity check for tests.

fendermint/eth/api/src/apis/eth.rs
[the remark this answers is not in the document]
cryptoAtwill: Yeah, a lot of methods are not exposed. Could be helpful util tools.

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -621,23 +632,23 @@ where
...
+
     let sighash = tx.sighash();
-    let msghash = et::TxHash::from(ethers_core::utils::keccak256(rlp.as_raw()));
+    let msghash = tx.hash(&sig);
     tracing::debug!(?sighash, eth_hash = ?msghash, ?tx, "received raw transaction");
 
-    if let Some(tx) = tx.as_eip1559_ref() {
-        let tx = from_eth::to_eth_transaction(tx.clone(), sig, msghash);
-        data.tx_cache.insert(msghash, tx);
-    }
```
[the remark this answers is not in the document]
cryptoAtwill: It won't remain forever as it's using LRU cache. It will just stay for some time, depending on the configuration.

fendermint/eth/api/src/mpool.rs
cryptoAtwill: I feel the `conv::*` should not be in `message` crate. It should be shifted to `eth` crate. That might be better.

raulk (approved):
Let's go ahead and merge this to unblock users who are waiting for this feature to land. We can clean this up in main later.

PR #1239, opened 2024-12-19
fix(node): harden transaction inclusion logic to account for fee market

raulk:
This PR improves correctness of tx inclusion logic on two fronts.

1. Begins rejecting messages with invalid limits at CheckTx time:
    - Gas fee cap below minimal base fee.
    - Gas limit above maximum block gas limit.

Some type shenanigans via downcasts were necessary to deal with #1241. Tested by extending a materializer test.

2. Exclude non-includable messages at proposal time.
    - I removed the MessagesSelector trait because it was unusable. The vector holding selectors cannot used boxed objects because the trait is not object-safe as it has methods with generic parameters. Also, there's no point over-engineering this prematurely.
    - Also removed the decreasing sorting by gas limit; we prefer to randomize for security (otherwise an attacker could drip brutally overestimated transactions just below the block gas limit to DoS the chain).

3. Randomize message selection to prevent attackers from predicting or controlling message inclusion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1239)


fendermint/vm/interpreter/src/chain.rs
```diff
@@ -228,6 +230,14 @@ where
                     }
                 }
                 ChainMessage::Signed(signed) => {
+                    // We do not need to check against the minimium base fee because the gas market
+                    // is guaranteed to be capped at it anyway.
+                    if signed.message.gas_fee_cap < gas_market.base_fee {
```
LePremierHomme, line 235: What about doing that in tx selection for block inclusion (as a leader)? When checking for mempool inclusion only the min fee is verified.

fendermint/testing/materializer/tests/docker_tests/standalone.rs
```diff
@@ -90,6 +90,16 @@ async fn test_sent_tx_found_in_mempool() {
                     }
                 }
 
+                // This transaction is invalid because the gas fee cap is below the network's minimum base fee.
+                // It should be rejected at submission time via CheckTx (which will also reject it at block validation time).
+                let invalid = Eip1559TransactionRequest::new()
+                    .to(to)
+                    .max_fee_per_gas(1)
```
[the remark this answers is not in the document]
raulk: We just don't have access to the min base fee here; nor is it available through an API. We could technically fetch the block header of the genesis block and take that base fee as the min base fee, but I'm sure it would buy us anything.

fendermint/vm/interpreter/src/chain.rs
raulk: @cryptoAtwill do we not already exclude transactions whose gas fee cap is below the network's base fee at block production time? I was strangely not able to find that logic here: https://github.com/consensus-shipyard/ipc/blob/702d49b619623915772fe935a86e59a89744c3b1/fendermint/vm/interpreter/src/selector.rs

(Do we need a second selector?)

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -228,6 +230,14 @@ where
                     }
                 }
                 ChainMessage::Signed(signed) => {
+                    // We do not need to check against the minimium base fee because the gas market
+                    // is guaranteed to be capped at it anyway.
```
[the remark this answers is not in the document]
raulk: Good point; I think we should check the min base fee is satisfied here since CheckTx won't be invoked for transactions we never saw but that got included by another validator in a proposal.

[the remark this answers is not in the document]
raulk: Sorry, I was mistaken. This is code is correct. The current gas market conditions are determined by our **own** state transitions. So here we validate message's gas fee cap is above or equal to the **current** gas market base fee, as determined by our own state. The EIP1559 gas market actor floors the base fee to the minimal base fee here:

https://github.com/consensus-shipyard/ipc/blob/7e4b87c1065ac2d8ac5cdfb7980fad236b9ea46e/fendermint/actors/gas_market/eip1559/src/lib.rs#L150

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: @raulk I dont remember discussing that in `prepare`. I think we were mostly discussing the cometbft side of things.

raulk: What do you mean with "the CometBFT side of things"? We agreed that we'd assign negative priorities to non-includable transactions in CheckTX, to instruct CometBFT to prefer sending us includable transactions, but we'd still have to filter out non-includable transactions in `prepare`.

raulk: I'll fix it in this PR.

cryptoAtwill: if we need to enforce below base fee exclusion, we probably should add the other selector to enforce in fendermint just to be sure. Cometbft should have done one round of selection based on priority.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -475,6 +486,22 @@ where
...
+                //  In the meantime we are compelled to downcast to access the actual types in use.
+                //  This check cannot be performed in the last interpreter (which has access to the FvmExecState) because
+                //  that one returns an exit code, whereas illegal messages do not lead to any execution and therefore
+                //  they do not benefit from an exit code.
+                let fvm_exec_state = (&state as &dyn Any)
+                    .downcast_ref::<FvmExecState<ReadOnlyBlockstore<DB>>>()
+                    .context("inner state not downcastable to FvmExecState")?;
+                let gas_market = fvm_exec_state.block_gas_tracker().current_gas_market();
+                if msg.message.gas_limit > gas_market.block_gas_limit
+                    || msg.message.gas_fee_cap < gas_market.min_base_fee
```
cryptoAtwill, line 500: A quick question, why using `min_base_fee` here? Currently `min_base_fee` is not updated so I assume it's updated in a system upgrade? Is it a fixed value that can be smaller than `base_fee`?

raulk: This is to reject transactions whose base fee cap is below the minimum acceptable base fee. Otherwise those transactions would sit forever in the mempool because the base fee would **never** go low enough to include them.

cryptoAtwill: I misunderstood, should have no issue here.

cryptoAtwill (approved):

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -555,30 +582,44 @@ fn relayed_bottom_up_ckpt_to_fvm(
...
+        let accepted = total_gas_limit_consumed + gas_limit <= total_gas_limit;
+        if accepted {
+            total_gas_limit_consumed += gas_limit;
+        }
+        accepted
+    };
 
-    Ok(user_msgs.into_iter().map(ChainMessage::Signed).collect())
+    // Perform the selection.
+    let selected = signed
```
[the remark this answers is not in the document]
raulk: added here: https://github.com/consensus-shipyard/ipc/pull/1239/commits/4d26e53bac9720c6abeeb5f1c109da5f4172fe62

LePremierHomme (approved):

phutchins:
Closing due to conflicts.

PR #1252, opened 2025-01-14
feat: add general cross messages docs

### LINKED ISSUE #1250: Add cross messaging docs

karlem:
Close #1250


docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+    /// @dev Outgoing nonce for the envelope, ensuring uniqueness per subnet and preventing replay attacks.
+    uint64 localNonce;
+    /// @dev Value being sent with the message (e.g., tokens).
+    uint256 value;
+    /// @dev ABI-encoded message payload.
+    bytes message;
+    /// @dev Original nonce from the source subnet, useful for tracing messages across subnets.
+    uint64 originalNonce;
+}
+```
```
raulk, line 44: Have we analyzed the packing efficiency of these parameters as per Solidity/EVM alignment? It might be more efficient to put the `value` after `from`, the nonces together, and leave the `message` last.

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Message Types of Interest
+
+- **Call**: Represents a request sent to a destination contract.
```
raulk, line 48: It's worth to note what would happen if the destination is an account, since a caller cannot reliably determine the kind of the callee at call time.

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+- **Result**: Represents the response to a `Call` message, which could either be the result or an error report.
```
raulk, line 49: Don't `Transfer` messages also generate a `Result`?

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Contract Requirements
+
+Both sending and receiving contracts must implement the `IpcExchange` abstract contract from the IPC SDK to handle cross messages. Key methods to implement include:
```
raulk, line 53: The correct term is "cross-net messages" or "xnet messages". "Cross messages" was a poor past usage we're trying to deprecate.

raulk, line 53: Technically speaking they do NOT need to implement this abstract contract. This contract is provided for convenience and ease of usage; it deals with the low-level plumbing, e.g. correlation, error handling, etc. The framework will simply call the interface methods -- we should make that distinction clear.

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Handling Incoming Calls
```
raulk, line 55: When documenting interfaces like this, it's advisable to also document the functionality, error handling, edge cases, etc. Some questions a reader will predictably have:

- When exactly is this called? (answer: as part of the checkpoint execution)
- What happens if this method reverts? How do I signal an error?
- What does the return value represent?
- How should I size the logic in this method? (e.g. what happens if it's computationally too expensive to fit in a block because I do crazy stuff)

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+This must be implemented in the contract receiving the message:
+
+```solidity
+function _handleIpcCall(
+    IpcEnvelope memory envelope,
+    CallMsg memory callMsg
+) internal virtual returns (bytes memory);
+```
+
+#### Handling Results
```
raulk, line 66: Similar to above, this should clearly indicate how the correlation works, and what guarantees it provides. Also, what happens if I revert?

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Outcome Types
+
+- **Ok**: Message was delivered successfully, and the `ret` field contains the response from the called contract.
+- **SystemErr**: An error occurred during message transmission. The return value indicates the `InvalidXnetMessageReason` (see [IPCErrors.sol](../../contracts/contracts/errors/IPCErrors.sol)).
+- **ActorErr**: A custom error returned by the called contract.
+
+## How It Works
+
+1. A custom contract calls `sendContractXnetMessage` on the local `Gateway` with a message inside the IPC envelope (of kind `Call`). The envelope includes the destination subnet ID, which is used for routing.
```
raulk, line 105: Hm, the envelope contains the destination _address_, which is prefixed with the destination _subnet path_.

docs/fendermint/general_cross_messages.md
```diff
@@ -0,0 +1,120 @@
+## Overview
+
+General cross messages enable the transmission of arbitrary messages between IPC subnets. They are particularly useful for transferring tokens (when supported, more on this later) or transmitting arbitrary data. This is not a feature that allows users to trigger messages from wallets; rather, it is a framework for cross-subnet contract-to-contract communication.
```
raulk, line 3: But also account to contract communication? A wallet _can_ send a transfer to an account or contract on another subnet, and it travels through the same mechanism as described here.

raulk (changes requested):

docs/fendermint/general_cross_messages.md
karlem: Good point! I’ve rearranged the fields so that smaller fields (including the nonces) are grouped together, followed by value, and finally the dynamic message. This takes advantage of how EVM allocates storage slots, reducing wasted space and helping optimize gas usage.

docs/fendermint/general_cross_messages.md
karlem: Yes, but we do not mention transfer in this context.

docs/fendermint/general_cross_messages.md
karlem: Only smart contracts can trigger `sendContractXnetMessage`. While a wallet can interact with a smart contract that makes this call, it would be inaccurate to imply a direct invocation from a wallet.

raulk: AFAIK the intention of this document is to describe the General Message Passing rails and APIs, not just the Call message type, right? Wallets can call `Gateway#release` with no restrictions, which under the hood emits a `Transfer` type cross-net message, which in turn results in a `Result` response eventually, right? My impression was that this doc is not restricted to what's downstream of `sendContractXnetMessage` only.

raulk (approved):
Approved in anticipation. Three pending items, though:

1. Resolve this comment thread: https://github.com/consensus-shipyard/ipc/pull/1252#discussion_r1915203039
2. Fix tests -- probably due to binary layout changes in structs.
3. Rebuild and commit the new storage layout descriptors.

docs/fendermint/general_cross_messages.md
karlem: Yes, that's because these documents are intended for L3+ subnets, and currently the release and fund actions are limited to one level of communication (parent to child and vice versa). I will add a mention about this.

PR #1256, opened 2025-01-21
fix(node): clippy

cryptoAtwill:
Currently CI is auto using latest ubuntu, which is now 24.04. This version of ubuntu is using glibc 2.39 and causing the debian image to fail as it only supports 2.36. This PR mainly downgrade ubuntu to 22.04 so that the image can be built correctly.

At the same time, fixing some minor clippy issue.


karlem:
I think this change should be good to go, but we need the end-to-end (E2E) tests to pass to ensure that nothing has stopped working.

cryptoAtwill:
Closing this PR and in favour of #1257.

PR #1269, opened 2025-02-06
rename infra to deployment

drahnr:


crates/binding/src/checkpointing_facet.rs
```diff
@@ -0,0 +1,2291 @@
+pub use checkpointing_facet::*;
```
cryptoAtwill, line 1: I would say not committing these generated rust abi files, they will create tons of conflicts among PRs. Why not just generated them on CI?

Makefile
```diff
@@ -5,6 +5,8 @@ default:
 	cd contracts && make gen
 	cd crates && (cargo build --release && ./target/release/ipc-cli --version && ./target/release/fendermint --version)
 
+# podman has a compatible API, and is better supported
```
karlem, line 8:
```suggestion
# Uses 'docker' as the default container runtime.
# Can be overwritten to 'podman' for a compatible API and potentially enhanced support
```

karlem (approved):
@drahnr A NIT suggestion. Otherwise looks go to go.

crates/binding/src/checkpointing_facet.rs
drahnr: Addressed in #1268

PR #1270, opened 2025-02-07
docs: add markdown linter

drahnr:
All our documentation is in markdown format, but it's not linted for consistency. This PR adds `markdownlint-cli2` checks per PR, to replicated locally a `Makefile` rule was added.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1270)

### DIFF
--- a/.github/workflows/commonmark-lint.yaml
+++ b/.github/workflows/commonmark-lint.yaml
@@ -1,0 +1,25 @@
+name: "Commonmark/Markdown: Lint"
+
+# This workflow is triggered from the main CI workflow.
+on:
+  workflow_call:
+
+jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/commonmark-lint.yaml
```diff
@@ -0,0 +1,25 @@
+name: "Commonmark/Markdown: Lint"
+
+# This workflow is triggered from the main CI workflow.
```
karlem, line 3: I might be missing something, but where does it get triggered?

drahnr: Good catch

PR #1271, opened 2025-02-07
feat(node): add connection limits

drahnr:
During a previous migration, the connection limits were disabled, this re-introduces them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1271)

### DIFF
--- a/ipld/resolver/src/service.rs
+++ b/ipld/resolver/src/service.rs
@@ -16,6 +16,7 @@ use ipc_api::subnet_id::SubnetID;
 use ipc_observability::emit;
 use libipld::store::StoreParams;
 use libipld::Cid;
+use libp2p::connection_limits::ConnectionLimits;
 use libp2p::futures::StreamExt;
 use libp2p::swarm::SwarmEvent;
 use libp2p::{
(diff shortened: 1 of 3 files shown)


ipld/resolver/src/service.rs
```diff
@@ -160,35 +161,29 @@ where
...
+            limits,
             store,
         )?;
 
-        // NOTE: Hardcoded values from Forest. Will leave them as is until we know we need to change.
+        let swarm_config = libp2p::swarm::Config::with_tokio_executor()
+            .with_notify_handler_buffer_size(std::num::NonZeroUsize::new(20).expect("Not zero"))
+            .with_per_connection_event_buffer_size(64);
 
-        // TODO: Where this these go? Used to be `SwarmBuilder::connection_limits`
```
karlem, line 173: I think it looks good, but out of curiosity, how do you interpret this comment? I’m wondering why it was commented out.

ipld/resolver/src/service.rs
```diff
@@ -301,6 +296,7 @@ where
             BehaviourEvent::Discovery(e) => self.handle_discovery_event(e),
             BehaviourEvent::Membership(e) => self.handle_membership_event(e),
             BehaviourEvent::Content(e) => self.handle_content_event(e),
+            BehaviourEvent::ConnectionLimits(_) => {}
```
cryptoAtwill, line 299: any log you want to emit here?

cryptoAtwill (approved):

ipld/resolver/src/service.rs
drahnr: The argument is `type Void`, so carries zero information. The `ConnectionHandler` is a `dummy`. So the path should never be reached, however, I wasn't comfortable putting an `unreachable!` there.

ipld/resolver/src/service.rs
drahnr: The person migration did not know where this functionality migrated to.

Issue #1236, opened 2024-12-17
refactor: expose API to perform intermediate read-only queries on state


raulk:
There are several occasions in which the ABCI app needs to perform a query on the intermediate state prior to committing it. One such example is #1234.

Currently we're working around the lack of such an API by using modify_exec_state, which downgrades safety.

Let's implement the corresponding API so we can retain safety, and let's migrate illegitimate use cases to it.

## Prior discussion

> @cryptoAtwill Unfortunately the FVM executor does not expose read-only semantics through its `execute_message` API. 
> 
> So we'd have to add a method to "fork" the executor/machine using the intermediate state tree root (which you can get from `machine.state_tree()`) and a [`BufferedBlockstore`](https://github.com/filecoin-project/ref-fvm/blob/9abb85b8e6b15b945834692f0421ed7ade20e4d9/fvm/src/blockstore/buffered.rs#L201) over the original Blockstore (`machine.blockstore()`), which we discard after we finish the query.

_Originally posted by @raulk in https://github.com/consensus-shipyard/ipc/pull/1234#pullrequestreview-2508707731_

cryptoAtwill:
@raulk I think the above mentioned method to obtain intermediate state tree root is not possible. Currently the `StateTree` implementation does not support reading the root cid unless it's `flushed`. To obtain the intermediate root cid, we have to flush  state tree in between the blocks, which might not be desirable.

One way I can think of is just create a read only `Executor` implementation. But this could be error prone. 

Another way I can think of is forking `ref-fvm` and make some changes. Currently the difficulty without forking is:
1. `DefaultExecutor` directly referencing the `Machine` in this [line](https://github.com/filecoin-project/ref-fvm/blob/9abb85b8e6b15b945834692f0421ed7ade20e4d9/fvm/src/executor/default.rs#L43).
2. `Machine` exposes state tree directly in this [line](https://github.com/filecoin-project/ref-fvm/blob/9abb85b8e6b15b945834692f0421ed7ade20e4d9/fvm/src/machine/mod.rs#L69).
3. In `DefaultExecutor`, it is directly changing the mut ref to`StateTree` instead of calling through `Machine`, such as this [line](https://github.com/filecoin-project/ref-fvm/blob/9abb85b8e6b15b945834692f0421ed7ade20e4d9/fvm/src/executor/default.rs#L460).
4. All the intermediate changes are cached in `StateTree` and the state tree cannot be swapped to another implementation

[...]

PR #1276, opened 2025-02-12
feat(node): support read only state

cryptoAtwill:
This PR addresses https://github.com/consensus-shipyard/ipc/issues/1236, i.e. read only state by directly working on the `StateTree`. For read only transactions, they will be applied in a new transaction on top of the `StateTree`, at the end of the txn, they will be reverted ALWAYS.

Because `Machine` holds the state tree, one has to convert the `FvmExecState` => `Machine` => `ReadOnlyExecState` => `FvmExecState`.

As the the `DefaultExecutor` in `ref-fvm` owns `Machine` and `Machine` owns the `StateTree`, working directly on the `StateTree` requires replicating some simplified behaviour from `DefaultExecutor`, such as mapping `from`/`to` to actor id, processing input parameters.

Action item:
- [ ] Adding more test cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1276)

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/state/exec.rs
+++ b/fendermint/vm/interpreter/src/fvm/state/exec.rs
@@ -7,7 +7,7 @@ use crate::fvm::activity::actor::ActorActivityTracker;
 use crate::fvm::externs::FendermintExterns;
 use crate::fvm::gas::BlockGasTracker;
 use crate::fvm::state::priority::TxnPriorityCalculator;
-use anyhow::Ok;
+use anyhow::{anyhow, Ok};
 use cid::Cid;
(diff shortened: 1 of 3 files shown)


raulk (changes requested):
I'd rather contribute a patch to ref-fvm to expose with_transaction on the DefaultExecutor. That should be generally useful to others, and just a few lines of code upstream vs. a few hundreds for us to maintain forever.

If that stalls, we should fork ref-fvm, since we'll anyway need to soon to introduce conditional gas charging, and address modularity (IPC-BTC).

cryptoAtwill:
@raulk sounds good, I will migrate there

PR #1275, opened 2025-02-11
(WIP) docs revamp.

raulk:
This reorganizes our docs tree in the following ways:

- Merges inner docs subtrees into a top level docs folder
- Restructures the site docs (docs.ipc.space) into these sections:
  - Overview (could be Introduction)
  - Concepts
  - Getting Started (Setup)
  - Develop (developer guides)
  - Usage (user guides)
  - Troubleshooting
  - Reference (API, SDK, CLI...)
  - Specs (a symlink to docs/specs)

## TODO

- [ ] Discuss and align on final structure
- [ ] Establish redirects from old content to new content (pretty bad for SEO to destroy links)
- [ ] Revisit all existing content and re-write/improve/deprecate
- [ ] Decide which pieces of new content we need and track those separately

PR #1282, opened 2025-02-13
docs: fmt all commonmark

drahnr:
Since #1275 might still take a while, we might want to land the formatting earlier and move on

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1282)


karlem (approved):
Cool!

PR #1253, opened 2025-01-17
fix(node): migrate crates to ext folder

cryptoAtwill:
This PR migrates some of the crates in existing cargo toml to ext by forking them to address #1000. The approach is just forking the github repos and migrate to `ext` folder.

closes #1000.

PR #1283, opened 2025-02-14
feat: use fs-err for sake of better debugability

drahnr:
As part of trying to debug #1253 , I repeatedly found myself debugging path issues. Using `fs-err` yields a `path` with every error message, over just an `os error 2`. In some cases, using `strace` is burdenful and we might be able to avoid reaching for it with this PR in most cases.


.github/workflows/contracts-test.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
           cache: false
 
       - name: Cache Solidity ABI artifacts
-        uses: actions/cache@v2
+        uses: actions/cache@v4
```
karlem, line 24: Why in this PR?

contracts/ops/check-rust-binding.sh
```diff
@@ -3,10 +3,10 @@ set -eu
 set -o pipefail
 
 # checks if there are changes in rust binding
-if [[ `git status ./binding --porcelain` ]]; then 
+if [[ `git status ../contract-bindings --porcelain` ]]; then
```
karlem, line 6: Why is this part of this PR?

.github/workflows/contracts-test.yaml
drahnr: Otherwise you get warnings / errors due to obsoletion

karlem (approved):

PR #1286, opened 2025-02-14
feat: support dynamic Fendermint config loading in Materializer

karlem:
Support dynamically loading the Fendermint configuration in Materialier without rebuilding the image. This will significantly improve the testing feedback loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)

### DIFF
--- a/fendermint/testing/materializer/src/arb.rs
+++ b/fendermint/testing/materializer/src/arb.rs
@@ -289,6 +289,7 @@ fn gen_subnets(
                 } else {
                     Some(choose_one(g, parent_nodes))
                 },
+                fendermint_additional_config: None,
             };
             let id = NodeId::arbitrary(g);
             node_ids.push(id.clone());
--- a/fendermint/testing/materializer/src/docker/mod.rs
+++ b/fendermint/testing/materializer/src/docker/mod.rs
@@ -767,8 +767,8 @@ impl Materializer<DockerMaterials> for DockerMaterializer {
         's: 'a,
     {
         // Overwrite the env file which has seed addresses, then start the node (unless it's already running).
-        node.start(seed_nodes).await?;
-        node.wait_for_started(*STARTUP_TIMEOUT).await?;
+        node.start_all_containers(seed_nodes).await?;
+        node.wait_for_apis_to_start(*STARTUP_TIMEOUT).await?;
         // Trying to avoid `Tendermint RPC error: server returned malformatted JSON (no 'result' or 'error')` on first subnet creation attempt.
         tokio::time::sleep(Duration::from_secs(5)).await;
         Ok(())
--- a/fendermint/testing/materializer/src/docker/node.rs
+++ b/fendermint/testing/materializer/src/docker/node.rs
@@ -9,6 +9,8 @@ use std::{
     time::{Duration, Instant},
 };
 
+use tokio::try_join;
+
 use anyhow::{anyhow, bail, Context};
 use bollard::Docker;
 use ethers::{providers::Middleware, types::H160};
@@ -400,15 +402,19 @@ impl DockerNode {
         let fendermint = match fendermint {
             Some(c) => c,
             None => {
-                let creator = make_runner(
-                    FENDERMINT_IMAGE,
-                    volumes(vec![
-                        (keys_dir.clone(), "/fendermint/keys"),
-                        (fendermint_dir.join("data"), "/fendermint/data"),
-                        (fendermint_dir.join("logs"), "/fendermint/logs"),
-                        (fendermint_dir.join("snapshots"), "/fendermint/snapshots"),
-                    ]),
-                );
+                let mut volume_mappings = vec![
+                    (keys_dir.clone(), "/fendermint/keys"),
+                    (fendermint_dir.join("data"), "/fendermint/data"),
+                    (fendermint_dir.join("logs"), "/fendermint/logs"),
+                    (fendermint_dir.join("snapshots"), "/fendermint/snapshots"),
+                ];
+
+                if let Some(ref additional_config) = node_config.fendermint_additional_config {
+                    volume_mappings
+                        .push((additional_config.clone(), "/fendermint/config/dev.toml"));
+                }
+
+                let creator = make_runner(FENDERMINT_IMAGE, volumes(volume_mappings));
 
                 creator
                     .create(
@@ -481,7 +487,7 @@ impl DockerNode {
         })
     }
 
-    pub async fn start(&self, seed_nodes: &[&Self]) -> anyhow::Result<()> {
+    pub async fn start_all_containers(&self, seed_nodes: &[&Self]) -> anyhow::Result<()> {
         let cometbft_seeds = collect_seeds(seed_nodes, |n| {
             let host = &n.cometbft.hostname();
             let id = n.cometbft_node_id()?;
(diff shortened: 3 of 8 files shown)


drahnr (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 7 files reviewed, 1 unresolved discussion

___
*[`fendermint/testing/materializer/src/manifest.rs` line 56 at r2](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJJKUepCc8SUnrWP2N4:-OJJKUepCc8SUnrWP2N5:b-me2pi) ([raw file](https://github.com/consensus-shipyard/ipc/blob/d8a3622a971a1d97f1da5103fdaed338073e3051/fendermint/testing/materializer/src/manifest.rs#L56)):*
> ```XML
> 
> impl Manifest {
>     /// Read a manifest from file and resolve any relative extra_config paths.
> ```

nit: We should strive to encode code as code, here: `extra_config`

drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @cryptoAtwill and @karlem)

___
*[`fendermint/testing/materializer/tests/manifests/layer2.yaml` line 66 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJJPh_o6dZbVxXBEm8o:-OJJPh_o6dZbVxXBEm8p:b-yhk8ag) ([raw file](https://github.com/consensus-shipyard/ipc/blob/afe837f26fa0c92645370fdd5b2af2c812f0eb13/fendermint/testing/materializer/tests/manifests/layer2.yaml#L66)):*
> ```YAML
>         seed_nodes: []
>         ethapi: true
>         fendermint_additional_config: ./test.toml
> ```

Can't we use inline yaml/json rather than pulling more fs dependencies?

karlem (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @cryptoAtwill and @drahnr)

___
*[`fendermint/testing/materializer/tests/manifests/layer2.yaml` line 66 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJJPh_o6dZbVxXBEm8o:-OJJUL79CIc60EGNyjaj:b-7zi8ni) ([raw file](https://github.com/consensus-shipyard/ipc/blob/afe837f26fa0c92645370fdd5b2af2c812f0eb13/fendermint/testing/materializer/tests/manifests/layer2.yaml#L66)):*
<details><summary><i>Previously, karlem (Karel Moravec) wrote…</i></summary><blockquote>

We can, but then it's not very DRY, and it will bloat the file. Imagine setting the same exact config for all nodes—that's not very ergonomic, in my opinion.
</blockquote></details>

WDYT?

drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @cryptoAtwill and @karlem)

___
*[`fendermint/testing/materializer/tests/manifests/layer2.yaml` line 66 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJJPh_o6dZbVxXBEm8o:-OJJWiX84YDcIDMZZaHY:b11t8rx) ([raw file](https://github.com/consensus-shipyard/ipc/blob/afe837f26fa0c92645370fdd5b2af2c812f0eb13/fendermint/testing/materializer/tests/manifests/layer2.yaml#L66)):*
<details><summary><i>Previously, karlem (Karel Moravec) wrote…</i></summary><blockquote>

WDYT?
</blockquote></details>

What I was getting at, the configuration should either have a consolidated representation without any paths and filesystem references, or have a way to transform itself into such a type.

karlem (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @cryptoAtwill and @drahnr)

___
*[`fendermint/testing/materializer/src/docker/node.rs` line 677 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJNa0804haNGbLr95xg:-OJNezwh2dpuvNvEXtns:b-5pys4d) ([raw file](https://github.com/consensus-shipyard/ipc/blob/43a9dcf8e63c5e1bda7e8e9090f59a1092ba3a91/fendermint/testing/materializer/src/docker/node.rs#L677)):*

When do we clean these up? Can't we use `tempfile` with an appropriate lifetime that discards itself on-drop?
</blockquote></details>

I have intentionally kept the file there. Sometimes, after the test finishes running or if the network is dropped due to a failure, it is useful to retain the generated files (which amount to many on the host disk) for debugging purposes. I believe keeping this file is beneficial, as we retain all the other files as well. It also allows re-running the container without the materializer—a technique I have used a couple of times for debugging. In my opinion, it makes sense to leave the file there.

karlem (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @cryptoAtwill and @drahnr)

___
*[`fendermint/testing/materializer/src/manifest.rs` line 84 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJNZ2aV5z9MSy0IIq2w:-OJNk99K9P39u3gIdZb0:bdcv6kv) ([raw file](https://github.com/consensus-shipyard/ipc/blob/43a9dcf8e63c5e1bda7e8e9090f59a1092ba3a91/fendermint/testing/materializer/src/manifest.rs#L84)):*

Do we really expect the given `path` to be `/`? I think that's an error, since we should have at least `/foo.bar`, where `parent()` would resolve to `Some("/")`
</blockquote></details>

Good point!
___
*[`fendermint/testing/materializer/src/manifest.rs` line 291 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJN_1ZcAdhtXsFqRAWf:-OJNuHu02pQUSbfu4Vob:bmdwvm8) ([raw file](https://github.com/consensus-shipyard/ipc/blob/43a9dcf8e63c5e1bda7e8e9090f59a1092ba3a91/fendermint/testing/materializer/src/manifest.rs#L291)):*

Style: I think the abstraction is a bit sheer, we should _not_ load the optional config lazily but upfront, so two types would imho make more sense. This would mean splitting the `Node` type into the incoming definitions with `Path` references, and then a fully expanded version for internal consumption, the transition would be a failable function call.

Discard at will, it's a bit more work but a cleaner approach.
</blockquote></details>

I considered that approach yesterday, and it certainly makes sense. The only caveat is that it would require all other types to also have a "raw" version, which is a significant change. I think this change makes sense moving forward, but not as part of this PR. I will create a ticket to address it in the future instead.
___
*[`fendermint/testing/materializer/src/docker/node.rs` line 677 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJNa0804haNGbLr95xg:-OJNm7LO1gt4hnRqNQab:b8uyc7g) ([raw file](https://github.com/consensus-shipyard/ipc/blob/43a9dcf8e63c5e1bda7e8e9090f59a1092ba3a91/fendermint/testing/materializer/src/docker/node.rs#L677)):*
<details><summary><i>Previously, karlem (Karel Moravec) wrote…</i></summary><blockquote>

I have intentionally kept the file there. Sometimes, after the test finishes running or if the network is dropped due to a failure, it is useful to retain the generated files (which amount to many on the host disk) for debugging purposes. I believe keeping this file is beneficial, as we retain all the other files as well. It also allows re-running the container without the materializer—a technique I have used a couple of times for debugging. In my opinion, it makes sense to leave the file there.
</blockquote></details>

I also believe there should be a dropper for the entire folder if we decide to add one in the future, rather than for individual generated files.

drahnr (approved):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1286)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @cryptoAtwill)

___
*[`fendermint/testing/materializer/src/docker/node.rs` line 677 at r4](https://reviewable.io/reviews/consensus-shipyard/ipc/1286#-OJNa0804haNGbLr95xg:-OJODHc6CvFormcR8iAo:b-y1putu) ([raw file](https://github.com/consensus-shipyard/ipc/blob/43a9dcf8e63c5e1bda7e8e9090f59a1092ba3a91/fendermint/testing/materializer/src/docker/node.rs#L677)):*
<details><summary><i>Previously, karlem (Karel Moravec) wrote…</i></summary><blockquote>

I also believe there should be a dropper for the entire folder if we decide to add one in the future, rather than for individual generated files.
</blockquote></details>

I think this'd call for error handling logic and a wrapping call, to retain in the error case, othewise drop.

PR #1289, opened 2025-02-17
fix(node): always use foundry `v0.3.0`

cryptoAtwill:
Fix foundry version to `v0.2.0`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1289)

### DIFF
--- a/.github/workflows/demos.yaml
+++ b/.github/workflows/demos.yaml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: v0.3.0
 
       - name: pnpm install
         run: |


.github/workflows/demos.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
         uses: foundry-rs/foundry-toolchain@v1
         with:
           version: nightly
+
```
cryptoAtwill, line 32: not sure if there is a specific reason for nightly, leaving this unchanged.

drahnr: Why not move to a fixed version? Even if it's nightly, having a fixed version is important

drahnr (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1289)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @cryptoAtwill)

.github/workflows/demos.yaml
cryptoAtwill: sad thing is it seems does not support version in nightly: https://github.com/foundry-rs/foundry-toolchain?tab=readme-ov-file#inputs

cryptoAtwill (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1289)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @drahnr)

.github/workflows/demos.yaml
drahnr: Why do we need nightly in the first place? Can we get away with using a newer fixed version?

drahnr (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1289)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @cryptoAtwill)

.github/workflows/demos.yaml
karlem: > Can we get away with using a newer fixed version?

I am using 0.3 locally and it works.

karlem (approved):

drahnr (approved):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1289)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @karlem)

PR #1295, opened 2025-02-20
feat(node): validator gas exemption

### LINKED ISSUE #1255: Rework gas parameters in Broadcaster
I have been investigating the test failures in https://github.com/consensus-shipyard/ipc/pull/1239.

The latest failure was caused by checkpoints not being signed by validators and not reaching quorum. The root cause was that the `Broadcaster` component (which validators use to dispatch signatures on-chain) was using static `gas_fee_cap` and `gas_premium` values configured in `config.toml`. By default, these values are 0.

Since we no longer accept a `gas_fee_cap` below the minimum base fee, we were correctly rejecting these signature attempts. Therefore the checkpoint remained devoid of quorum leading to a timeout in the test.

## Potential solutions

- Solution A: Auto-adjust gas parameters.
    - Setting static gas parameters against a changing gas market simply isn't sound.
    - We could repurpose the configuration settings to determine _maximum_ limits, and the Broadcaster would then calculate the appropriate values on submission.
    - How to determine values:
        - Gas fee cap. We could, for example, attempt to guarantee block inclusion within a specific time bound (e.g. the next block). Ideally we'd include some well-informed slack, but the problem is that we don't have an interface to query the on-chain gas market for the worst case scenario of base fee increase within a specific window. The simplest option is to use the current base fee * some fixed factor.
        - Premium. Ideally we'd reuse the logic of eth_maxPriorityFeePerGas; however it sits in the Eth API layer, and we can't access it from the Broadcaster. Alternatively, we could migrate the premium estimation logic to an ABCI query, where both the Eth API and the Broadcaster could access it.

- Solution B: Exempt these messages from gas costs.
    - The primary argument is that these are system messages; they're vital for continued operations.
    - However, blindly exempting these messages is risky as it can enable malicious validators to spam the chain for free.
    - Alternatively, we could conditionally exempt the message only if it succeeds, while continuing to charge if it fails. That way, invalid checkpoint signatures would cost, while correct ones would be free.
    - This would require changes to the FVM, but gas cost application is something we wanted to modularise anyway (cc @sanderpick).
    - This could solve another recurring painpoint for genuine validators: the need for validators to have subnet balance only to cover checkpoint signatures.

cryptoAtwill:
This PR introduces bottom up signature submission to the chain where txn gas will not be charged for validators. 

Closes #1255


drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1295)* status: 0 of 11 files reviewed, 4 unresolved discussions

___
*[`fendermint/rpc/src/message.rs` line 194 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYxzxeCrDAH7pP0qUS:-OJYxzxeCrDAH7pP0qUT:b-9rcx31) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/rpc/src/message.rs#L194)):*
> ```XML
>     }
> 
>     /// Send a validator message to the fendermint
> ```

nit: `to fendermint` 
___
*[`fendermint/vm/interpreter/src/chain.rs` line 424 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYyQyZ3hV_BPv4OEDq:-OJYyQyZ3hV_BPv4OEDr:blgky3b) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/vm/interpreter/src/chain.rs#L424)):*
> ```XML
>                 }
>             },
>             ChainMessage::Validator(v) => match v {
> ```

JFYI: I don't have enough context yet to judge if this design is good or not
___
*[`fendermint/vm/interpreter/src/validator.rs` line 19 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYyY8KEQdK_hTcxdgq:-OJYyY8KEQdK_hTcxdgr:b9sl3qc) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/vm/interpreter/src/validator.rs#L19)):*
> ```XML
> const SOLIDITY_SELECTOR_BYTES: usize = 8;
> 
> pub(crate) fn execute_bottom_up_signature<DB: Blockstore + Clone + 'static>(
> ```

A doc comment 'd be nice
___
*[`fendermint/rpc/src/tx.rs` line 115 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYy6W675kHc3D7zIEh:-OJYy6W675kHc3D7zIEi:bgr6uf8) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/rpc/src/tx.rs#L115)):*
> ```XML
> 
>         if calldata.len() < SOLIDITY_SELECTOR_BYTES {
>             return Err(anyhow!("invalid validator calldata"));
> ```

nit: `bail!()`
nit: the error message is not actionable, nor does it yield enough info to debug, here we clearly check for length, so something like: `received message is {x} bytes, but must be at least {SOLIDITY_SELECTOR_BYTES} for the selector`

cryptoAtwill (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1295)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @drahnr)

___
*[`fendermint/rpc/src/message.rs` line 194 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYxzxeCrDAH7pP0qUS:-OJbjrlP4BDVa9Af6MGP:b-cypxrp) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/rpc/src/message.rs#L194)):*

nit: `to fendermint`
</blockquote></details>

updated!
___
*[`fendermint/rpc/src/tx.rs` line 115 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYy6W675kHc3D7zIEh:-OJbk4iV7YUoeZOHgQn6:b-cypxrp) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/rpc/src/tx.rs#L115)):*

nit: `bail!()`
nit: the error message is not actionable, nor does it yield enough info to debug, here we clearly check for length, so something like: `received message is {x} bytes, but must be at least {SOLIDITY_SELECTOR_BYTES} for the selector`
</blockquote></details>

updated!
___
*[`fendermint/vm/interpreter/src/chain.rs` line 424 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYyQyZ3hV_BPv4OEDq:-OJbjYPq9BQkVL8w8IhS:b-regmqz) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/vm/interpreter/src/chain.rs#L424)):*

JFYI: I don't have enough context yet to judge if this design is good or not
</blockquote></details>

take your time, can discuss over this again next monday as well.
___
*[`fendermint/vm/interpreter/src/validator.rs` line 19 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJYyY8KEQdK_hTcxdgq:-OJbkMLs8bp0cdU1zboA:b-t9490a) ([raw file](https://github.com/consensus-shipyard/ipc/blob/898ba1741b8bbff8cad41d8bbba321cd0c8dfc89/fendermint/vm/interpreter/src/validator.rs#L19)):*

A doc comment 'd be nice
</blockquote></details>

added!

drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1295)* status: 0 of 13 files reviewed, 7 unresolved discussions

___
*[`fendermint/rpc/src/tx.rs` line 116 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJcgifD53s_7X_SOG3I:-OJcgifD53s_7X_SOG3J:b-5ysyy8) ([raw file](https://github.com/consensus-shipyard/ipc/blob/39022cdfb29bed23d7acaa26654003b0c7ed0d0a/fendermint/rpc/src/tx.rs#L116)):*
> ```XML
>         if calldata.len() < SOLIDITY_SELECTOR_BYTES {
>             return Err(anyhow!(
>                 "invalid validator calldata, expected at least {} but found only {}",
> ```

.. bytes
___
*[`Cargo.toml` line 250 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJcfmqdDQ6UCvjMkPjG:-OJcfmqdDQ6UCvjMkPjH:bi7geq3) ([raw file](https://github.com/consensus-shipyard/ipc/blob/39022cdfb29bed23d7acaa26654003b0c7ed0d0a/Cargo.toml#L250)):*
> ```TOML
> 
> [patch.crates-io]
> fvm = { git = "https://github.com/consensus-shipyard/ref-fvm.git", branch = "customize-gas" }
> ```

Are we planning to merge that branch into our forked main?
___
*[`fendermint/rpc/src/message.rs` line 208 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJcg55KBpGig3Wa6zZq:-OJcg55KBpGig3Wa6zZr:bxm9ure) ([raw file](https://github.com/consensus-shipyard/ipc/blob/39022cdfb29bed23d7acaa26654003b0c7ed0d0a/fendermint/rpc/src/message.rs#L208)):*
> ```XML
>             .inner
>             .transaction(to, method_num, params, value, gas_params);
>         let signed = SignedMessage::new_secp256k1(message, &self.sk, &self.chain_id)?;
> ```

Can be future refactor, but I really would like to see the following pattern:

```
let signed_msg: Signed<Message> = Message::new().sign(keystore, pubk, chain_id);
// and reverse
let signed_msg: Signed<Message> = bytes.parse()?;
let message: Message = signed_msg.verify()?;
```
___
*[`fendermint/testing/materializer/tests/docker_tests/layer2.rs` line 22 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJcgkdT8fS1whngif5y:-OJcgkdT8fS1whngif5z:b-khskac) ([raw file](https://github.com/consensus-shipyard/ipc/blob/39022cdfb29bed23d7acaa26654003b0c7ed0d0a/fendermint/testing/materializer/tests/docker_tests/layer2.rs#L22)):*
> ```XML
> const CHECKPOINT_PERIOD: u64 = 2;
> const SLEEP_SECS: u64 = 5;
> const MAX_RETRIES: u32 = 10;
> ```

Can we leave a comment why we need so many retries and such a short checkpoint period?

cryptoAtwill (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1295)* status: 0 of 15 files reviewed, 7 unresolved discussions (waiting on @drahnr)

___
*[`fendermint/rpc/src/tx.rs` line 116 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJcgifD53s_7X_SOG3I:-OM6J1zu77g8-NMibZk6:b-t9490a) ([raw file](https://github.com/consensus-shipyard/ipc/blob/39022cdfb29bed23d7acaa26654003b0c7ed0d0a/fendermint/rpc/src/tx.rs#L116)):*

.. bytes
</blockquote></details>

added!
___
*[`fendermint/testing/materializer/tests/docker_tests/layer2.rs` line 22 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1295#-OJcgkdT8fS1whngif5y:-OM6J2oe3kaxlsKqDe6Y:b-6971z1) ([raw file](https://github.com/consensus-shipyard/ipc/blob/39022cdfb29bed23d7acaa26654003b0c7ed0d0a/fendermint/testing/materializer/tests/docker_tests/layer2.rs#L22)):*

Can we leave a comment why we need so many retries and such a short checkpoint period?
</blockquote></details>

added why

fendermint/rpc/src/message.rs
```diff
@@ -191,6 +191,24 @@ impl SignedMessageFactory {
         Ok(chain)
     }
 
+    /// Send a validator message to the node
+    pub fn create_chain_message<T, F: Fn(SignedMessage) -> T>(
```
LePremierHomme, line 195: You usually call it "validator message", but here "chain message".

Issue #1210, opened 2024-11-20
Panic in fendermint when starting relayer


sam701:
I observe the following panic message in fendermint

```
thread 'tokio-runtime-worker' panicked at fendermint/vm/topdown/src/finality/null.rs:111:9:
assertion failed: previous_finality == self.last_committed_finality.read_clone()?
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-11-20T13:53:09.593342Z ERROR fendermint/app/src/main.rs:24: panicking stacktrace="   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2: std::backtrace::Backtrace::create
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/backtrace.rs:331:13
   3: fendermint::init_panic_handler::{{closure}}
             at /Users/alexei/projects/ipc/fendermint/app/src/main.rs:21:26
   4: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/alloc/src/boxed.rs:2084:9
   5: std::panicking::rust_panic_with_hook
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/panicking.rs:808:13
   6: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/panicking.rs:667:13
   7: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/sys/backtrace.rs:168:18
   8: rust_begin_unwind
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/panicking.rs:665:5
   9: core::panicking::panic_fmt
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/panicking.rs:74:14
  10: core::panicking::panic
[...]

PR #1307, opened 2025-03-07
fix(node): remove unused debug assert

cryptoAtwill:
Removing a useless debug assert that causing confusion from https://github.com/consensus-shipyard/ipc/issues/1210.

When the parent syncer launches, it will query the latest parent finality. If the network is new, the genesis epoch will be used as the `current_finality`, see https://github.com/consensus-shipyard/ipc/blob/0fa83145bf8e2569e45ba99c9ece1c17234cc8dd/fendermint/vm/topdown/src/sync/mod.rs#L140, while previous finality is None.

When the first topdown finality is committed at https://github.com/consensus-shipyard/ipc/blob/0fa83145bf8e2569e45ba99c9ece1c17234cc8dd/fendermint/vm/interpreter/src/chain.rs#L398 the `prev_finality` is coming from committing https://github.com/consensus-shipyard/ipc/blob/0fa83145bf8e2569e45ba99c9ece1c17234cc8dd/fendermint/vm/interpreter/src/chain.rs#L322

At this moment, the child subnet gateway contract does not store any parent finality, the genesis finality is only used in fendermint cache. So after committing the parent finality, `None` is returned as the previous finality, so `prev_finality == None`, when it's used in `set_new_finality`, the internal `prev_finality` is actually the genesis which does not equal to None.

This is only an issue for fresh new subnet and running with debug, other running conditions will not trigger panic. Remove this `debug_assert`. Alternatively, we can patch the contract to return `genesis finality` as the initial finality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1307)


drahnr (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1307#-:-OL-P6mG0iBy3rLZGwNa:bnfp4nl)* status: 0 of 7 files reviewed, all discussions resolved

PR #1311, opened 2025-03-12
feat(node): parse contract revert

### LINKED ISSUE #1155: ipc-cli: parse and humanize Solidity contract errors
Some contract calls revert with specific errors. For instance, trying to join a subnet as a collateral validator in federated mode fails with `revert MethodNotAllowed(ERR_PERMISSIONED_AND_BOOTSTRAPPED)`. The CLI should parse these errors and report the reason back to the user.

cryptoAtwill:
This PR closes #1155.

In `ipc-actor-abi` build process, it generates the code to scan the target contracts' abi errors and generate a static mapping that mapping error selector to the ethers AbiError. Then in the JsonClient, extracts the error message from contract and parse it against the error mapping. Reason for favouring JsonClient over middleware is inner middleware modifies the original error message and does not expose the contract revert data directly which makes obtaining the data string difficult. 

On my local computer, I didn't notice any difference in build speed from the extra step.

One downside of the current approach is that the middleware modifies or expects the error data in a specific format, so the custom JsonClient does not seem able to modify the error bytes (maybe there is a way to bypass middleware expections?) and instead can just log an error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)


drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)* status: 0 of 8 files reviewed, 7 unresolved discussions

___
*[`contract-bindings/build.rs` line 194 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZKzpB52AMjZrhFEB:-OL9ZKzpB52AMjZrhFEC:bu28of8) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/build.rs#L194)):*
> ```XML
> }
> 
> fn error_mapping_gen(mod_f: &mut fs_err::File, all_contracts: &[&str]) -> color_eyre::Result<()> {
> ```

This definitely needs documentation. It should also take an `impl std::io::Write`, and make a unit test to supplement documentation
___
*[`contract-bindings/build.rs` line 200 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9Z_beCw7lAR3quNiF:-OL9Z_beCw7lAR3quNiG:b-u3p2l6) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/build.rs#L200)):*
> ```XML
>         .iter()
>         .map(|s| {
>             // Need to convert contract name to
> ```

Is this a TODO or an explanation?
___
*[`contract-bindings/src/error_macro.rs` line 8 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZiGw4fmeu9NuqKub:-OL9ZiGw4fmeu9NuqKuc:bf6fp6s) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L8)):*
> ```XML
>             use ethers::utils::hex;
> 
>             lazy_static::lazy_static! {
> ```

If we anticipate this being used outside this crate, then we should ensure the `lazy_static` is also re-exported and use `crate-name` crate as a prefix operation.
___
*[`contract-bindings/src/error_macro.rs` line 9 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZzkI6PwTT1MvYP1r:-OL9ZzkI6PwTT1MvYP1s:b-v2gyxw) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L9)):*
> ```XML
> 
>             lazy_static::lazy_static! {
>                 static ref MAP: std::collections::BTreeMap<String, ethers::abi::ethabi::AbiError> = {
> ```

For `std` use `::std` in `macro_rules`
___
*[`contract-bindings/src/error_macro.rs` line 32 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9_Hd9-NmsoIMLuk9R:-OL9_Hd9-NmsoIMLuk9S:b39b3hn) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L32)):*
> ```XML
> 
>             impl ContractErrorParser {
>                 pub fn parse_from_bytes(bytes: &[u8]) -> anyhow::Result<Option<String>> {
> ```

I'd expect an error type with a specific variant for `UnknownErrorType{ source: String }` variant.
___
*[`contract-bindings/src/error_macro.rs` line 51 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9_4DeCqbiZnjihyyq:-OL9_4DeCqbiZnjihyyr:b-ol3pts) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L51)):*
> ```XML
> 
>                 pub fn parse_from_hex_str(err: &str) -> anyhow::Result<Option<String>> {
>                     let bytes = hex::decode(err)?;
> ```

`const_hex` is a possilby preferable impl
___
*[`contract-bindings/src/error_macro.rs` line 94 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9_fPi31JNlFazNuIe:-OL9_fPi31JNlFazNuIf:b-hqfe19) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L94)):*
> ```XML
>             None
>         );
>     }
> ```

I really really like the automatism!

drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)* status: 0 of 8 files reviewed, 8 unresolved discussions (waiting on @cryptoAtwill)

___
*[`contract-bindings/src/error_macro.rs` line 21 at r2](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9biXG8Db58v0KD6qp:-OL9biXG8Db58v0KD6qq:b-9i1hr3) ([raw file](https://github.com/consensus-shipyard/ipc/blob/12988866b05615e1ff4411c6cace51a4c96f4b2e/contract-bindings/src/error_macro.rs#L21)):*
> ```XML
>                     for (_, v) in errors.iter() {
>                         for i in v {
>                             let selector = ethers::utils::hex::encode(&i.signature().0[0..SOLIDITY_SELECTOR_BYTE_SIZE]);
> ```

Needs additional documentation _why_ we use the prefix only of the hex encdoded and some spec ref.

drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)* status: 0 of 8 files reviewed, 9 unresolved discussions (waiting on @cryptoAtwill)

___
*[`contract-bindings/build.rs` line 202 at r2](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9cSbVBELZjpJ6Vloj:-OL9cSbVBELZjpJ6Vlok:b-6nt53o) ([raw file](https://github.com/consensus-shipyard/ipc/blob/12988866b05615e1ff4411c6cace51a4c96f4b2e/contract-bindings/build.rs#L202)):*
> ```XML
>             let snake_case = camel_to_snake(s);
>             let upper_case = s.to_uppercase();
>             format!("[{snake_case}, {upper_case}_ABI]")
> ```

Smells like this could be a function call just as well?

cryptoAtwill (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)* status: 0 of 8 files reviewed, 9 unresolved discussions (waiting on @drahnr)

___
*[`contract-bindings/build.rs` line 194 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZKzpB52AMjZrhFEB:-OLDDYoj1hitOxelJ47h:b-697fwt) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/build.rs#L194)):*

This definitely needs documentation. It should also take an `impl std::io::Write`, and make a unit test to supplement documentation
</blockquote></details>

added doc
___
*[`contract-bindings/build.rs` line 200 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9Z_beCw7lAR3quNiF:-OLDDZx_38wUbADLefpA:b-avzzjz) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/build.rs#L200)):*

Is this a TODO or an explanation?
</blockquote></details>

explanation, but I removed it, should be clear from the naming.
___
*[`contract-bindings/build.rs` line 202 at r2](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9cSbVBELZjpJ6Vloj:-OLDDgpy8lpBNU_oQpb4:bvnrpcm) ([raw file](https://github.com/consensus-shipyard/ipc/blob/12988866b05615e1ff4411c6cace51a4c96f4b2e/contract-bindings/build.rs#L202)):*

Smells like this could be a function call just as well?
</blockquote></details>

updated
___
*[`contract-bindings/src/error_macro.rs` line 8 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZiGw4fmeu9NuqKub:-OLD6JiL6N_4aoY0PPxI:bjzh2fh) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L8)):*

If we anticipate this being used outside this crate, then we should ensure the `lazy_static` is also re-exported and use `crate-name` crate as a prefix operation.
</blockquote></details>

I think `MAP` should just be internal
___
*[`contract-bindings/src/error_macro.rs` line 32 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9_Hd9-NmsoIMLuk9R:-OLDO7DB2amle2C3PCV0:b-ajevs0) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L32)):*

I'd expect an error type with a specific variant for `UnknownErrorType{ source: String }` variant.
</blockquote></details>

I think if the error cannot be parsed, we should just return as it is instead of throwing an error. Because contract can call other contracts and probably we dont have the abi for the called contract. It's not an error, just some information we don't have. This will be so for cross message calls.
___
*[`contract-bindings/src/error_macro.rs` line 51 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9_4DeCqbiZnjihyyq:-OLDNzJw1j9bJoJzgrYn:bvnrpcm) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L51)):*

`const_hex` is a possilby preferable impl
</blockquote></details>

updated
___
*[`contract-bindings/src/error_macro.rs` line 21 at r2](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9biXG8Db58v0KD6qp:-OLDN8Pa7Ztz-TTPK1tm:b-7tfmjp) ([raw file](https://github.com/consensus-shipyard/ipc/blob/12988866b05615e1ff4411c6cace51a4c96f4b2e/contract-bindings/src/error_macro.rs#L21)):*

Needs additional documentation _why_ we use the prefix only of the hex encdoded and some spec ref.
</blockquote></details>

added

drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @cryptoAtwill)

___
*[`contract-bindings/src/error_macro.rs` line 8 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZiGw4fmeu9NuqKub:-OLD_JRf9AUIN0y-Ng4b:bkbf4g1) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L8)):*
<details><summary><i>Previously, cryptoAtwill wrote…</i></summary><blockquote>

I think `MAP` should just be internal
</blockquote></details>

I am not talking about the usage of the static variable`MAP` but about the symbol resolution of i.e. `lazy_static` at call site.
___
*[`contract-bindings/src/error_macro.rs` line 32 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9_Hd9-NmsoIMLuk9R:-OLDaBzQ2IbIwIBtJG7R:bxmr6ut) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L32)):*
<details><summary><i>Previously, cryptoAtwill wrote…</i></summary><blockquote>

I think if the error cannot be parsed, we should just return as it is instead of throwing an error. Because contract can call other contracts and probably we dont have the abi for the called contract. It's not an error, just some information we don't have. This will be so for cross message calls.
</blockquote></details>

That's a question of the call-site not of the return type. The return type should allow usage of either, a `Result<_>` type allows that with `.ok()`. 
___
*[`contract-bindings/src/error_macro.rs` line 14 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OLD_iT5EqPbgYm_U90g:-OLD_iT5EqPbgYm_U90h:bikl6cw) ([raw file](https://github.com/consensus-shipyard/ipc/blob/21254f728951d905428fcd424992dec759b68612/contract-bindings/src/error_macro.rs#L14)):*
> ```XML
> 
>                     $(
>                         errors.extend($crate::gen::$snake_case::$snake_case::$abi.errors.clone());
> ```

If this the only reason for using a macro, we could just as well use a function and generate the function call invocation, which would simplify this greatly and remove lot's of the above concerns
___
*[`contract-bindings/build.rs` line 196 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OLDZk0PDn04z3Qqo2Ep:-OLDZk0Q6mtIEG-6Tu-n:b69orga) ([raw file](https://github.com/consensus-shipyard/ipc/blob/21254f728951d905428fcd424992dec759b68612/contract-bindings/build.rs#L196)):*
> ```XML
> /// generate the mapping between contract error selectors to the ethers abi error type for parsing
> /// potential contract errors.
> /// This function generates a rust file that creates the error mapping, see `gen_contract_error_mapping`
> ```

nit: Please use links for symbols and capitalize terms like ABI and word starts. This will become public rendered documentation.

drahnr (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @cryptoAtwill)

___
*[`contract-bindings/src/error_macro.rs` line 9 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OL9ZzkI6PwTT1MvYP1r:-OLEeMb46ZCQ-1sb2BT0:bb6ywmx) ([raw file](https://github.com/consensus-shipyard/ipc/blob/4c41752d4eaf798959719bdba827fecd4248ed01/contract-bindings/src/error_macro.rs#L9)):*

For `std` use `::std` in `macro_rules`
</blockquote></details>

LGTM
___
*[`contract-bindings/src/error_macro.rs` line 14 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OLD_iT5EqPbgYm_U90g:-OLEePDx8Ry0r34M4yGt:bb6ywmx) ([raw file](https://github.com/consensus-shipyard/ipc/blob/21254f728951d905428fcd424992dec759b68612/contract-bindings/src/error_macro.rs#L14)):*

If this the only reason for using a macro, we could just as well use a function and generate the function call invocation, which would simplify this greatly and remove lot's of the above concerns
</blockquote></details>

LGTM
___
*[`contract-bindings/src/error_parser.rs` line 67 at r5](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-OLEdlD_-TiCLT2ueyhr:-OLEdlD_-TiCLT2ueyhs:b-rf64s7) ([raw file](https://github.com/consensus-shipyard/ipc/blob/2b5989d7b74f7b1370d8403e0792a1b80df313d8/contract-bindings/src/error_parser.rs#L67)):*
> ```XML
> 
>     pub fn parse_from_hex_str(err: &str) -> Result<String, ParseContractError> {
>         let bytes = const_hex::hex::decode(err).map_err(|e| ParseContractError::ErrorNotHexStr(e.to_string()))?;
> ```

nit: We don't need to stringify here, using a `#[source] const_hex::FromHexError`

drahnr (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1311#-:-OLEezJX8nHkyWKRDxWi:bnfp4nl)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @cryptoAtwill)

ipc/provider/src/manager/evm/error_parsing.rs
```diff
@@ -0,0 +1,62 @@
...
+use async_trait::async_trait;
+use ethers::providers::{Http, HttpClientError, JsonRpcClient};
+use ipc_actors_abis::error_parser::ContractErrorParser;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use std::fmt::{Debug, Formatter};
+
+/// A fvm contract revert parsing util
+#[derive(Clone)]
+pub struct FvmHttp {
```
karlem, line 13: Strictly speaking this is not by no means related to Fvm right? It is for EVM contracts.

ipc/provider/src/manager/evm/error_parsing.rs
```diff
@@ -0,0 +1,62 @@
...
+    async fn request<T, R>(&self, method: &str, params: T) -> Result<R, Self::Error>
+    where
+        T: Debug + Serialize + Send + Sync,
+        R: DeserializeOwned + Send,
+    {
+        self.inner
+            .request(method, params)
+            .await
+            .map_err(|client_error| match client_error {
+                HttpClientError::JsonRpcError(e) => {
```
karlem, line 42: I guess we are confident now that the data is there?

contract-bindings/src/error_parser.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const SOLIDITY_SELECTOR_BYTE_SIZE: usize = 4;
+
+pub fn extend_errors(
+    map: &mut BTreeMap<String, AbiError>,
+    contract_errors: BTreeMap<String, Vec<AbiError>>,
+) {
+    for (_, v) in contract_errors.iter() {
+        for i in v {
```
karlem, line 28: Nit: `i` sounds like an index. Can we call it `e` like an error instenad?

contract-bindings/src/error_parser.rs
```diff
@@ -0,0 +1,108 @@
+use ethers::abi::ethabi::AbiError;
+use std::collections::BTreeMap;
+
```
karlem, line 3:
```
/// Macro to extend the contract error mapping by aggregating errors from multiple contracts.
/// Each contract is specified as a pair: `[snake_case, ABI_IDENTIFIER]` where:
/// - `snake_case` is the contract name converted to snake case.
/// - `ABI_IDENTIFIER` is the contract name in uppercase with `_ABI` appended.
///
/// This macro builds a lazily initialized static map (`MAP`) that associates Solidity error selectors 
/// (first 4 bytes of an error signature, hex-encoded) with their corresponding `AbiError`.
```

karlem, line 3:
```suggestion
/// Macro to extend the contract error mapping by aggregating errors from multiple contracts.
/// Each contract is specified as a pair: `[snake_case, ABI_IDENTIFIER]` where:
/// - `snake_case` is the contract name converted to snake case.
/// - `ABI_IDENTIFIER` is the contract name in uppercase with `_ABI` appended.
///
/// This macro builds a lazily initialized static map (`MAP`) that associates Solidity error selectors 
/// (first 4 bytes of an error signature, hex-encoded) with their corresponding `AbiError`.
```

contract-bindings/src/error_parser.rs
```diff
@@ -0,0 +1,108 @@
...
+                )*
+
+                errors
+            };
+        }
+    }
+}
+
+const SOLIDITY_SELECTOR_BYTE_SIZE: usize = 4;
+
```
karlem, line 22:
```suggestion
/// Extends the provided error map with errors from a contract’s error collection.
/// For each error, it extracts the Solidity selector (first 4 bytes of the error signature),
/// hex-encodes it, and maps that selector to a clone of the `AbiError`.
///
/// If a selector already exists in the map, a warning is logged.
```

karlem (approved):
Just couple of nits and comments. But otherwise it looks good!

karlem:
@cryptoAtwill The check fails though.

ipc/provider/src/manager/evm/error_parsing.rs
cryptoAtwill: yep, should be. If not nothing we can do.

ipc/provider/src/manager/evm/error_parsing.rs
cryptoAtwill: changed!

PR #1318, opened 2025-03-17
feat: add contracts deployer

### LINKED ISSUE #1316: Remove Hardhat Dependency: Use IPC CLI to Deploy Contracts on Parent L1
**Description:**
We should update the IPC CLI to deploy contracts directly on the parent L1, removing our dependency on Hardhat. This enhancement will improve usability by enabling the IPC CLI to handle the entire deployment process.

**Background:**
Currently, we rely on Hardhat for deploying contracts parent L1. We already have a tool that utilizes Hardhat-compiled contracts and deploys them to genesis state (see [this reference](https://github.com/consensus-shipyard/ipc/blob/3174af5bd62edbb3a42fb5e87a6b09e9fc3bbee3/fendermint/eth/hardhat/src/lib.rs#L34)).
Additionally, the contracts deployer in our codebase ([this reference](https://github.com/consensus-shipyard/ipc/blob/3174af5bd62edbb3a42fb5e87a6b09e9fc3bbee3/fendermint/vm/interpreter/src/genesis.rs#L617)) could be adapted. Instead of operating directly on the FVM state, we would provide it with an Ethereum provider and deploy these contracts to the parent via transactions.

**Proposed Changes:**

- Refactor Contract Deployer:
Copy and modify the existing contracts deployer to accept an Ethereum provider, enabling it to deploy contracts to the parent L1 via transactions instead of directly modifying the FVM state.

- Integrate with IPC CLI:
Update the IPC CLI to use the refactored deployer, thereby removing the dependency on Hardhat.

**Benefits:**

- Streamlined deployment process with one unified tool.
- Reduced external dependencies, leading to a simpler development workflow.
- Improved usability as users can perform all necessary actions directly via the IPC CLI.

karlem:
Close #1316


ipc/cli/src/commands/deploy/mod.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright 2022-2024 Protocol Labs
+// SPDX-License-Identifier: MIT
+
+use anyhow::anyhow;
+use async_trait::async_trait;
+use clap::{Args, ValueEnum};
+use ethers::types::Address;
+use fendermint_eth_deployer::{EthContractDeployer, SubnetCreationPrivilege};
+use fendermint_eth_hardhat::Hardhat;
```
cryptoAtwill, line 9: will it create some kind of cycler dependency in the future?

karlem: Good question. I thought about it quite a bit and I wasn't sure where exactly to put it. Also didn't want to re-shuffle too much. I honestly think that moving forward we should have some crates that are outside of ipc and fendermint and can be used by both. WDYT?

cryptoAtwill (approved):

ipc/cli/src/commands/deploy/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        required = true,
+        help = "Submitter address, must be in the keystore"
+    )]
+    from: Address,
+
+    /// The chain ID of the target network.
+    #[arg(short, long, required = true, help = "Ethereum chain ID")]
+    chain_id: u64,
+
+    /// Directory containing the contract files.
```
LePremierHomme, line 102:
```suggestion
    /// Directory containing the compiled contract files.
```

LePremierHomme (approved):
Works well!

PR #1321, opened 2025-03-19
feat(node): filter txns by base fee

cryptoAtwill:
This PR adds extra step to filter out txn below the base fee.

Previous, if a txn's fee cap is below the base fee, the txn priority is set to I64::MIN so that cometbft will deprioritise this kind of transactions. From stress testing, it turns out, cometbft will still send them over after some time. Adding this extra layer to ensure they wont be executed on chain.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1321)


fendermint/vm/interpreter/src/fvm/interpreter.rs
```diff
@@ -213,6 +215,17 @@ where
             .check()
             .map_err(|e| CheckMessageError::InvalidMessage(e.to_string()))?;
 
+        // regardless it is recheck or not, ensure gas fee cap is more than current base fee
+        let base_fee = state.block_gas_tracker().base_fee();
```
cryptoAtwill, line 219: actually it's filtered by a config in mempool: https://github.com/ethereum/go-ethereum/blob/0983cd789ee1905aedaed96f72793e5af8466f34/core/txpool/txpool.go#L261 and https://github.com/ethereum/go-ethereum/blob/0983cd789ee1905aedaed96f72793e5af8466f34/core/txpool/validation.go#L137C29-L137C35

PR #1323, opened 2025-03-26
feat: add skeleton runner with tools

karlem:


phutchins:
Discussed adding ENV variables for each of the currently customized config values. Will do another PR later for all config values as ENV variables.

Variables could be like:
IPC_NODE_PORT
IPC_COMETBFT_PORT

phutchins:
Discussed collapsing all 3 configs (IPC Config, Fendermint, CometBFT, ETH API) into a single config which will override the default values that are hard coded.

PR #1327, opened 2025-04-01
feat(node): topdown voting refactoring

cryptoAtwill:
Previously topdown process is not as reliable as desired due to
- Indeterministic chain execution after crash because everything is stored in memory. Once the app crashed, `abci::process` will not give the same result.
- `abci::process`, `abci::deliver` makes async call to external RPC.
- Poorly implemented and complex vote tally process.
- Difficult to recover if validators are not responding due to whatever RPC issues.

This PR simplifies a lot of the above by:
- Shifting voting on chain. This is not perfect, but simplifies tons of implementation and debugging. Validator slashing and recovery is now possible. Also, fast finality will be integrated and in the pipeline, it should remove the voting completely in the future.
- No more external RPC call in `abci` and execution is deterministic
- Gas exemption for validators making votes

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1327)


ipc/provider/src/lib.rs
```diff
@@ -722,11 +724,10 @@ impl IpcProvider {
         conn.manager().list_bootstrap_nodes(subnet).await
     }
 
-    /// Returns the latest finality from the parent committed in a child subnet.
-    pub async fn latest_parent_finality(&self, subnet: &SubnetID) -> anyhow::Result<ChainEpoch> {
+    /// Returns the latest parent state from the parent committed in a child subnet.
+    pub async fn latest_parent_state(&self, subnet: &SubnetID) -> anyhow::Result<ChainEpoch> {
```
drahnr, line 728: Could you elaborate what this is? Is it a gossiped block that was imported but not finalized? How do we handle competing forks or is that not a concern at this level?

contracts/contracts/gateway/router/TopDownVotingFacet.sol
```diff
@@ -0,0 +1,153 @@
...
+
+    function castVote(bytes32 vote) internal {
+        if (!s.topdownVoting.ongoingVoteHashes.contains(vote)) {
+            revert VoteNotProposed(vote);
+        }
+
+        uint256 totalValidators = s.currentMembership.validators.length;
+        uint256 totalWeight = 0;
+
+        // can shift this query to off chain
```
karlem, line 50: ? Either do it or remove the comment?

contracts/contracts/gateway/router/TopDownVotingFacet.sol
```diff
@@ -0,0 +1,153 @@
...
+
+    event TopdownQuorumFormed(bytes32 vote, uint256 quorumThreshold, uint256 totalWeight);
+    event VotingAborted(bytes32[] votedHashes);
+
+    function latestCommitted() external view returns (uint64 blockHeight, bytes32 blockHash) {
+        blockHeight = s.topdownVoting.committedParentHeight;
+        blockHash = s.topdownVoting.committedBlockHash;
+    }
+
+    function propose(TopdownCheckpoint calldata checkpoint) external {
```
karlem, line 31: I think there is couple of issue with this contract. Please correct if I am wrong.

- Multiple Concurrent Checkpoints: Multiple proposals can be active at the same time, leading to split votes.
- Stale Votes / No Timeout: Votes can remain open indefinitely if consensus isn’t reached, potentially blocking progress.
- Out-of-Order Execution: No checks enforce sequential execution; checkpoints can be executed out of order.
- Re-Execution of Checkpoints: The same checkpoint can be re-proposed and executed again after being cleared.- Validator Lookup Efficiency: Iterating through the validator list for each vote may become inefficient with a large set.
- totalPowerVoted does not seems to be updated anywhere?

karlem (changes requested):

contracts/contracts/gateway/router/TopDownFinalityFacet.sol
```diff
@@ -16,25 +16,6 @@ contract TopDownFinalityFacet is GatewayActorModifiers {
...
-
-    /// @notice Store the validator change requests from parent.
-    /// @param changeRequests - the validator changes
-    function storeValidatorChanges(PowerChangeRequest[] calldata changeRequests) external systemActorOnly {
-        s.validatorsTracker.batchStoreChange(changeRequests);
-    }
-
     /// @notice Returns the next and start configuration numbers in the tracker of changes
     /// from the parent in the child gateway
     function getTrackerConfigurationNumbers() external view returns (uint64, uint64) {
```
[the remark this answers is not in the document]
cryptoAtwill: for getters to debug.

contracts/contracts/gateway/router/TopDownVotingFacet.sol
cryptoAtwill: Good call. Here are my responses:
- We should support this though, because some validator might propose a wrong vote and should allow others wot vote the correct one.
- I'm not sure if time out actually works, but I think you are referring to recovery mechanism?
- At the moment the best we can do is enforcing configuration number is sequential and top down nonce too, parent height does not go backwards, I will add these checks.
- Should not be this case if propose is rejecting checkpoint 
- This is an issue from `Membership`, it should be added there. I can add a follow up to optimize this. But so far I think should be ok.
- Good call, I accidentally removed it. Adding it back.

karlem: - Ok, that is a valid point — but then **how do we enforce order**? There should also be **pruning** for the stale ones.
- Yes, so it does not need to be a timeout *per se*. But right now, the **voting can just get stuck forever**, and there is no good way of resolving it, right? For example, if **malicious validators decide to withdraw from voting**.
- Yes, sounds good.

---

I do not understand those two:

- > *"Should not be this case if propose is rejecting checkpoint"*

- > *"This is an issue from Membership, it should be added there. I can add a follow-up to optimize this. But so far I think should be ok."*

What do you mean? You're **not rejecting anything right now**, right?

cryptoAtwill: - There is order by not allowing the checkpoint to go backwards, nonce and configuration number are continuous. But other than this, it's quite difficult to enforce order as it's unpredictable. There is pruning when if voting is abandoned or executed.
- We can add the exponentially decreasing power if validator is idle, but makes it more complex.
- "Should not be this case if propose is rejecting checkpoint": now proposals can be rejected if the first bullet point is not satisfied. 
- The total power is not tracked in `Membership` struct, which it should. I will add this in follow ups.

drahnr (changes requested):
> Reviewed 1 of 65 files at r1, 2 of 6 files at r2, 1 of 26 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1327)* status: 4 of 71 files reviewed, 11 unresolved discussions (waiting on @cryptoAtwill and @karlem)

___
*[`contracts/contracts/gateway/router/TopDownVotingFacet.sol` line 96 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1327#-ONty8-A7J1ommWnY3lL:-ONty8-A7J1ommWnY3lM:b8q1zs0) ([raw file](https://github.com/consensus-shipyard/ipc/blob/bf4cdabb19b373301cfa7fe439c258bf640030b9/contracts/contracts/gateway/router/TopDownVotingFacet.sol#L96)):*
> ```Solidity
>             }
> 
>             totalWeight += s.currentMembership.validators[i].weight;
> ```

I don't see `totalWeight` being used later on
___
*[`contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol` line 16 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1327#-ONtlXpc-rYvi0a_kPi0:-ONtlXpc-rYvi0a_kPi1:btzk0no) ([raw file](https://github.com/consensus-shipyard/ipc/blob/bf4cdabb19b373301cfa7fe439c258bf640030b9/contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol#L16)):*
> ```Solidity
> 
> enum VoteOutcome {
>     NotEnoughVote,
> ```

nit: `NotEnoughVotes`
___
*[`contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol` line 43 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1327#-ONtnd-w7xjDQjLTgEhF:-ONtnd-w7xjDQjLTgEhG:b-nfvxjn) ([raw file](https://github.com/consensus-shipyard/ipc/blob/bf4cdabb19b373301cfa7fe439c258bf640030b9/contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol#L43)):*
> ```Solidity
>         (, uint256 quorumThreshold) = getQuorumThreshold();
> 
>         uint256 totalNumVotes = s.topdownVoting.ongoingVoteHashes.length();
> ```

where does `s` come from? I pressumed `GatewayActorModifiers`
___
*[`contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol` line 53 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1327#-ONtuLQH0o651eT1cU4D:-ONtuLQIChWxF7v30oMC:byr3zwt) ([raw file](https://github.com/consensus-shipyard/ipc/blob/bf4cdabb19b373301cfa7fe439c258bf640030b9/contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol#L53)):*
> ```Solidity
>             }
> 
>             uint256 numXnetMsgs = s.topdownVoting.votes[vote].payload.xnetMsgs.length;
> ```

Can you elaborate, on the exact accounting.
___
*[`contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol` line 71 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1327#-ONtuaae8bswmgpupbOp:-ONtuaae8bswmgpupbOq:b6uk06k) ([raw file](https://github.com/consensus-shipyard/ipc/blob/bf4cdabb19b373301cfa7fe439c258bf640030b9/contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol#L71)):*
> ```Solidity
>     }
> 
>     function clearVotes() external systemActorOnly() {
> ```

Can you leave a comment on the preconditions? There are a few `requires`
___
*[`contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol` line 113 at r3](https://reviewable.io/reviews/consensus-shipyard/ipc/1327#-ONtvJDqAdALYWoKY7E-:-ONtvJDqAdALYWoKY7E0:b-wmwpmm) ([raw file](https://github.com/consensus-shipyard/ipc/blob/bf4cdabb19b373301cfa7fe439c258bf640030b9/contracts/contracts/gateway/router/TopDownVotingExecuteFacet.sol#L113)):*
> ```Solidity
>             }
>         }
>         threshold = totalWeight * 2 / 3;
> ```

We want at least 2/3 majority, but for integers this is `(tW * 2 + (3-1)) /3`, no? It rounds up properly

PR #1328, opened 2025-04-01
fix: avoid take and put of state

drahnr:
Avoids take and put of state

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1328)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3239,6 +3239,7 @@ dependencies = [
  "fendermint_vm_snapshot",
  "fendermint_vm_topdown",
  "fs-err",
+ "futures-util",
  "fvm",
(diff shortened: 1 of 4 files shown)


karlem (commented):
@drahnr Given the modify\_exec\_state is only used at one place, can't we just get rid of it entirely?

> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1328#-:-OMr-BxM-hS0qpReIdor:bslks3c)* status: 0 of 2 files reviewed, all discussions resolved

drahnr:
@karlem technically yes, but since this already solves it, we can revisit at a later point in time

PR #1337, opened 2025-04-16
fix: add additional require

drahnr:


contracts/contracts/gateway/GatewayManagerFacet.sol
```diff
@@ -33,6 +33,9 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
     /// @dev The subnet can optionally pass a genesis circulating supply that would be pre-allocated in the
     /// subnet from genesis (without having to wait for the subnet to be spawned to propagate the funds).
     function register(uint256 genesisCircSupply, uint256 collateral) external payable {
+        // The supplied amount must align with the arguments
+        require(msg.value == (genesisCircSupply + collateral), "Deposit amount mismatched");
```
cryptoAtwill, line 37: I think this is incorrect because the genesisCircSupply does not need to be native token, could be erc20 though. Same for collateral. This implementation assumes it's all native token.

cryptoAtwill: These lines should have the parameters checked: https://github.com/consensus-shipyard/ipc/blob/373cd58b0975703e15961daf9f5e4fa47f1261f8/contracts/contracts/gateway/GatewayManagerFacet.sol#L60?

karlem: @drahnr What do we want to do with this?

PR #1338, opened 2025-04-21
fix: submit bottom up checkpoint performs signature check

cryptoAtwill:
Currently submit bottom up checkpoint does not check for duplicated bottom up signers. This exposes multiple security risks. Solidity does not have a gas efficient way of doing sorting or in memory hashmap. The approach is to enforce the signer addresses to be sorted in ascending order. This way we can easily enforce no duplication.

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1338)

### DIFF
--- a/ipc/provider/src/checkpoint.rs
+++ b/ipc/provider/src/checkpoint.rs
@@ -241,13 +241,22 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
         bundle: BottomUpCheckpointBundle,
         event: QuorumReachedEvent,
     ) -> Result<(), anyhow::Error> {
+        let BottomUpCheckpointBundle {
+            checkpoint,
+            signatures,
+            signatories,
+        } = bundle;
+
+        // sort by address in ascending order as the contract requires it.
+        let mut pairs = signatories
+            .into_iter()
+            .zip(signatures.into_iter())
+            .collect::<Vec<_>>();
+        pairs.sort_by(|a, b| a.0.cmp(&b.0));
+        let (signatories, signatures): (Vec<_>, Vec<_>) = pairs.into_iter().unzip();
+
         let epoch = parent_handler
-            .submit_checkpoint(
-                &submitter,
-                bundle.checkpoint,
-                bundle.signatures,
-                bundle.signatories,
-            )
+            .submit_checkpoint(&submitter, checkpoint, signatures, signatories)
             .await
             .map_err(|e| {
                 anyhow!(
--- a/contracts/contracts/errors/IPCErrors.sol
+++ b/contracts/contracts/errors/IPCErrors.sol
@@ -84,6 +84,7 @@ error ValidatorAlreadyClaimed();
 error InvalidActivityProof();
 error NotOwner();
 error SignatureAddressesNotSorted();
+error DuplicateValidatorSignaturesFound();
 
 enum InvalidXnetMessageReason {
     Sender,
--- a/contracts/contracts/errors/IPCErrors.sol
+++ b/contracts/contracts/errors/IPCErrors.sol
@@ -83,6 +83,7 @@ error MissingActivityCommitment();
 error ValidatorAlreadyClaimed();
 error InvalidActivityProof();
 error NotOwner();
+error SignatureAddressesNotSorted();
 
 enum InvalidXnetMessageReason {
     Sender,
--- a/contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol
+++ b/contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity ^0.8.23;
 
-import {InvalidBatchEpoch, MaxMsgsPerBatchExceeded, InvalidSignatureErr, SignatureAddressesNotSorted, BottomUpCheckpointAlreadySubmitted, CannotSubmitFutureCheckpoint, InvalidCheckpointEpoch} from "../errors/IPCErrors.sol";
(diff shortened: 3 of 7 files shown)


drahnr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1338)* status: 0 of 6 files reviewed, 3 unresolved discussions

___
*[`ipc/provider/src/checkpoint.rs` line 254 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMJT6FAlO7mrzEZsfT:-OOMJT6FAlO7mrzEZsfU:bn57bl1) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/ipc/provider/src/checkpoint.rs#L254)):*
> ```XML
>             .into_iter()
>             .zip(signatures.into_iter())
>             .collect::<Vec<_>>();
> ```

do we want to also add a `dedup()` to the processing pipeline?
___
*[`contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol` line 70 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMIW7T9DyLB7kk1Yrp:-OOMIW7T9DyLB7kk1Yrq:b-kijr4i) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol#L70)):*
> ```Solidity
>     ) public view {
>         for (uint256 i = 1; i < signatories.length; ) {
>             if (signatories[i] <= signatories[i - 1]) {
> ```

Do we want to expose a different error for equality? I think that be more helpful. `DuplicateValidatorSignaturesFound` and do we want to bail or just ignore them?
___
*[`contracts/test/helpers/TestUtils.sol` line 164 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMILjn7wrWGJcHyK_r:-OOMILjn7wrWGJcHyK_s:ba383zd) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/contracts/test/helpers/TestUtils.sol#L164)):*
> ```Solidity
>     // function derivePubKey(uint8 seq) internal pure returns (address addr, bytes memory data) {
>     //     data = new bytes(65);
>     //     data[1] = bytes1(seq);
> ```

Can we delete this?
___
*[`contracts/test/integration/SubnetActorDiamond.t.sol` line 672 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMIl3sCRLNL7nemEiM:-OOMIl3t5aS-S5MwovNb:b1mkhj8) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/contracts/test/integration/SubnetActorDiamond.t.sol#L672)):*
> ```Solidity
>     }
> 
>     function testSubnetActorDiamond_validateActiveQuorumSignatures_DuplicatedValidators() public {
> ```

:heart_decoration:

cryptoAtwill (commented):
> *[Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1338)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @drahnr)

___
*[`contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol` line 70 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMIW7T9DyLB7kk1Yrp:-OOMeb9H3I_yGbJiwd5A:b-896fix) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol#L70)):*

Do we want to expose a different error for equality? I think that be more helpful. `DuplicateValidatorSignaturesFound` and do we want to bail or just ignore them?
</blockquote></details>

Done.
___
*[`contracts/test/helpers/TestUtils.sol` line 164 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMILjn7wrWGJcHyK_r:-OOMdbno2f5f5O6l7eeh:b-896fix) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/contracts/test/helpers/TestUtils.sol#L164)):*

Can we delete this?
</blockquote></details>

Done.
___
*[`ipc/provider/src/checkpoint.rs` line 254 at r1](https://reviewable.io/reviews/consensus-shipyard/ipc/1338#-OOMJT6FAlO7mrzEZsfT:-OOMddrS34zWA6-8Zg6F:b-npsaia) ([raw file](https://github.com/consensus-shipyard/ipc/blob/b039fec84d7f481d8c1adcec3d2029501dcdb41c/ipc/provider/src/checkpoint.rs#L254)):*

do we want to also add a `dedup()` to the processing pipeline?
</blockquote></details>

I dont think we have to though, this is coming from contract and as long as contract checks it, it should be fine.

ipc/provider/src/checkpoint.rs
```diff
@@ -243,13 +243,22 @@ impl<T: BottomUpCheckpointRelayer + Send + Sync + 'static> BottomUpCheckpointMan
...
+        let (signatories, signatures): (Vec<_>, Vec<_>) = pairs.into_iter().unzip();
+
         let epoch = parent_handler
-            .submit_checkpoint(
-                &submitter,
-                bundle.checkpoint,
-                bundle.signatures,
-                bundle.signatories,
-            )
+            .submit_checkpoint(&submitter, checkpoint, signatures, signatories)
```
drahnr, line 261: nit: it might be better to introduce a `SignatureWithSignatory` type and pass that to `submit_checkpoint` and make the relationship explicit, rather than relying on identical ordering, but jump through some extra lines of code without much need

drahnr (approved):

PR #1339, opened 2025-04-22
fix: be more conservative with checks `leave`

drahnr:
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1339)

### DIFF
--- a/contracts/contracts/subnet/SubnetActorManagerFacet.sol
+++ b/contracts/contracts/subnet/SubnetActorManagerFacet.sol
@@ -296,13 +296,17 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
                 LibSubnetActor.rmAddressFromBalanceKey(msg.sender);
                 s.supplySource.transferFunds(payable(msg.sender), genesisBalance);
             }
+        }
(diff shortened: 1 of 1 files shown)


contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -296,13 +296,17 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
                 LibSubnetActor.rmAddressFromBalanceKey(msg.sender);
                 s.supplySource.transferFunds(payable(msg.sender), genesisBalance);
             }
+        }
+        // stake might be racing with `bootstrapSubnetIfNeeded` so we need to check again
 
-            // interaction must be performed after checks and changes
+        // interaction must be performed after checks and changes
+        // we do check bootstrapping again, deliberately to avoid a complex exploit scenario
+        if (!s.bootstrapped) {
```
cryptoAtwill, line 304: Adding this check is to prevent `s.supplySource` call `stake`?

drahnr: I looked into reentrant, but it only protects against a call of the same function. Here two calls interact with the `bootstrapped` flag and hence it's an issue

karlem (approved):

PR #1340, opened 2025-04-23
fix: gensis validator power

drahnr:
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1340)

### DIFF
--- a/contracts/contracts/subnet/SubnetActorManagerFacet.sol
+++ b/contracts/contracts/subnet/SubnetActorManagerFacet.sol
@@ -196,9 +196,7 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
             // nth-call to `_increaseGenesisValidatorCollateral`
             s.genesisValidators[i].weight += amount;
         } else {
-            // otherwise we need to lookup the power from the current power table
-            // and initialize the amount from that
-            uint256 collateral = s.validatorSet.validators[validator].currentPower;
+            // otherwise we need to add new validator with given amount
             Validator memory val = Validator({
                 addr: validator,
                 weight: amount,
--- a/contracts/contracts/subnet/SubnetActorManagerFacet.sol
+++ b/contracts/contracts/subnet/SubnetActorManagerFacet.sol
@@ -167,7 +167,7 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
             LibPower.setMetadataWithConfirm(msg.sender, publicKey);
             LibPower.depositWithConfirm(msg.sender, amount);
 
-            _patchGenesisValidators(msg.sender);
+            _increaseGenesisValidatorCollateral(msg.sender);
 
             LibSubnetActor.bootstrapSubnetIfNeeded();
         } else {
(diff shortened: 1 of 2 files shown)


contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -177,31 +177,56 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
...
             unchecked {
                 ++i;
             }
         }
-        if (!alreadyValidator) {
+        // already validator if we iterate through all of them but didn't get a match
+        if (i >= length) {
+            s.genesisValidators[i].weight += amount;
+        } else {
             uint256 collateral = s.validatorSet.validators[validator].currentPower;
```
cryptoAtwill, line 198: do we still need this line?

drahnr: Hmm, I don't know. I think this is the piece of the original code I did not fully grasp. Why are we setting the new power to the `current` power, rather than the actual amount, since it didn't exist before?

contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -167,7 +167,7 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
             LibPower.setMetadataWithConfirm(msg.sender, publicKey);
             LibPower.depositWithConfirm(msg.sender, amount);
 
-            _patchGenesisValidators(msg.sender);
+            _increaseGenesisValidatorCollateral(msg.sender, amount);
```
karlem, line 170: I think the underscore is not necessary. The function is marked as internal anyway.

contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -177,31 +177,56 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
...
                 break;
             }
             unchecked {
                 ++i;
             }
         }
-        if (!alreadyValidator) {
+        // already validator if we iterate through all of them but didn't get a match
+        if (i >= length) {
+            s.genesisValidators[i].weight += amount;
```
karlem, line 196: This is incorrect right? Give we break early when found then `i < length` means found.

karlem (changes requested):
It would be good to write simple tests for both functions and besides of the comments I left we could also:
1. Emit an event when the genesis amount increases or decreases (not mandatory)
2. Decides whether we want to remove the validator from list if the weight is 0?

Both are not mandatory.

contracts/contracts/subnet/SubnetActorManagerFacet.sol
drahnr: To my understanding `_` is a helper to disambiguate, no internal vs external at the call site more easily, but doesn't have a factual call limiting fact iiuc.

contracts/contracts/subnet/SubnetActorManagerFacet.sol
drahnr: Yeah, exactly. By definition we did not find a validator key if we reach `i == length` which is the loop abort condition, so we can use `i >= length` as a defensive way of checking `none found` vs `some found`.

contracts/contracts/subnet/SubnetActorManagerFacet.sol
```diff
@@ -177,31 +177,59 @@ contract SubnetActorManagerFacet is SubnetActorModifiers, ReentrancyGuard, Pausa
...
-        if (!alreadyValidator) {
+        // if we iterate through all of them, and found one
+        // condition must hold
+        if (i < length) {
+            // nth-call to `_increaseGenesisValidatorCollateral`
+            s.genesisValidators[i].weight += amount;
+        } else {
+            // otherwise we need to lookup the power from the current power table
+            // and initialize the amount from that
             uint256 collateral = s.validatorSet.validators[validator].currentPower;
```
karlem, line 201: But it is not being used? @drahnr

drahnr: Deleted the line, I don't think it ever made much sense

PR #1341, opened 2025-04-24
fix: remove pruneBottomUpCheckpoints

cryptoAtwill:
It is reported that `pruneBottomUpCheckpoints` is never used, removing it.


LePremierHomme (commented):
Why remove it instead of using it?

cryptoAtwill:
@LePremierHomme I think the protocol does not have garbage collection at the moment. We can add it back later when GC is included, but for now, removing just keeping the code base clean.

LePremierHomme (approved):

drahnr:
@cryptoAtwill please create an issue to re-introduce / revert his commit once garbage collection is implmented

PR #1343, opened 2025-04-24
fix: add ExcessivelySafeCall

cryptoAtwill:
This PR adds `ExcessivelySafeCall` to limit the amount of bytes that will be loaded to memory. The gas however is forwarded fully.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "demos/axelar-token/lib/forge-std"]
 	path = demos/axelar-token/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "contracts/lib/ExcessivelySafeCall"]
+	path = contracts/lib/ExcessivelySafeCall
+	url = https://github.com/nomad-xyz/ExcessivelySafeCall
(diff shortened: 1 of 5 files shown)


drahnr (approved):

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "demos/axelar-token/lib/forge-std"]
 	path = demos/axelar-token/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "contracts/lib/ExcessivelySafeCall"]
```
LePremierHomme, line 13: Why using git submodule?

contracts/contracts/lib/AssetHelper.sol
```diff
@@ -10,8 +11,11 @@ import {ISubnetActor} from "../interfaces/ISubnetActor.sol";
 
 /// @notice Helpers to deal with a supply source.
 library AssetHelper {
+    using ExcessivelySafeCall for address;
     using SafeERC20 for IERC20;
 
+    uint16 constant private MAX_MEMORY_SIZE = 128;
```
LePremierHomme, line 17: Can you elaborate on the attack vector (in the case of `IERC20.transfer` call)?

.gitmodules
cryptoAtwill: I think this is the default way to install forge modules.

contracts/contracts/lib/AssetHelper.sol
cryptoAtwill: probably the IERC20 implementation is custom and the behaviour is unknown, limit to how much it loads in memory.

PR #1346, opened 2025-05-12
refactor: keystore sanity

drahnr:
Currently only the `fvm` keystore implements an encrypted variant, but we'd want encryption as an abstract overlay so it can also be utilized for the `evm` implementation.

This PR refactors the _`fvm`_ keystore implementation such that the now renamed `EncryptionOverlay` is re-usable for the `evm` as well (pending impl in a follow up PR).

- - -
This change is [Reviewable](https://reviewable.io/reviews/consensus-shipyard/ipc/1346)


drahnr:
Superseded by #1352

PR #1349, opened 2025-05-22
fix: tracing ID generation mismatch

karlem:
Match the SDK inflight msgs with correct result IDs.


drahnr:
Bar the test that needs fixin ( https://github.com/consensus-shipyard/ipc/blob/f1b6eca45de008e426c898737af817691d4aee38/contracts/test/sdk/IpcContract.t.sol#L223 ) LGTM

drahnr (approved):

PR #1351, opened 2025-05-28
fix: check return value from transfer funds

karlem:
Check the return value of the fund transfer. If it fails, the call should revert.


contracts/contracts/gateway/GatewayManagerFacet.sol
```diff
@@ -126,7 +133,13 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         delete s.subnets[id];
 
         s.subnetKeys.remove(id);
-        SubnetActorGetterFacet(msg.sender).collateralSource().transferFunds(payable(msg.sender), stake);
+        (bool success, ) = SubnetActorGetterFacet(msg.sender).collateralSource().transferFunds(
```
cryptoAtwill, line 136: can add a safe transfer method for Asset to avoid duplicated success check.

cryptoAtwill (approved):

PR #1352, opened 2025-05-28
refactor: and unify keystores

drahnr:


LePremierHomme:
If you think that all the Forest-redundant code isn't an issue, we should approach them for properly exposing it (they moved to a single-crate layout recently, but said they are willing to do such things). Otherwise, redundant code and abstractions should be removed, as it making reviewing/maintaining this more difficult.

drahnr:
After doing a pass:

I don't think the type unification of `EthAddress`, `ethers::types::Address` and the introduced type `EthKeyAddress`  is in scope of this PR. We do mix them all over the place, and the newly introduced `EthKeyAddress` _should_ remain separate to be able to evolve the storage keys independently.

phutchins:
Closing this PR for now as we will pull parts of it in as we have the bandwidth

PR #1353, opened 2025-06-02
fix: prevent xnet call messages with non-zero value from inflating child subnet circulating supply

karlem:
Prevent xnet call messages with non-zero value from inflating child subnet circulating supply


cryptoAtwill (commented):
@karlem is it possible to add a comment or doc somewhere explaining what happens to circulation for different msg kind, especially if it is a result message of a failed execution.

karlem:
@cryptoAtwill not sure what you mean. Given the PR makes all messages equal. Can you explain?

PR #1355, opened 2025-06-04
feat(node): fix checkpoint interval

cryptoAtwill:
This PR forces the next checkpoint height to be exactly "bottom up checkpoint period" away since the previous bottom up checkpoint. This is now possible after bottom up message execution is separated from bottom up checkpoint submission.


contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -106,11 +106,6 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
     /// @dev The checkpoint block height must be equal to the last bottom-up checkpoint height or
     /// @dev the next one or the number of bottom up messages exceeds the max batch size.
     function ensureValidCheckpoint(BottomUpCheckpoint calldata checkpoint) internal view {
-        uint64 maxMsgsPerBottomUpBatch = s.maxMsgsPerBottomUpBatch;
-        if (checkpoint.msgs.totalNumMsgs > maxMsgsPerBottomUpBatch) {
-            revert MaxMsgsPerBatchExceeded();
```
LePremierHomme, line 111: What is this removed?

contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -120,22 +115,9 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
...
-        }
-
-        // the expected bottom up checkpoint height, valid height
-        if (checkpoint.blockHeight == nextCheckpointHeight) {
-            return;
-        }
-
-        // if the bottom up messages' length is max, we consider that epoch valid, allow early submission
-        if (checkpoint.msgs.totalNumMsgs == s.maxMsgsPerBottomUpBatch) {
-            return;
```
LePremierHomme, line 135: Why is this removed? what will happen in case of early submission?

contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -120,22 +115,9 @@ contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard,
...
+        if (checkpoint.blockHeight != nextCheckpointHeight) {
```
LePremierHomme, line 118: Why was the previous implementation a problem needed to be fixed?

contracts/contracts/subnet/SubnetActorCheckpointingFacet.sol
cryptoAtwill: only the next immediate checkpoint can be submitted

PR #1363, opened 2025-06-12
feat(node): fix checkpoint interval

cryptoAtwill:
This PR forces the next checkpoint height to be exactly "bottom up checkpoint period" away since the previous bottom up checkpoint. This is now possible after bottom up message execution is separated from bottom up checkpoint submission.


cryptoAtwill:
@LePremierHomme Please help have another pass, I resolved all conflicts, should not have missed anything, but just another check would help, thx!

LePremierHomme (approved):
LGTM

PR #1545, opened 2026-03-11
fix: subnet bootstrapping

phutchins:
> [!NOTE]
> **Medium Risk**
> Medium risk because it changes CI build OS/container base and substantially rewires remote subnet initialization/bootstrapping, config generation, and node startup behavior, which could break deployments if assumptions about paths/ports/SSH differ.
> 
> **Overview**
> **Aligns build/runtime environments to Ubuntu 24.04** by moving GitHub Actions runners to `ubuntu-24.04` and switching the `fendermint` runner image base to `ubuntu:24.04` to match the newer glibc requirement.
> 
> **Expands `ipc-subnet-manager` to support fresh-host bootstrapping and more resilient remote ops.** Adds a `bootstrap` command (installs deps, clones/builds IPC), an `init --resume` mode, and a `diagnose` command; updates configs to support `internal_ip` for peer traffic, safer local-vs-remote path handling for `~/.ipc`, and more robust genesis/config propagation across hosts.
> 
> **Hardens node start/health/peering flows.** Node startup now reliably sets resolver/subnet env vars via a generated `start-node.sh`, health checks gain optional `--wait`, log tailing avoids remote `grep` pipes, and multiple SSH/sed operations are reworked to avoid quoting/login-shell hangs (new `exec_on_host_simple`, keepalive options, and script-based remote edits).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7431ec8873df3dd6e6f540818c2407c75f9266cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


karlem:
@phutchins thanks for the PR, but can you please clarify why bootstrap provisions a full dev/build environment on each validator host (repo clone + Rust + Foundry + Node/pnpm) instead of using prebuilt/pinned ipc-cli and fendermint artifacts?
What specific runtime or operational requirement makes on-host source builds necessary here, and is a forked repo actually required for this flow?

phutchins:
> @phutchins thanks for the PR, but can you please clarify why bootstrap provisions a full dev/build environment on each validator host (repo clone + Rust + Foundry + Node/pnpm) instead of using prebuilt/pinned ipc-cli and fendermint artifacts? What specific runtime or operational requirement makes on-host source builds necessary here, and is a forked repo actually required for this flow?

We should cross compile and push the binaries but it can wait for another iteration.

karlem (approved):

